PR #9, opened 2017-11-27
added generated code for ingress CRD

xh3b4sd:


pkg/apis/cluster/v1alpha1/ingress_types.go
```diff
@@ -0,0 +1,51 @@
+package v1alpha1
```
kopiczko: `ingress-config.giantswarm.io` from https://github.com/giantswarm/ingresstpr/blob/master/meta.go#L7

I'm not sure if we want ingress to sit in cluster group? Also it will be Ingress kind of IngressConfig we had so far. How do you want to proceed with migration when it isn't backward compatible?

xh3b4sd: The move from TRPs to CRDs is not backwards compatible anyway. I thought we write some migration code anyway. What group should it then be? I also put certs in the cluster group.

kopiczko: Let's do it that way and revisit later then.

xh3b4sd: Moving out of the cluster group makes sense to me though. I am just not sure what group it should be then. WDYT?

kopiczko: I thought about `core`, `secrets` maybe. All group names are plural in K8s. I also think it might be good to rename `cluster` to `clusters` since we don't need to be backward compatible.

```
$ curl 127.0.0.1:8001/apis | jq '.groups[].name'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  6000    0  6000    0     0   979k      0 --:--:-- --:--:-- --:--:-- 1171k
"apiregistration.k8s.io"
"extensions"
"apps"
"authentication.k8s.io"
"authorization.k8s.io"
"autoscaling"
"batch"
"certificates.k8s.io"
"networking.k8s.io"
"policy"
"rbac.authorization.k8s.io"
"settings.k8s.io"
"storage.k8s.io"
"admissionregistration.k8s.io"
"apiextensions.k8s.io"
"giantswarm.io"
```

xh3b4sd:
I merge this now and create another PR for the group name changes.

PR #10, opened 2017-11-27
readme: add contributing info

kopiczko:


README.md
```diff
@@ -4,3 +4,81 @@
...
+## Contributing
+
+### Adding a New Custom Object
+
+This is example skeleton for adding new object.
+
+- Replace `NewObj` with your object name.
+- Put struct definitions inside a proper package denoted by group and version
+  in file named `newobj_types.go`. Replace `newobj` with lowercased object
+  name.
```
xh3b4sd: It is not clear where to put the file. There is the `pkg/group/version` folder structure which we should mention and explain.

README.md
```diff
@@ -4,3 +4,81 @@
...
+### Naming Convention
+
+Custom object structs are placed in packages corresponding to the endpoints in
+Kubernetes API. E.g. structs in package
+`github.com/giantswarm/apiextensions/pkg/apis/cluster/v1alpha1` are created
+from objects under `/apis/cluster.giantswarm.io/v1alpha1/` endpoint.
+
+As this is common to have name collisions between field type names in different
+custom objects sharing the same group and version we prefix all type names
+referenced inside custom object with custom object name.
```
xh3b4sd: When there is no further rule I was still conform to the convention. :D

kopiczko:
PTALA

README.md
```diff
@@ -4,3 +4,154 @@
...
+	metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
+	return nil
+}
+```
+
+### Adding a New Custom Object
+
+This is example skeleton for adding new object.
+
+- Make sure group and version of the to add object exists (described in
```
xh3b4sd: `version of the to add object`, not sure what you want to say.

README.md
```diff
@@ -4,3 +4,154 @@
...
+  [previous paragraph](#adding-a-new-group-andor-version)).
+- Replace `NewObj` with your object name.
```
xh3b4sd: Above there is `//&Object{},` and now the object is described. Would be good to move this here a bit up.

README.md
```diff
@@ -4,3 +4,154 @@
...
+- Put struct definitions inside a proper package denoted by group and version
+  in file named `newobj_types.go`. Replace `newobj` with lowercased object
```
xh3b4sd: When it is `NewObj` the file name should be `new_obj`.

README.md
```diff
@@ -4,3 +4,154 @@
...
+}
+
+// NewObjSpec godoc.
+type NewObjSpec struct {
+	FieldName string `json:"fieldName", yaml:"fieldName"`
+}
+
+// ...
+// ...
+// ...
```
xh3b4sd: I would remove this.

pkg/apis/cluster/v1alpha1/register.go
```diff
@@ -11,6 +11,18 @@ const (
 	version = "v1alpha1"
 )
 
+// knownTypes is the full list of objects to register with the scheme. It
+// should conaint all zero values of custom objects and custom object lists
+// in the group version.
+var knownTypes = []runtime.Object{
```
xh3b4sd: I would actually make this public to have the godoc right away. This is pretty useful IMO.

kopiczko: I don't understand. There is godoc. Can you please elaborate?

Note to self: there is a typo `conaint`

README.md
kopiczko: `Make sure group and version of the object to add exists`

README.md
kopiczko: You mean reorder paragraphs? You need to have a package before creating an object, so I think it's natural order.

pkg/apis/cluster/v1alpha1/register.go
xh3b4sd: Gogoc shows you only information about public types by default. In order to see godoc for private types you need to set a URL query parameter. Nobody is doing this though. I suggested to make the type public so you see this directly because I think it is useful.

kopiczko: You mean godoc.org? Does it affect your workflow somehow? If not I'd like to leave it as it is.

xh3b4sd: I already approved your PR. When this would have been super important to me I would have requested changes.

PR #13, opened 2017-11-28
removed deep copy support

xh3b4sd:
This PR disables the generation for deep copy functions. The advantage of disabling this for now is that we can simply vendor apiextensions into our projects. The deep copy interface is for newer versions and we had conflicts. With this change we do not have to do dirty vendor hacks during our transition period. My plan is to enable the deep copy generation again as soon as we moved to CRDs and we bumped the k8s deps.


xh3b4sd:
I am probably going to pin this branch in operators since the repo here on its own fails to compile but works out when being used in operators. I do not really get it but it works and this is good enough for me as a temporary solution.

kopiczko:
Yes, let's please pin the branch for the transition period.

xh3b4sd:
I think we can close this now and delete the branch. We should not use it anymore anywhere. WDYT?

kopiczko:
kubernetesd still uses it.

xh3b4sd:
I think we can close this now. TPR migration is done.

kopiczko:
:tada:

PR #17, opened 2017-11-30
added Config suffix

xh3b4sd:


pkg/apis/cluster/v1alpha1/aws_types.go
```diff
@@ -7,39 +7,39 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// NewAWSCRD returns a new custom resource definition for AWS. This might look
-// something like the following.
+// NewAWSConfigCRD returns a new custom resource definition for AWSConfig. This
```
xh3b4sd: Kubernetes seems still to be confused about singular/plural forms. I talked with Pawel and we agreed on suffixing the types. I don't mind. We can cleanup later if this is necessary. Problem for instance is that we call a type after a Kubernetes specific type like `Ingress`. Now it is `IngressConfig` and we can move on with our lives.

pkg/apis/cluster/v1alpha1/kvm_types.go
```diff
@@ -7,39 +7,39 @@ import (
...
 //       names:
-//         kind: KVM
-//         plural: kvms
-//         singular: kvm
+//         kind: KVMConfig
+//         plural: kvmconfigs
+//         singular: kvmconfig
 //
-func NewKVMCRD() *apiextensionsv1beta1.CustomResourceDefinition {
+func NewKVMConfigCRD() *apiextensionsv1beta1.CustomResourceDefinition {
```
xh3b4sd: I did this everywhere for consistency reasons.

scripts/gen.sh
```diff
@@ -19,3 +19,5 @@ cd ${dir}/../vendor/k8s.io/code-generator && ./generate-groups.sh \
     github.com/giantswarm/apiextensions/pkg \
     github.com/giantswarm/apiextensions/pkg/apis \
     "cluster:v1alpha1 core:v1alpha1"
+
+sed -i '' -e 's,"/api","/apis",g' $(find ${dir}/../pkg/* -name "*_client.go")
```
xh3b4sd: That is super disappointing but I do not know yet why the core client has the wrong api base. I reckon we can remove this hack after having the migration finished.

pkg/clientset/versioned/typed/core/v1alpha1/core_client.go
```diff
@@ -78,7 +78,7 @@ func New(c rest.Interface) *CoreV1alpha1Client {
 func setConfigDefaults(config *rest.Config) error {
 	gv := v1alpha1.SchemeGroupVersion
 	config.GroupVersion = &gv
-	config.APIPath = "/api"
+	config.APIPath = "/apis"
```
xh3b4sd: This is now fixed by the ugly hack in our generation script.

scripts/gen.sh
kopiczko: I guess this is because of `core` package name (but the group is in fact `core.giantswarm.io`)

There is a K8s legacy `core` group which has `/api` endpoint not `/apis`: https://kubernetes.io/docs/reference/api-overview/#api-groups

xh3b4sd: Does that mean we should not use a `core` group or how could we make this work?

kopiczko: It's just a guess, but for safety we should avoid 2 things:

- Naming our objects after Kubernetes objects.
- Naming our groups after Kubernetes groups (even though they are suffixed with `.giantswarm.io`).

But this isn't 100%. I haven't experimented yet.

PR #20, opened 2017-12-01
fixed KVM types

xh3b4sd:


xh3b4sd:
This is boring. Going with it to keep moving.

PR #27, opened 2017-12-07
provider: aling azure types

kopiczko:


scripts/gen.sh
```diff
@@ -20,4 +20,6 @@ cd ${dir}/../vendor/k8s.io/code-generator && ./generate-groups.sh \
     github.com/giantswarm/apiextensions/pkg/apis \
     "core:v1alpha1 provider:v1alpha1"
 
-sed -i '' -e 's,"/api","/apis",g' $(find ${dir}/../pkg/* -name "*_client.go")
+for f in $(find ${dir}/../pkg/* -name "*_client.go"); do 
+    sed -i $f -e 's,"/api","/apis",g'
+done
```
kopiczko: I had to fix due to: https://stackoverflow.com/questions/43171648/sed-gives-sed-cant-read-no-such-file-or-directory

xh3b4sd: When this still works, fine. Bash is just voodoo for me personally.

PR #31, opened 2017-12-07
dep: pin everything to kubernetes-1.8.2

kopiczko:


kopiczko:
Why? It was `release-1.8` before. You have random stuff anyway every time you do `glide up`. I also doubt glide reads dep config.

PR #36, opened 2017-12-15
added generated draughtsman client

xh3b4sd:


xh3b4sd:
It is basically the same as it is now. I just want to make sure we have the CRD magic for the draughtsman-operator prepared as well.

PR #40, opened 2018-01-09
added node CRD types

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2334.


pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -0,0 +1,106 @@
...
+
+// +genclient
+// +genclient:noStatus
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type NodeConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              NodeConfigSpec   `json:"spec"`
+	Status            NodeConfigStatus `json:"status"`
```
xh3b4sd: We start using CRD status structures now.

pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+type NodeConfigSpec struct {
+	Guest         NodeConfigSpecGuest         `json:"guest" yaml:"guest"`
+	Host          NodeConfigSpecHost          `json:"host" yaml:"host"`
+	VersionBundle NodeConfigSpecVersionBundle `json:"versionBundle" yaml:"versionBundle"`
```
xh3b4sd: IMO we should start using version bundles in the new node operator right away. Will make development and maintenance of the operator a lot easier.

pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -0,0 +1,106 @@
...
+type NodeConfigStatus struct {
+	Conditions []NodeConfigStatusCondition `json:"conditions" yaml:"conditions"`
+}
+
+// NodeConfigStatusCondition expresses a condition in which a node may is.
+type NodeConfigStatusCondition struct {
+	// Type may be Pending, Ready, Draining, Terminating.
+	Type string `json:"type" yaml:"type"`
+	// Status may be True, False or Unknown.
+	Status string `json:"status" yaml:"status"`
```
xh3b4sd: This is inspired by https://github.com/kubernetes/community/blob/4d715ddf08fc31e7d2b8bc247c55fe8ee9bf7f06/contributors/devel/api-conventions.md#typical-status-properties.

pkg/apis/core/v1alpha1/register.go
```diff
@@ -23,6 +23,10 @@ var knownTypes = []runtime.Object{
 	&FlannelConfigList{},
 	&IngressConfig{},
 	&IngressConfigList{},
+	&NodeConfig{},
+	&NodeConfigList{},
+	&StorageConfig{},
+	&StorageConfigList{},
```
xh3b4sd: The storage types where not registered apparently. I wonder why we need to do this though because this seemed to have worked anyway.

pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -0,0 +1,106 @@
...
+	Version string `json:"version" yaml:"version"`
+}
+
+type NodeConfigStatus struct {
+	Conditions []NodeConfigStatusCondition `json:"conditions" yaml:"conditions"`
+}
+
+// NodeConfigStatusCondition expresses a condition in which a node may is.
+type NodeConfigStatusCondition struct {
+	// Type may be Pending, Ready, Draining, Terminating.
```
xh3b4sd: This is a bit rough. I think we start with something and iterate. Need to get to the real implementation and need to see the systems working together. Then we can adapt if necessary.

pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -0,0 +1,106 @@
...
+type NodeConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              NodeConfigSpec   `json:"spec"`
+	Status            NodeConfigStatus `json:"status"`
+}
+
+type NodeConfigSpec struct {
+	Guest         NodeConfigSpecGuest         `json:"guest" yaml:"guest"`
+	Host          NodeConfigSpecHost          `json:"host" yaml:"host"`
```
kopiczko: I think Guest and Host could be just string types. That would save some clutter IMO. When we want to extend those types and this is in production we should create new version of NodeConfig and do the migration anyway, so this is safe to have minimalistic approach.

xh3b4sd: This results into the following information. 

```
NodeConfig.Spec.Guest.Node.Name
NodeConfig.Spec.Host.Deployment.Name
NodeConfig.Spec.Host.Pod.Name
```

I am not sure how we can express this with one string.

kopiczko: I didn't read careful enough. What I mean here is omitting `Name` fields. In this case it would be

```
NodeConfig.Spec.Guest.Node
NodeConfig.Spec.Host.Deployment
NodeConfig.Spec.Host.Pod
```

xh3b4sd: No, I hate that. 

> Be explicit, not smart. - golang

PR #42, opened 2018-01-11
provider/v1alpha1: remove Azure node SSH config

kopiczko:
In the operator it is reused from Cluster part of the object.
See: https://github.com/giantswarm/azure-operator/pull/97.


kopiczko:
I ran ./scripts/gen.sh but it removes nothing.

PR #43, opened 2018-01-13
provider/v1alpha1: add calico subnet for AzureConfig

kopiczko:


pkg/apis/provider/v1alpha1/azure_types.go
```diff
@@ -100,10 +100,15 @@ type AzureConfigSpecAzureHostCluster struct {
...
+
+	// MasterSubnetCIDR is the CIDR for the master subnet.
 	MasterSubnetCIDR string `json:"masterSubnetCIDR" yaml:"masterSubnetCIDR"`
-	// WorkerSubnetCIDR is the CIDR for the worker subnet,
+	// WorkerSubnetCIDR is the CIDR for the worker subnet.
 	WorkerSubnetCIDR string `json:"workerSubnetCIDR" yaml:"workerSubnetCIDR"`
+
+	// CalicoSubnetCIDR is the CIDR for the calico subnet. It has to be
+	// also a worker subnet (Azure limitation).
+	CalicoSubnetCIDR string `json:"calicoSubnetCIDR" yaml:"calicoSubnetCIDR"`
```
kopiczko: Unfortunately this has to be dynamic for Azure.

@r7vme I wonder if this needs to be worker subnet of virtual network subnet.

For me it would make sense if we took 1/4 of the subnet for masters, 1/4 for workers and 1/2 for calico. It's nothing special that you have 2x more pods than workers. WDYT?

kopiczko:
I go with that now. Happy to discuss later.

PR #44, opened 2018-01-31
added deprecation comments to KVMConfig

xh3b4sd:
I want to get rid of certain settings in the custom objects. Here we deprecate the Docker image of the KVM node-controller. In the future we can then remove all related code as soon as no existing guest cluster makes use of this anymore. See also https://github.com/giantswarm/kubernetesd/pull/359 and https://github.com/giantswarm/kvm-operator/pull/366.


kopiczko:

PR #45, opened 2018-02-05
Fix copyright holder

tuommaki:
Code generator's boilerplate template had original text assigning
copyright to Kubernetes Authors. Fix it to generate right copyright
header to generated source code.


kopiczko:
Please also run `scripts/gen.sh` to update current headers.

kopiczko:
Delete branch ;)

PR #46, opened 2018-02-05
Add initial version of ClusterConfig CRD

tuommaki:
This is a first version of ClusterConfig CRD for use by
cluster-operator. This CRD shall contain required information to create
guest cluster.


pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+type ClusterConfigSpec struct {
+	APIEndpoint       string                     `json:"apiEndpoint"`
+	ID                string                     `json:"id"`
+	KubernetesVersion string                     `json:"kubernetesVersion,omitempty"`
+	Masters           []ClusterMasterSpec        `json:"masters,omitempty"`
+	Name              string                     `json:"name,omitempty"`
+	Owner             string                     `json:"owner,omitempty"`
+	VersionBundles    []ClusterVersionBundleSpec `json:"versionBundles,omitempty"`
```
tuommaki: I'm not sure if list of `VersionBundles` is the right thing here? That's still what it was in `kubernetesdclient` request.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,109 @@
...
+	Vault             ClusterVaultSpec           `json:"vault,omitempty"`
```
tuommaki: Would it be a better idea to pass vault secrets via Kubernetes secret which name would be linked here?

pkg/apis/core/v1alpha1/cluster_types.go
xh3b4sd: I think this here is similar to the node-operator. It does not need to expose version bundles but can use it to version resources for custom object reconciliation. 

UPDATE

I thought this represents the version bundles being provided by the operator. Now I see that this seems to be the version bundle the guest cluster wants to use. IMO we should scope this further to prevent confusions. 

```
type ClusterConfigSpec struct {
  Guest ClusterConfigSpecGuest `json:"guest" yaml:"guest"`
}

type ClusterConfigSpecGuest struct {
  APIEndoint string `json:"apiEndpoint" yaml:"apiEndpoint"`
  ...
  VersionBundles    []versionbundle.Bundle `json:"versionBundle" yaml:"versionBundle"`
}
```

Note the extra block for `Guest` to scope everything concerning the guest cluster. In case the cluster-operator needs to expose version bundles the version bundles block for the guest cluster would conflict with the block for the operator. Also note that I personally would create also a separate block for `API` which would have a property `Endpoint`. Also note that we should just use the `versionbundle` package and not duplicate a lot of types in case we have it defined already somewhere.

pkg/apis/core/v1alpha1/cluster_types.go
xh3b4sd: I think Vault data should not be part of the interface between cluster-service and cluster-operator. The operator should just manage it internally.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,109 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              ClusterConfigSpec `json:"spec"`
+}
+
+type ClusterConfigSpec struct {
+	APIEndpoint       string                     `json:"apiEndpoint"`
+	ID                string                     `json:"id"`
+	KubernetesVersion string                     `json:"kubernetesVersion,omitempty"`
+	Masters           []ClusterMasterSpec        `json:"masters,omitempty"`
```
xh3b4sd: When you want to follow the naming convention it would be `ClusterConfigSpecMaster`.

pkg/apis/core/v1alpha1/cluster_types.go
tuommaki: What about the maintainability of `versionbundle` deep copy functions? Shall I create a PR for them in `versionbundle` repo or how should we manage cases like this?
> pkg/apis/core/v1alpha1/zz_generated.deepcopy.go:247:12: (*in)[i].DeepCopyInto undefined (type versionbundle.Bundle has no field or method DeepCopyInto)

I think there is a risk of deep copy functions dropping out-of-sync when nearly all of the other deep copy functions are automatically generated by `scripts/gen.sh`. WDYT?

xh3b4sd: Can we generate the deep copy functions? Is the codegenerator able to provide some snippets we can borrow? I am not sure why the codegenerator cannot simply create the deep copy functions for the imported dependencies. Is this maybe a bug or a feature? We should check upstream. /cc @kopiczko WDYT?

tuommaki: We can create them automatically (with same/similar script) or manually (writing the functions by hand) in `versionbundle` repository, but not here because the package wouldn't match import. The thing is that anyone updating `versionbundle` package needs to remember update the functions after modifications. Theoretically this could possibly be implemented with GitHub pre-receive hooks, but I'm not sure how robust it would eventually be.

kopiczko: I think ClusterVersionBundleSpec has only values, that's why nothing is generated.

kopiczko: I personally wouldn't mess with `versionbundle` repo and just create an object with minimal set of required fields like it is right now.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,109 @@
...
+	InstanceType  string  `json:"instanceType,omitempty"`
+	CPUCores      int     `json:"cpuCores,omitempty"`
+	ID            string  `json:"id"`
+	MemorySizeGB  float64 `json:"memorySizeGB,omitempty"`
+	StorageSizeGB float64 `json:"storageSizeGB,omitempty"`
+}
+
+type ClusterVersionBundleSpec struct {
+	Components   []ClusterComponentVersionSpec `json:"components,omitempty"`
+	Dependencies []ClusterComponentVersionSpec `json:"dependencies,omitempty"`
```
kopiczko: Do we need that field (`Dependencies`)?

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,109 @@
...
+	ClusterNodeSpec
+}
+
+type ClusterWorkerSpec struct {
+	ClusterNodeSpec
+	Labels map[string]string `json:"labels"`
+}
+
+type ClusterNodeSpec struct {
+	InstanceType  string  `json:"instanceType,omitempty"`
```
kopiczko: InstanceType is for AWS, for Azure we'll need also `VMSize`. I wonder if it wouldn't be a good idea to prefix AWS/Azure/KVM specific stuff with prefixes.

Like:
```
AWSInstanceType
AzureVMSize
KVMCPUCores
KVMMemorySizeGB
```

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,109 @@
...
+	CPUCores      int     `json:"cpuCores,omitempty"`
+	ID            string  `json:"id"`
+	MemorySizeGB  float64 `json:"memorySizeGB,omitempty"`
+	StorageSizeGB float64 `json:"storageSizeGB,omitempty"`
```
kopiczko: What is StorageSizeGB, do we use that anywhere?

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+type ClusterComponentVersionSpec struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+}
+
+type ClusterVaultSpec struct {
+	Address string `json:"address"`
+	Token   string `json:"token"`
```
kopiczko: I have a feeling that keeping the token here is a bad idea. How is it generated? Can this be created inside cluster-operator?

pkg/apis/core/v1alpha1/cluster_types.go
tuommaki: I also think that in the long run we have less problems if we keep CRD types separated from our "domain model". With the `versionbundle` there would also be a lot of irrelevant (that is irrelevant in this use case) fields which would kind of pollute the CRD data structure. I'll continue with this simple model.

pkg/apis/core/v1alpha1/cluster_types.go
tuommaki: I'll create own types for AWS, Azure and KVM and then put platform specific variables there.

pkg/apis/core/v1alpha1/cluster_types.go
tuommaki: I'm not sure, but it's present in `kubernetesd` Request. Pure guess: Is it KVM virtual machine storage allocation?

pkg/apis/core/v1alpha1/cluster_types.go
tuommaki: I agree. I'll remove Vault specific fields from here. If those are absolutely needed, let's figure a better way then.

pkg/apis/core/v1alpha1/cluster_types.go
tuommaki: I'll try to find out. Not sure atm. It was present in `kubernetesd` Request.

tuommaki: Hmm... Since I think the context here is such that client requests specific versionbundle to be used for guest cluster construction, is it actually so that `Components` should be removed and `Dependencies` is the necessary one. Explanation on these fields is here: https://godoc.org/github.com/giantswarm/versionbundle#Bundle

kopiczko: AFAIK Dependencies are used only to create aggregated version bundles. That implies that here you only want Components that met the dependencies but that is ensured in a different place in the stack.

pkg/apis/core/v1alpha1/cluster_types.go
xh3b4sd: My thinking was that we should just be transparent and explicit and not smart. When you think this is right, go ahead.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+type ClusterConfigSpec struct {
+	Guest ClusterConfigSpecGuest `json:"guest" yaml:"guest"`
+}
+
+type ClusterConfigSpecGuest struct {
+	API               ClusterConfigSpecGuestAPI        `json:"api" yaml:"api"`
+	ID                string                           `json:"id" yaml:"id"`
+	KubernetesVersion string                           `json:"kubernetesVersion,omitempty" yaml:"kubernetesVersion,omitempty"`
```
xh3b4sd: This is actually not used at all anymore. What the oldest clusters use now is an empty version bundle version.

pkg/apis/core/v1alpha1/cluster_types.go
kopiczko: I think having a lot of irrelevant fields and polluted data structures isn't transparent. Then it's difficult to reason about the thing using the structure.

I don't understand the being "smart" part here.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,125 @@
...
+	ClusterConfigSpecGuestNode
+	Labels map[string]string `json:"labels" yaml:"labels"`
+}
+
+type ClusterConfigSpecGuestNode struct {
+	AWS   ClusterConfigSpecAWS   `json:"aws,omitempty" yaml:"aws,omitempty"`
+	Azure ClusterConfigSpecAzure `json:"azure,omitempty" yaml:"azure,omitempty"`
+	ID    string                 `json:"id" yaml:"id"`
+	KVM   ClusterConfigSpecKVM   `json:"kvm,omitempty" yaml:"kvm,omitempty"`
+}
```
xh3b4sd: What is the idea on the reconciliation here? I think it would be nice to have separate reconciliation loops and thus separately versioned resources for the providers. This would make the cluster-operator implementation cleaner and we would only watch and reconcile what an installation is interested in. Now with this mixed structure we would have ugly conditions and try to be "smart" based on the content. IMO we should not do this because it does not help maintaining the project in the long run. Having separate frameworks and informers to watch for separate things makes total sense to me.

pkg/apis/core/v1alpha1/cluster_types.go
xh3b4sd: I think everything but the name and the version is irrelevant. I am not sure why the changelogs and components are there in case you want to only use what is necessary. And then tomorrow the operator needs more info and we start to patch this here and have more work. The whole version bundle is forwarded now and we could just keep it as it is. Trying to be smart here is just a waste of time and does not hold any benefit IMO. Disagree and commit.

pkg/apis/core/v1alpha1/cluster_types.go
tuommaki: Did I understood this correctly: You suggest to separate platform specific settings to own CRDs and reconcile one those? Wouldn't this complicate `cluster-operator` implementation because it needs to track / connect separated CRDs in order to form internal consensus on what's the current guest cluster status regarding overall cluster reconciling? For example, by looking at current KVM code in `kubernetesd`: https://github.com/giantswarm/kubernetesd/blob/master/service/creator/kvm.go#L50-L64 The KVM guest node config is bundled under the same spec as Flannel VNI. From logical point of view, I still consider Flannel VNI separate from KVM guest node config so it doesn't belong under that CRD. If they are separated, my current view is that there needs to be some kind of overall achievement of consensus on ready status before`kvm-operator` CRD(s) can be created for guest cluster creation.

Am I on a wrong track here? Can you explain me how do you see the process of reconciling multiple CRDs to form guest cluster?

xh3b4sd: Lets hang and talk. This sounds more complicated than I think it is.

pkg/apis/core/v1alpha1/cluster_aws_types.go
```diff
@@ -0,0 +1,85 @@
...
+	Workers []AWSClusterConfigSpecGuestWorker `json:"workers,omitempty" yaml:"workers,omitempty"`
+}
+
+type AWSClusterConfigSpecGuestMaster struct {
+	AWSClusterConfigSpecGuestNode
+}
+
+type AWSClusterConfigSpecGuestWorker struct {
+	AWSClusterConfigSpecGuestNode
+	Labels map[string]string `json:"labels" yaml:"labels"`
```
kopiczko: For my interest: why we have labels only for workers?

pkg/apis/core/v1alpha1/cluster_azure_types.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+type AzureClusterConfigSpecGuestWorker struct {
+	AzureClusterConfigSpecGuestNode
+	Labels map[string]string `json:"labels" yaml:"labels"`
+}
+
+type AzureClusterConfigSpecGuestNode struct {
+	ID           string `json:"id" yaml:"id"`
+	InstanceType string `json:"instanceType,omitempty" yaml:"instanceType,omitempty"`
```
kopiczko: InstanceType is only for AWS.

pkg/apis/core/v1alpha1/cluster_common_types.go
```diff
@@ -0,0 +1,25 @@
...
+	Owner          string                           `json:"owner,omitempty" yaml:"owner,omitempty"`
+	VersionBundles []CommonGuestConfigVersionBundle `json:"versionBundles,omitempty" yaml:"versionBundles,omitempty"`
+}
+
+type CommonGuestConfigAPI struct {
+	Endpoint string `json:"endpoint" yaml:"endpoint"`
+}
+
+type CommonGuestConfigVersionBundle struct {
+	Dependencies []CommonGuestConfigVersionBundleComponentVersion `json:"dependencies,omitempty" yaml:"dependencies,omitempty"`
```
kopiczko: So we need dependencies, not components?

pkg/apis/core/v1alpha1/cluster_azure_types.go
```diff
@@ -0,0 +1,87 @@
...
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              AzureClusterConfigSpec `json:"spec"`
+}
+
+type AzureClusterConfigSpec struct {
+	Guest AzureClusterConfigSpecGuest `json:"guest" yaml:"guest"`
+}
+
+type AzureClusterConfigSpecGuest struct {
+	CommonGuestConfig
```
xh3b4sd: Lets please just be explicit and deal with this like with the rest. I am not a big fan of this because it makes me think and it is quite inconsistent with the rest of the structure. It is also that the naming is now weird and the JSON/YAML tags are missing. I repeat this a lot in the past, but again, in golang a very common saying is "be explicit, not smart". I like this saying a lot. IMO it makes a lot things easier.

pkg/apis/core/v1alpha1/cluster_common_types.go
```diff
@@ -0,0 +1,25 @@
+package v1alpha1
+
+type CommonGuestConfig struct {
```
xh3b4sd: The structure is "common cluster" but the file is "cluster common". Can we have that matching?

pkg/apis/core/v1alpha1/cluster_common_types.go
xh3b4sd: Again, I think we actually only need the version AFAIK. Anyway I would use the full versionbundle structure. I still do not understand what the idea here is or why all the other information are required. I feel like I still miss something.

tuommaki: Please correct if I'm wrong, but I interpreted this so that when `cluster-service` asks for a new guest cluster, it has requirements for specific component versions (i.e. the version bundle). This is what `Dependencies` is (according to godoc in `versionbundle`). I omitted `Components` because according to aforementioned godoc it represents components that some party provides. I don't think it would semantically make sense in guest cluster order. At best it would be duplicate with `Dependencies`. WDYT?

pkg/apis/core/v1alpha1/cluster_azure_types.go
tuommaki: I'll remove this.

pkg/apis/core/v1alpha1/cluster_common_types.go
xh3b4sd: I try to draw a little picture. An operator defines and exposes version bundles. It knows everything about it. Cluster-service knows which operators to "scrape". Cluster-service aggregates releases based on the version bundles. The whole version bundles plus release version is tracked in the cluster object that cluster-service persists when a cluster is created. Now the only relevant information for the operators again is the version bundle version. Only the version is given to the operators. They check if they know it and execute a resource set accordingly.

pkg/apis/core/v1alpha1/cluster_azure_types.go
kopiczko: My 2c is it's very good use of embedded struct. I don't understand how this is inconsistent? `CommonGuestConfig` has JSON/YAML tags. Did you have any problems in the past with that approach?

xh3b4sd: When you initialize the object the key is `CommonGuestConfig` where it should be `Guest`.

tuommaki: No. As Pawel said, this is struct embedding here, which I think makes totally sense because the same structure is repeated for all platforms. The path in this case would be:
`AzureClusterConfig.Spec.Guest.{ID,Name...}`

pkg/apis/core/v1alpha1/cluster_common_types.go
tuommaki: This is *Guest* not Cluster. With your change this would be `GuestCommonConfig` which I think sounds a bit confusing. I'd strongly insist on keeping the filename as it is because that way all cluster related files group together logically and in alphabetical listing. I'd like to keep this as is, but if you insist I can change this to `GuestCommonConfig`. WDYT?

pkg/apis/core/v1alpha1/cluster_common_types.go
kopiczko: So when the only relevant information for the operators again is the version bundle version I think we should only have that. Like in other CRDs. I wouldn't feel comfortable with putting all that repeated info in the CO. Also it would be clear that it's only a reference to the aggregated version bundle and I wouldn't ask all those questions. So probably it's the same for future readers.

pkg/apis/core/v1alpha1/cluster_common_types.go
xh3b4sd: With you, but the structure and the file do not match anyway. When it is "CommonGuestConfig" it should at least be `common_guest_types.go`. I do not want to overdo it here though. Do what you think is right.

pkg/apis/core/v1alpha1/cluster_azure_types.go
xh3b4sd: This is accessing the object, not creating it. When creating the object you must define the key and the value in order to initialize the structure. And it was a team consensus to not initialize sructures with the dot path, but defining all braces and commas and having indentations and stuff. So this here means this. 

```
v := AzureClusterConfigSpecGuest{
  CommonGuestConfig: v1alpha1. CommonGuestConfig{
    ID: "foo",
  },
}
```

But it should at least be this. 

```
v := AzureClusterConfigSpecGuest{
  Guest: v1alpha1. CommonGuestConfig{
    ID: "foo",
  },
}
```

But this is not possible when you embed structures inline.

pkg/apis/core/v1alpha1/cluster_aws_types.go
xh3b4sd: Uncertain.

pkg/apis/core/v1alpha1/cluster_azure_types.go
kopiczko: Yes, that's how the language was designed.

There should be `json:",inline" yaml:",inline"` tag there.

pkg/apis/core/v1alpha1/cluster_common_types.go
xh3b4sd: I like your taste here. (But only here 😛)

kopiczko:
Did Tim revoked his `requested changes`?

tuommaki:
No, but this is something we discussed with Tim earlier that one should use common sense regarding waiting for that sort of thing. I figured we reached consensus on the outcome and therefore didn't wait for merge anymore because this was blocking my progress with `cluster-operator`.

xh3b4sd:
Yes this is legit. As a side note. I have multiple "projects" running in parallel most of the time because of the synchronization periods with other humans in the team. I do not need to stress all the time only because I am waiting for a review for an hour. I can simply work on something else meanwhile. That way I have a queue of PRs that are reviewed asynchronously and depending on the team reviewing my stuff I make progress in different directions. I figured this is better for the team and me and the product.

pkg/apis/core/v1alpha1/cluster_guest_types.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+type ClusterGuestConfigAPI struct {
+	Endpoint string `json:"endpoint" yaml:"endpoint"`
+}
+
+type ClusterGuestConfigVersionBundle struct {
+	Name    string `json:"name" yaml:"name"`
+	Version string `json:"version" yaml:"version"`
+	WIP     bool   `json:"wip" yaml:"wip"`
```
xh3b4sd: For my interest, why do we track the WIP flag here? I maybe missed this during the discussions.

pkg/apis/core/v1alpha1/kvm_cluster_types.go
```diff
@@ -0,0 +1,88 @@
...
+	Masters            []KVMClusterConfigSpecGuestMaster `json:"masters,omitempty" yaml:"masters,omitempty"`
+	Workers            []KVMClusterConfigSpecGuestWorker `json:"workers,omitempty" yaml:"workers,omitempty"`
+}
+
+type KVMClusterConfigSpecGuestMaster struct {
+	KVMClusterConfigSpecGuestNode
+}
+
+type KVMClusterConfigSpecGuestWorker struct {
+	KVMClusterConfigSpecGuestNode
```
xh3b4sd: I still do not like the inline magic. FYI the tags are missing here as well as in other inline places. Is this intentional?

pkg/apis/core/v1alpha1/cluster_guest_types.go
tuommaki: This originates from the `versionbundle` as I didn't know whether it's needed or not and nobody raised an issue about this. I'll remove this.

pkg/apis/core/v1alpha1/kvm_cluster_types.go
tuommaki: My apologies here. I  completely missed these when I concentrated on the other place with the inline discussion. I'll fix these together with WIP flag removal in a new PR.

pkg/apis/core/v1alpha1/cluster_guest_types.go
xh3b4sd: As I said, the only thing we forward to the operators is the version itself. The name is necessary to filter and dispatch in kubernetesd.

pkg/apis/core/v1alpha1/kvm_cluster_types.go
xh3b4sd: No worries. Just wanted to chime in here again. All good.

PR #48, opened 2018-02-06
added flag to CertConfig to configure regeneration

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2449.

### DIFF
--- a/pkg/apis/core/v1alpha1/cert_types.go
+++ b/pkg/apis/core/v1alpha1/cert_types.go
@@ -59,14 +59,15 @@ type CertConfigSpec struct {
 }
 
 type CertConfigSpecCert struct {
-	AllowBareDomains bool     `json:"allowBareDomains" yaml:"allowBareDomains"`
-	AltNames         []string `json:"altNames" yaml:"altNames"`
-	ClusterComponent string   `json:"clusterComponent" yaml:"clusterComponent"`
-	ClusterID        string   `json:"clusterID" yaml:"clusterID"`
-	CommonName       string   `json:"commonName" yaml:"commonName"`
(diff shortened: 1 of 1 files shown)


pkg/apis/core/v1alpha1/cert_types.go
```diff
@@ -59,14 +59,15 @@ type CertConfigSpec struct {
...
-	AllowBareDomains bool     `json:"allowBareDomains" yaml:"allowBareDomains"`
-	AltNames         []string `json:"altNames" yaml:"altNames"`
-	ClusterComponent string   `json:"clusterComponent" yaml:"clusterComponent"`
-	ClusterID        string   `json:"clusterID" yaml:"clusterID"`
-	CommonName       string   `json:"commonName" yaml:"commonName"`
-	IPSANs           []string `json:"ipSans" yaml:"ipSans"`
-	Organizations    []string `json:"organizations" yaml:"organizations"`
-	TTL              string   `json:"ttl" yaml:"ttl"`
+	AllowBareDomains  bool     `json:"allowBareDomains" yaml:"allowBareDomains"`
+	AllowRegeneration bool     `json:"allowRegeneration" yaml:"allowRegeneration"`
```
xh3b4sd: The idea here is to allow regeneration for all certs but for the service account.

kopiczko: Isn't it better to give a notion that we allow regeneration by default? I would add `DisableRegatneration` instead. Up to you.

xh3b4sd: Good idea. I was thinking about this but did not find a cool name. It is `DisableRegeneration` now.

PR #50, opened 2018-02-07
improved node config structure and docs

xh3b4sd:


pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -77,7 +77,15 @@ type NodeConfigSpecGuestClusterAPI struct {
 }
 
 type NodeConfigSpecGuestNode struct {
-	Name string `json:"name" yaml:"name"`
+	// ID is the identifier of the guest cluster's master and worker nodes. In
+	// Kubernetes/Kubectl they are represented as node names. We manage them in an
+	// abstracted way because of provider specific differences.
+	//
+	//     AWS: EC2 instance DNS.
```
tuommaki: I'm curious, why instance DNS instead of instance ID?

pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -77,7 +77,15 @@ type NodeConfigSpecGuestClusterAPI struct {
...
+	//     Azure: ???.
```
kopiczko: Profit :trollface:

pkg/apis/core/v1alpha1/node_types.go
xh3b4sd: You have to ask the AWS pros. See https://gigantic.slack.com/archives/C2MS2VB37/p1518003020000197.

pkg/apis/core/v1alpha1/node_types.go
xh3b4sd: I actually wanted to have a real answer here. Do we know this already?

xh3b4sd:
@kopiczko I changed `Name` to `ID` because you said this is better in the thread in Slack. Then you changed your mind and I still think `Name` is totally fine since it is about the Kubernetes node name, which is just a k8s specific thing. So I would like to keep `Name` but go with the godoc here.

kopiczko:
@xh3b4sd I'm sorry for the introduced confusion. I didn't realize this is Kubernetes node name. `Name` makes total sense there. The godoc is very nice still.

xh3b4sd:
I still miss information for Azure though. I want to replace `???` with something valuable. `Profit` was a nice joke but I am still waiting for a real answer from you.

pkg/apis/core/v1alpha1/node_types.go
kopiczko: I can't provide any useful information before we implement scaling or https://github.com/giantswarm/azure-operator/issues/119 in Azure.

xh3b4sd: I see. Thanks for the answer. Will go with this then and we update later on when we know more.

PR #52, opened 2018-02-15
update node config docs with azure details

kopiczko:
I played with scale sets and now I know we can prefix names as we want, only
number will be generated. At the end of the day it will look like
`gs-eggs2-000004` or something similar. Kubernetes node names are basically
machine names.


pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -82,7 +82,7 @@ type NodeConfigSpecGuestNode struct {
 	// in an abstracted way because of provider specific differences.
 	//
 	//     AWS: EC2 instance DNS.
-	//     Azure: ???.
+	//     Azure: VM name.
```
xh3b4sd: Is the concept in Azure just called VM? Is there not another term like EC2 for AWS?

kopiczko: Oh shit, this is old :)

Yes in Azure it's just VM. That's official name.

PR #54, opened 2018-03-06
cleanup node config based on recent advances

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2507.


pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -88,19 +87,6 @@ type NodeConfigSpecGuestNode struct {
 	Name string `json:"name" yaml:"name"`
 }
 
-type NodeConfigSpecHost struct {
-	Deployment NodeConfigSpecHostDeployment `json:"deployment" yaml:"deployment"`
-	Pod        NodeConfigSpecHostPod        `json:"pod" yaml:"pod"`
```
xh3b4sd: I noticed that we do not need this. My thinking back then was probably towards some informational benefit but this would only benefit KVM, which makes no sense, because the node-operator is an abstracted tool for general purpose k8s node draining.

pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -111,10 +97,10 @@ type NodeConfigStatus struct {
 
 // NodeConfigStatusCondition expresses a condition in which a node may is.
 type NodeConfigStatusCondition struct {
-	// Type may be Pending, Ready, Draining, Terminating.
-	Type string `json:"type" yaml:"type"`
 	// Status may be True, False or Unknown.
 	Status string `json:"status" yaml:"status"`
+	// Type may be Pending, Ready, Draining, Drained.
+	Type string `json:"type" yaml:"type"`
```
xh3b4sd: Fixing alphabetical order.

pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -111,10 +97,10 @@ type NodeConfigStatus struct {
 
 // NodeConfigStatusCondition expresses a condition in which a node may is.
 type NodeConfigStatusCondition struct {
-	// Type may be Pending, Ready, Draining, Terminating.
-	Type string `json:"type" yaml:"type"`
 	// Status may be True, False or Unknown.
 	Status string `json:"status" yaml:"status"`
+	// Type may be Pending, Ready, Draining, Drained.
```
xh3b4sd: I changed `Terminating` to `Drained`. Reason is the node-operator sets this state and should only reason about its own final status, which is in this case that it drained a node. It did not terminate a node. We can probably remove the other statuses later on but lets just keep this now and make more experiences.

PR #55, opened 2018-03-06
Fix chartconfig spec

rossf7:

Fixes a copy paste fail in my previous PR.


pkg/apis/core/v1alpha1/chart_types.go
```diff
@@ -50,7 +50,7 @@ func NewChartConfigCRD() *apiextensionsv1beta1.CustomResourceDefinition {
 type ChartConfig struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata"`
-	Spec              CertConfigSpec `json:"spec"`
+	Spec              ChartConfigSpec `json:"spec"`
```
rossf7: Here is the fail.

PR #56, opened 2018-03-09
adds version bundle version to ingress configs

xh3b4sd:


pkg/apis/core/v1alpha1/ingress_types.go
```diff
@@ -81,6 +82,10 @@ type IngressConfigSpecProtocolPort struct {
 	Protocol    string `json:"protocol" yaml:"protocol"`
 }
 
+type IngressConfigSpecVersionBundle struct {
+	Version string `json:"version" yaml:"version"`
+}
```
xh3b4sd: Right now we are not really using it but I want to align the operators and how the resource routing works. We should also start using versions here even if we only have one and it is not expected to change very often.

PR #58, opened 2018-03-14
Add release name to chartconfig

rossf7:

Adds the release name as its needed in the chart resource to get the current state.


rossf7:
With the naming pattern we use for our releases we could just remove the `-chart` suffix.

I don't like that. it depends on us naming the chart correctly. It also makes it harder for external users of the operator. Better to be explicit and have the release name in the CRD.

PR #59, opened 2018-03-14
Towards guest cluster common domain

tuommaki:
Instead of having specific guest cluster Kubernetes API endpoint URL,
store guest cluster common domain which is supplemented with more
specific sub-domain prefixes when needed. This makes e.g. certificate
construction cleaner when just prefixing CommonDomain with relevant
sub-domain prefix such as "api" or "etcd" etc.


pkg/apis/core/v1alpha1/cluster_guest_types.go
```diff
@@ -1,17 +1,13 @@
 package v1alpha1
 
 type ClusterGuestConfig struct {
-	API            ClusterGuestConfigAPI             `json:"api" yaml:"api"`
+	CommonDomain   string                            `json:"commonDomain" yaml:"commonDomain"`
```
kopiczko: What do you think about `DNSZone`? I think it's more understandable name.

> A DNS zone is any distinct, contiguous portion of the domain name space

From https://en.wikipedia.org/wiki/DNS_zone

kopiczko: I think we should also state in the godoc that only a single label will be added to this zone. This makes it clear that the zone should start with `k8s` in our case.

tuommaki: I was a little too quick here, but this is easily changeable. I'm fine with `DNSZone` too.

PR #60, opened 2018-03-14
Improve naming: CommonDomain -> DNSZone

tuommaki:
As suggested in previous [PR](https://github.com/giantswarm/apiextensions/pull/59), DNSZone communicates purpose of this field
better than CommonDomain.

### DIFF
--- a/pkg/apis/core/v1alpha1/cluster_guest_types.go
+++ b/pkg/apis/core/v1alpha1/cluster_guest_types.go
@@ -1,7 +1,7 @@
 package v1alpha1
 
 type ClusterGuestConfig struct {
-	CommonDomain   string                            `json:"commonDomain" yaml:"commonDomain"`
+	DNSZone        string                            `json:"dnsZone" yaml:"dnsZone"`
 	ID             string                            `json:"id" yaml:"id"`
 	Name           string                            `json:"name,omitempty" yaml:"name,omitempty"`
 	Owner          string                            `json:"owner,omitempty" yaml:"owner,omitempty"`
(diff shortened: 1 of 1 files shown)


xh3b4sd:
Will all this magic now break kubernetesd in case vendor deps are updated?

kopiczko:
kubernetesd doesn't use ClusterGuestConfig AFAIK.

pkg/apis/core/v1alpha1/cluster_guest_types.go
```diff
@@ -1,7 +1,7 @@
 package v1alpha1
 
 type ClusterGuestConfig struct {
-	CommonDomain   string                            `json:"commonDomain" yaml:"commonDomain"`
+	DNSZone        string                            `json:"dnsZone" yaml:"dnsZone"`
```
kopiczko: How about godoc we add only single level label to this zone?

tuommaki: I'll add this.

tuommaki:
Yeah, this doesn't affect `kubernetesd` as this whole structure was only added for `cluster-operator`. `kubernetesd` still uses its internal `Request` which has API Endpoint.

pkg/apis/core/v1alpha1/cluster_guest_types.go
tuommaki: @kopiczko Just to verify, do you agree with this comment?

kopiczko: I was thinking about something more generic. There isn't requirement for zones to start with k8s. BUT this is good enough. If someone is confused we can further clarify.

Thanks!

PR #62, opened 2018-03-21
provider: remove AzureConfig HostCluster and Location

kopiczko:
HostCluster settings will be passed with azure-operator flag. It makes
sense as there is only single operator per host cluster. It will also
make host cluster network preallocation easier.

As for Location we don't have a choice. Because of current peering
limitations we have to have the same location for host and guest
cluster. If needed it will be easier to re-add it than support it.


MarcelMue (approved):
LGTMIIWFY

fgimenez (commented):
Are the changes for passing the `HostCluster` settings via flag already in place? This PR only removes the current configuration.

kopiczko:
@fgimenez I'm working on that with this branch imported. I won't merge before testing.

kopiczko:
Tested on godsmack.

PR #63, opened 2018-03-21
bump k8s libs

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2844.


Gopkg.toml
```diff
@@ -21,25 +21,48 @@
 #  version = "2.4.0"
 #
 required = [
-  "k8s.io/code-generator/cmd/client-gen"
+  "k8s.io/apimachinery/pkg/apimachinery/registered",
```
xh3b4sd: For code generation we also need to make sure this is not pruned.

Gopkg.toml
```diff
@@ -21,25 +21,48 @@
...
+  "k8s.io/code-generator/cmd/client-gen",
 ]
 
+
+
 [[constraint]]
   name = "k8s.io/apiextensions-apiserver"
-  version = "kubernetes-1.8.5"
+  version = "kubernetes-1.9.3"
```
xh3b4sd: Bumping the actual libs.

Gopkg.toml
```diff
@@ -21,25 +21,48 @@
...
+
+
+
+[[override]]
+  name = "k8s.io/api"
+  version = "kubernetes-1.9.3"
+
+[[override]]
+  name = "k8s.io/apiserver"
+  version = "kubernetes-1.9.3"
```
xh3b4sd: To be safe we should define the overwrites so the versions match. Otherwise projects depending on the `apiextensions` repo might have to deal with weird inheritances down the road.

Gopkg.toml
```diff
@@ -21,25 +21,48 @@
...
+  go-tests = true
+  unused-packages = true
+
+[[prune.project]]
+    name = "k8s.io/code-generator"
+    unused-packages = false
+
+[[prune.project]]
+    name = "k8s.io/gengo"
+    unused-packages = false
```
xh3b4sd: TIL we can exclude certain projects from being pruned. This here is necessary for the code generation.

pkg/apis/core/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -16,7 +16,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// This file was autogenerated by deepcopy-gen. Do not edit it manually!
+// Code generated by deepcopy-gen. DO NOT EDIT.
```
xh3b4sd: This is all the generated crap.

scripts/gen.sh
```diff
@@ -2,24 +2,9 @@
...
-# boilerplate.go.txt is copied from
-# https://github.com/kubernetes/kubernetes/blob/45db5e7260d47f01106c3f0401f9c779f9b386c0/hack/boilerplate/boilerplate.go.txt
-if [[ -z $GOPATH ]]; then
-    echo "GOPATH env var must be set" >&2
-    exit 1
-fi
-if [[ ! -f ${GOPATH}/src/k8s.io/kubernetes/hack/boilerplate/boilerplate.go.txt ]]; then
-    mkdir -p ${GOPATH}/src/k8s.io/kubernetes/hack/boilerplate
-    cp ${dir}/boilerplate.go.txt ${GOPATH}/src/k8s.io/kubernetes/hack/boilerplate/boilerplate.go.txt
-fi
```
xh3b4sd: This could be resolved using `--go-header-file ${dir}/boilerplate.go.txt` below. Now it also finally works for me, which it did not all the time before since we introduced the custom boilerplate file.

scripts/gen.sh
```diff
@@ -2,24 +2,9 @@
 
 dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-# TODO Workaround https://github.com/kubernetes/code-generator/issues/6
-# boilerplate.go.txt is copied from
-# https://github.com/kubernetes/kubernetes/blob/45db5e7260d47f01106c3f0401f9c779f9b386c0/hack/boilerplate/boilerplate.go.txt
-if [[ -z $GOPATH ]]; then
-    echo "GOPATH env var must be set" >&2
-    exit 1
-fi
```
xh3b4sd: I think this is not necessary. The compiler and generator will tell you when your GOPATH is shite.

scripts/gen.sh
```diff
@@ -2,24 +2,9 @@
...
-
 cd ${dir}/../vendor/k8s.io/code-generator && ./generate-groups.sh \
-    all \
+    "deepcopy,client" \
     github.com/giantswarm/apiextensions/pkg \
     github.com/giantswarm/apiextensions/pkg/apis \
-    "core:v1alpha1 provider:v1alpha1"
-
-for f in $(find ${dir}/../pkg/* -name "*_client.go"); do 
-    sed -i $f -e 's,"/api","/apis",g'
```
xh3b4sd: To me it looks like this is not necessary anymore. The generated code looks still good to me.

scripts/gen.sh
```diff
@@ -2,24 +2,9 @@
...
-    exit 1
-fi
-if [[ ! -f ${GOPATH}/src/k8s.io/kubernetes/hack/boilerplate/boilerplate.go.txt ]]; then
-    mkdir -p ${GOPATH}/src/k8s.io/kubernetes/hack/boilerplate
-    cp ${dir}/boilerplate.go.txt ${GOPATH}/src/k8s.io/kubernetes/hack/boilerplate/boilerplate.go.txt
-fi
-
 cd ${dir}/../vendor/k8s.io/code-generator && ./generate-groups.sh \
-    all \
+    "deepcopy,client" \
```
xh3b4sd: `all` did now also create new separate folders for informers and listers. I think we do not need this so I just made the generator do what it did so far and made the arguments here more specific.

xh3b4sd (commented):
So this here was a fucking ball ache to figure out. Please review and also try to run `./scripts/gen.sh` to see if the result is different for you.

Gopkg.toml
rossf7: This is cool.

scripts/gen.sh
rossf7: Change makes sense but I don't know much about this.

rossf7 (approved):
LGTM and I ran `/scripts/gen.sh` and no changes were made.

Gopkg.toml
kopiczko: I don't think so. It should be version clash, `dep` shouldn't allow that. AFAIK overrides work only for the built project, not for dependencies.

Gopkg.toml
kopiczko: Coooool

Gopkg.toml
xh3b4sd: Sure, but what `dep` is doing here is that it sets things to whatever matches first. So here we can end up sticking to master and the project importing `apiextensions` has to deal with this. This can cause conflicts when dep is resolving dependencies.

kopiczko: I don't think I follow.

Say we have a project using `apiextensions` and `operatorkit`. When they use different `k8s.io/apiserver` versions `dep` fails. I think it's a good thing, isn't it?

xh3b4sd: Yeah but we do not want it to fail we want to use this version and have to set it so we do not use master or whatever dep choses but the one we want to use.

fgimenez (approved):
LGTM

PR #64, opened 2018-03-22
Adds Namespace to ChartConfig

rossf7:

Adds the k8s namespace to the ChartConfig CRD.

We will make the namespace configurable in our charts but chart-operator still needs to know this to get the current state of the Helm release.


xh3b4sd (approved):

kopiczko (approved):

MarcelMue (approved):

pkg/apis/core/v1alpha1/chart_types.go
```diff
@@ -65,6 +65,10 @@ type ChartConfigSpecChart struct {
 	// Name is the fully qualified name of the Helm chart to deploy.
 	// e.g. quay.io/giantswarm/chart-operator-chart
 	Name string `json:"name" yaml:"name"`
+	// Namespace is the namespace where the Helm chart is to be deployed.
+	// This is used to get the current state of the Helm release.
+	// e.g. giantswarm
+	Namespace string `json:"namespace" yaml:"namespace"`
```
rossf7: So turns out the namespace isn't needed for getting the current state. The namespace is set when installing the chart.

https://github.com/kubernetes/helm/blob/d44853ef9ecf0baeabd9743fcbde0c8d2235d7a2/pkg/helm/interface.go#L27-L28

@fgimenez Do you need this for installing charts? If so I'll update the godoc and merge. Otherwise we can close.

fgimenez: Yes, ok to close, we only need the organization and that is already set in `chart-operator` flags. 

From appr point of view namespace is the same as organization, see for instance how the `showChannel` resource is documented https://github.com/app-registry/appr/blob/master/Documentation/server/appr-api-swagger.yaml#L590 and how we use organization in place of namespace when querying it https://github.com/giantswarm/chart-operator/blob/master/service/chartconfig/v1/appr/appr.go#L69.

rossf7:
Got it thanks, closing.

pkg/apis/core/v1alpha1/chart_types.go
fgimenez: @rossf7 sorry for the confusion, finally we need the namespace in the helm side, not needed in appr. Could you please reopen and merge?

fgimenez (approved):
Thanks!

pkg/apis/core/v1alpha1/chart_types.go
rossf7: No worries. I also removed the line below from the godoc as its too low level. Now it just explains what data the field stores.

`This is used to get the current state of the Helm release.`

PR #65, opened 2018-03-27
Add missing version bundle information to cluster configs

tuommaki:
{AWS,Azure,KVM}ClusterConfig CRs were missing CR version bundle version
so add it. It's required for correct operation of OperatorKit resource
routing.


pkg/apis/core/v1alpha1/aws_cluster_types.go
```diff
@@ -54,7 +54,8 @@ type AWSClusterConfig struct {
 }
 
 type AWSClusterConfigSpec struct {
-	Guest AWSClusterConfigSpecGuest `json:"guest" yaml:"guest"`
+	Guest         AWSClusterConfigSpecGuest         `json:"guest" yaml:"guest"`
+	VersionBundle AWSClusterConfigSpecVersionBundle `json:"versionBundle" yaml:"versionBundle"`
```
xh3b4sd: That is a bit confusing now. The provider CRs have this already: https://github.com/giantswarm/apiextensions/blob/1470b7cff67311b0f4d3c7d6349115ae568d59ee/pkg/apis/provider/v1alpha1/aws_types.go#L59. I assume this is for the cluster-operator CRs to transport the original version bundle version and set it accordingly for the provider operators.

xh3b4sd (approved):

kopiczko (approved):

pkg/apis/core/v1alpha1/aws_cluster_types.go
tuommaki: Oh no. I got fully confused here. I mixed the source CR and destination CR in provider specific resource implementation in `cluster-operator`. This is actually not required for now, but what do you think, is this good to have from the start for versioning in `cluster-service` or should this be dropped for now and only added later when needed?

tuommaki: Or not. I think I got this right. So my intention with this is to denote provider specific `cluster-operator` CR version so that `cluster-operator` can select correct resource implementation for given CR.

xh3b4sd: Then this is right here. How will then the version bundle version be transported and set for the provider operators?

tuommaki: This got sorted in Meet, but for the record: version bundle versions for provider operators are supplied in `ClusterGuestConfig.VersionBundles`: https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/core/v1alpha1/cluster_guest_types.go#L12

PR #66, opened 2018-03-29
Don't use embedded type for inheritance

tuommaki:
In Go type embedding cannot be used for inheritance. Interfaces are for
this purpose. When shared data type must be extracted for generic code,
it must have accessible field as embedded types cannot be "casted" like
in more traditional object-oriented design model.

Provider specific guest cluster config must be extracted for use in
provider independent resource implementations. Therefore this field must
be able to be extracted to shared type ClusterGuestConfig.


tuommaki:
I hit practical issue that prevented use of embedded type here. As explained in commit message, `ToClusterGuestConfig(obj interface{}) (v1alpha1.ClusterGuestConfig, error)` cannot be written if type is embedded as provider specific spec type cannot be "casted" to ClusterGuestConfig. Embedded type would have worked if provider independent resources took "accessor implementation" for CR field extraction, but as we agreed on having extractor function, this must be changed to field type.

xh3b4sd (approved):

rossf7 (approved):

MarcelMue (approved):

kopiczko:
> Provider specific guest cluster config must be extracted for use in
> provider independent resource implementations. Therefore this field must
> be able to be extracted to shared type ClusterGuestConfig.

`ToClusterGuestConfig` can just return `x.ClusterGuestConfig`. 
Same way as you can return `x.Config` now.
I don't see solution to a problem here, am I missing something?

tuommaki:
Whoa! You are indeed right: https://play.golang.org/p/Xt60Ddvltcf

Why on earth I didn't find this earlier... I was actually struggling with this similar issue earlier last week and spend some time to research if it would be somehow possible but all I could find (including consulting https://www.gopl.io/) on this was the same conclusion that this doesn't work in Go. Today when I hit this issue again, I did once more a quick search with same results. *DOH*

Good thing is that I learned something new. Big thanks @kopiczko ! I'll think about this over Easter and figure whether I'll revert this. Probably, but I'll think about this.

PR #67, opened 2018-04-06
Revert "Don't use embedded type for inheritance (#66)"

tuommaki:
This reverts commit bec18e979214981ec96c3dc71daa0b1e68da9e32.

This was unnecessary and caused by my ignorance. I tried to research for
this, but somehow missed the obvious until I was noted after merge in
PR.

Keeping shared type strictly embedded communicates better that shared
nature of it while it's also more "Go-like".


xh3b4sd (commented):
I do not see a need to revert this. I actually like it more how it is now.

kopiczko (approved):
LGTM from the yaml/json object layout is beneficial IMO.

tuommaki:
This form was already agreed on earlier. Due to my ignorance I thought this change (that would now be reverted) was needed, but as @kopiczko set me straight, this can be reverted to original form. As earlier explained, when sharing common struct, embedding is the default thing to do in Go. This is also what Kubernetes does (e.g. TypeMeta & ObjectMeta) and as @kopiczko mentioned, it also looks better in yaml/json. I will proceed with this.

PR #68, opened 2018-04-16
adds node config status primitives for common use

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918.


pkg/apis/core/v1alpha1/node_funcs.go
```diff
@@ -0,0 +1,23 @@
+package v1alpha1
+
+const (
+	NodeConfigStatusStatusTrue  = "True"
+	NodeConfigStatusTypeDrained = "Drained"
+)
+
+func (s NodeConfigStatus) HasFinalConditiion() bool {
```
xh3b4sd: We already have this in the [aws-operator](https://github.com/giantswarm/aws-operator/blob/4ed80749c47f24153e539f3c6f0d088a0f535a73/service/awsconfig/v10/resource/lifecycle/create.go#L302-L310). We will need this for KVM and Azure as well, so moving to here for common use.

pkg/apis/core/v1alpha1/node_funcs.go
```diff
@@ -0,0 +1,23 @@
...
+	for _, c := range s.Conditions {
+		if c.Type == NodeConfigStatusTypeDrained && c.Status == NodeConfigStatusStatusTrue {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (s NodeConfigStatus) NewFinalConditiion() NodeConfigStatusCondition {
```
xh3b4sd: We have this in the [node-operator](https://github.com/giantswarm/node-operator/blob/5530790c9ada231704ee296411e2d988585bc41b/service/nodeconfig/v1/resource/node/create.go#L163-L166). When we move the condition status checks to here it makes sense to have the magic strings in one place. So moving to here for the common ground.

MarcelMue (approved):
LGTM

pkg/apis/core/v1alpha1/node_funcs.go
```diff
@@ -0,0 +1,23 @@
...
+	return NodeConfigStatusCondition{
+		Status: NodeConfigStatusStatusTrue,
+		Type:   NodeConfigStatusTypeDrained,
+	}
+}
```
kopiczko: I don't think we should put any custom logic in this repo. We have repos like certs and radomkeys where we put the logic. This here isn't a big deal but let's be careful with adding things like that to apiextensions.

xh3b4sd: I hear you but I do not see how this is super specific. This is super generic. Everybody can use it to update the node config status to the final drained status and to detect such states.

kopiczko: By custom I mean not manipulating CRs in the K8s api. Let's leave it like it is, but if it grows I think we should extract the repo with shared logic.

xh3b4sd: Do you mean that there should only be type definitions and generated API clients in here?

kopiczko: That's my feeling. I think that's how kubernetes packages are designed.

xh3b4sd: I see. The line is a bit blurry here but I get you. Lets see how this goes.

PR #71, opened 2018-04-30
Add example group for memcached example operator

rossf7:
Adds the example group and types for MemcachedConfig for use in the example operator.

See https://github.com/giantswarm/operatorkit/tree/master/example/memcached-operator


pkg/apis/example/v1alpha1/memcached_types.go
```diff
@@ -0,0 +1,32 @@
...
+	Memory string `json:"memory" yaml:"memory"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type MemcachedConfigList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	Items           []MemcachedConfig `json:"items"`
+}
```
kopiczko: This looks cool, but I don't see generated client.

kopiczko (approved):
Great, thanks <3

Were client packages generated and you missed them, or there was something wrong with the gen script?

rossf7:
It was due to a problem with my local setup rather than the script. I use `gvm` but I always have problems running the gen script.

```
./scripts/gen.sh
Generating deepcopy funcs
./generate-groups.sh: line 65: /Users/ross/.gvm/pkgsets/go1.10/giantswarm:/Users/ross/.gvm/pkgsets/go1.10/global/bin/deepcopy-gen: No such file or directory
```

So I ran it in a docker image running the golang image which works fine but I hadn't copied the clientset files over :( I need to sort my gopath problem as it's not a good way to do this.

Anyway should be good now. I'll merge once CI is green.

PR #76, opened 2018-05-30
provider/v1alpha1: add new AWS DNS types

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2652
Towards https://github.com/giantswarm/giantswarm/issues/3314

The idea is to get HostedZone by DNS name, set ID into the status, cancel
reconciliation and use that information when creating CloudFormation stacks.

For reference how it's used: https://github.com/giantswarm/aws-operator/pull/974


kopiczko:
Title updated to `provider/v1alpha1: add new AWS DNS types`.

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -120,6 +149,24 @@ type AWSConfigSpecVersionBundle struct {
 	Version string `json:"version" yaml:"version"`
 }
 
+type AWSConfigStatus struct {
```
kopiczko: Not sure about the status part yet. Maybe I'll find a way to communicate between resources via context. I don't like either way, but it's better that passing hosted zones ID. Especially when it comes to migration and upgrades.

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -85,20 +100,34 @@ type AWSConfigCredentialSecret struct {
...
+
+// AWSConfigSpecAWSIngress deprecated since aws-operator v12 resources.
 type AWSConfigSpecAWSIngress struct {
 	HostedZones string                     `json:"hostedZones" yaml:"hostedZones"`
 	ELB         AWSConfigSpecAWSIngressELB `json:"elb" yaml:"elb"`
 }
 
+// AWSConfigSpecAWSIngressELB deprecated since aws-operator v12 resources.
 type AWSConfigSpecAWSIngressELB struct {
 	IdleTimeoutSeconds int `json:"idleTimeoutSeconds" yaml:"idleTimeoutSeconds"`
```
kopiczko: Those `IdleTimeoutSeconds` are just consts in kubernetesd. My idea is to move those into the aws-operator. I don't see a reason to expose them in the CR. Also it gives use better flexibility to tweak them on upgrades.

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -85,20 +100,34 @@ type AWSConfigCredentialSecret struct {
...
+// AWSConfigSpecAWSEtcdELB deprecated since aws-operator v12 resources.
 type AWSConfigSpecAWSEtcdELB struct {
 	IdleTimeoutSeconds int `json:"idleTimeoutSeconds" yaml:"idleTimeoutSeconds"`
 }
 
+type AWSConfigSpecAWSHostedZones struct {
+	API     AWSConfigSpecAWSHostedZonesZone `json:"api" yaml:"api"`
+	Etcd    AWSConfigSpecAWSHostedZonesZone `json:"etcd" yaml:"etcd"`
+	Ingress AWSConfigSpecAWSHostedZonesZone `json:"ingress" yaml:"ingress"`
+}
```
tuommaki: Do you know if there will be later more fields in addition to `Name` in `AWSConfigSpecAWSHostedZonesZone` struct? If yes, should there be an entry for wildcard record here as well?

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -120,6 +149,24 @@ type AWSConfigSpecVersionBundle struct {
...
+
+type AWSConfigStatusAWS struct {
+	HostedZones AWSConfigStatusAWSHostedZones `json:"hostedZones" yaml:"hostedZones"`
+}
+
+type AWSConfigStatusAWSHostedZones struct {
+	API     AWSConfigStatusAWSHostedZonesZone `json:"api" yaml:"api"`
+	Etcd    AWSConfigStatusAWSHostedZonesZone `json:"etcd" yaml:"etcd"`
+	Ingress AWSConfigStatusAWSHostedZonesZone `json:"ingress" yaml:"ingress"`
```
tuommaki: Is wildcard record missing here?

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -64,18 +65,32 @@ type AWSConfigSpecAWS struct {
...
+
+	// HostedZones is AWS hosted zones names in the host cluster account.
+	// For each zone there will be "CLUSTER_ID.k8s" NS record created in
+	// the host cluster account. Then for each created NS record there will
+	// be a zone created in the guest account. After that component
+	// specific records under those zones:
+	//	- api.CLUSTER_ID.k8s.{{ .Spec.AWS.HostedZones.API.Name }}
+	//	- etcd.CLUSTER_ID.k8s.{{ .Spec.AWS.HostedZones.Etcd.Name }}
+	//	- ingress.CLUSTER_ID.k8s.{{ .Spec.AWS.HostedZones.Ingress.Name }}
+	//	- *.CLUSTER_ID.k8s.{{ .Spec.AWS.HostedZones.Ingress.Name }}
```
tuommaki: Is this wildcard record literally about creating a wildcard record or just a notion for any other created record?

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -85,20 +100,34 @@ type AWSConfigCredentialSecret struct {
...
 	HostedZones string                  `json:"hostedZones" yaml:"hostedZones"`
 	ELB         AWSConfigSpecAWSEtcdELB `json:"elb" yaml:"elb"`
 }
 
+// AWSConfigSpecAWSEtcdELB deprecated since aws-operator v12 resources.
 type AWSConfigSpecAWSEtcdELB struct {
 	IdleTimeoutSeconds int `json:"idleTimeoutSeconds" yaml:"idleTimeoutSeconds"`
 }
 
+type AWSConfigSpecAWSHostedZones struct {
```
tuommaki: Are these actually records rather than zones?

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -120,6 +149,24 @@ type AWSConfigSpecVersionBundle struct {
...
 
+type AWSConfigStatus struct {
+	AWS AWSConfigStatusAWS `json:"aws" yaml:"aws"`
+}
+
+type AWSConfigStatusAWS struct {
+	HostedZones AWSConfigStatusAWSHostedZones `json:"hostedZones" yaml:"hostedZones"`
+}
+
+type AWSConfigStatusAWSHostedZones struct {
```
tuommaki: As above, are these actually records rather than zones?

pkg/apis/provider/v1alpha1/aws_types.go
kopiczko: No, they are zones.

pkg/apis/provider/v1alpha1/aws_types.go
kopiczko: There probably won't be more. I wanted to keep the structure like that to be closer to azure-operator where you have `Name` and `ResourceGroup`. In AWS HostedZone is global. The value here will be something like `gauss.eu-central-1.aws.gigantic.io` for each one - we could live with single field here but I wanted to make that distinct as it is in azure-operator.

Prefixes (including wildcard) and records are going to be computed and created inside operator.

pkg/apis/provider/v1alpha1/aws_types.go
kopiczko: No, they are zones :)

pkg/apis/provider/v1alpha1/aws_types.go
kopiczko: No, see other comment.

pkg/apis/provider/v1alpha1/aws_types.go
rossf7: Yes, we create a wildcard that points to the IC ELB. This makes it easier to test ingresses as you don't need to set up any other DNS.

pkg/apis/provider/v1alpha1/aws_types.go
rossf7: We made them configurable because customers were using Helm and making long-running connections to the API server. If we're sure this doesn't need to be configurable per cluster I'm fine with removing.

rossf7: Yes, constants or worst case installation flags should be enough. Not having it in the CR keeps it simpler.

rossf7 (approved):
This makes sense to me.

PR #77, opened 2018-05-30
Add ChartConfig status

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/3293

Enables the `status` subresource for ChartConfig. Also executed `scripts/gen.sh`.


kopiczko (approved):

MarcelMue (approved):

rossf7 (approved):
LGTM but we should check if we can scrape the status easily in Prometheus.

I'd missed that and although the status field will still be useful it's probably lower priority if we can't use it for monitoring.

fgimenez:
@rossf7 makes total sense, I'll setup a quick PoC to validate the initial approach.

indyfree (approved):
LGTM if the scraping works.

rossf7:
@fgimenez Status subresource is merged. So I think this can be closed. Thx helped a lot with getting it working!

fgimenez:
sure, thank you @rossf7!

PR #82, opened 2018-06-04
helm: unify commonDomain between charts

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/3314


xh3b4sd (approved):
When this works for you, go ahead. I am just wondering how these charts ever worked if this was wrong.

kopiczko:
I guess we use different schemes in many places. There is also a lot of env vars like `COMMON_DOMAIN`, `COMMON_DOMAIN_GUEST`, `COMMON_DOMAIN_GUEST_NO_K8S`.

I'd like there to be only a `COMMON_DOMAIN`, and make things straight.

PR #88, opened 2018-06-14
Add values to chartconfig

indyfree:
Towards https://github.com/giantswarm/giantswarm/issues/3389.


indyfree:
Is `interface{}` not allowed?
Circle complains:
> pkg/apis/core/v1alpha1/zz_generated.deepcopy.go:617:22: val.DeepCopyinterface undefined (type interface {} is interface with no methods)

rossf7 (commented):
Not an expert here sadly. Pawel is our apiextensions guru.

Could it be a problem with the JSON tags? Is something like `json:",inline,omitempty"` needed for the serialisation?

xh3b4sd:
Looks like `interface{}` is treated as generated type and a method is called on it, which obviously does not work. We should create an issue upstream and see what they can do about it. Maybe also hang in the k8s slack and see if you can get immediate help. Otherwise we need to work around this issue and maybe go with string values for now.

indyfree:
The json tags didn't do the trick sadly. I think Tim is right here. Since this is somehow urgent and I have to leave soon @rossf7 can you take this?

rossf7:
@indyfree Ack and no worries I can continue here.

@xh3b4sd Yes checking with sig-apimachinery is a good idea. Thx!

rossf7:
Closing. Design has changed. We'll now reference a config map with the values.

PR #89, opened 2018-06-18
Add ClusterNetworkConfig CRD

tuommaki:
ClusterNetworkConfig CR holds information required for guest cluster network configuration. Primarily this is just for IPAM when migrating AWS specific parts of `kubernetesd` to `cluster-operator`. Later there can be also `flanneld` related parts, if necessary.

Towards https://github.com/giantswarm/giantswarm/issues/2148

Related `cluster-operator` PRs are these:
- https://github.com/giantswarm/cluster-operator/pull/151
- https://github.com/giantswarm/cluster-operator/pull/153


pkg/apis/core/v1alpha1/cluster_network_types.go
```diff
@@ -0,0 +1,92 @@
...
+	Spec              ClusterNetworkConfigSpec   `json:"spec" yaml:"spec"`
+	Status            ClusterNetworkConfigStatus `json:"status" yaml:"status"`
+}
+
+type ClusterNetworkConfigSpec struct {
+	Cluster       ClusterNetworkConfigSpecCluster       `json:"cluster" yaml:"cluster"`
+	VersionBundle ClusterNetworkConfigSpecVersionBundle `json:"versionBundle" yaml:"versionBundle"`
+}
+
+type ClusterNetworkConfigSpecCluster struct {
```
rossf7: Could you add godoc with comments? We did that for chartconfigs and I think it makes it easier to understand what's going on. Especially as this is networking wizardry! 😄 

https://github.com/giantswarm/apiextensions/blob/7bd5a0ce29400731deed05bccaf559cedc6f2221/pkg/apis/core/v1alpha1/chart_types.go#L61-L74

rossf7 (commented):
This looks cool but I'd like to see godoc for the spec fields.

pkg/apis/core/v1alpha1/cluster_network_types.go
tuommaki: Certainly. One moment :)

rossf7 (approved):
LGTM

kopiczko:
I have mixed feelings about that. What is wrong with current ipam implementation?

xh3b4sd:
@kopiczko see https://github.com/giantswarm/giantswarm/issues/3352 for instance.

xh3b4sd (approved):

kopiczko:
Thanks for pointing out the issue.  I'll check it later.

tuommaki:
> I have mixed feelings about that. What is wrong with current ipam implementation?

It will go away when `kubernetesd` goes away. But, the truth is that the current ipam implementation is actually used with this. See this PR: https://github.com/giantswarm/cluster-operator/pull/153

Reason for this design is that if current code from `kubernetesd` would be taken directly out and used in provider specific resource implementations in `cluster-operator`, it would lead to implementation that has issues with subnet management in error situations. When CRUD resource executes various independent steps in order to reconcile the resource, it cannot properly handle releasing of unused subnet allocation in case of error. Therefore I created a separate CRD and controller for it so that single resource can manage it in a single function. Current implementation uses existing ipam implementation including `crdstorage` for state management, but my mid-far future plan is to remove this and start gathering the state purely from existing CR objects. Does this make sense to you?

kopiczko:
> it cannot properly handle releasing of unused subnet allocation in case of error

Yes, that speaks to me. Thanks for explanation. I'll check other PRs.

tuommaki:
After discussion on IPAM design, this will be implemented a bit differently. Therefore closing this.

PR #90, opened 2018-06-21
Azure VPN gateway subnet

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/3316
VPN Gateway in azure requires to have its own subnet.

### DIFF
--- a/helm/apiextensions-azure-config-e2e-chart/templates/cluster.yaml
+++ b/helm/apiextensions-azure-config-e2e-chart/templates/cluster.yaml
@@ -25,6 +25,7 @@ spec:
       cidr: "{{ .Values.azure.cidr }}"
       masterSubnetCIDR: "{{ .Values.azure.masterSubnetCIDR }}"
       workerSubnetCIDR: "{{ .Values.azure.workerSubnetCIDR }}"
+      vpnSubnetCIDR: "{{ .Values.azure.vpnSubnetCIDR }}"
   cluster:
     calico:
       cidr: 16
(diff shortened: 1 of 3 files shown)


helm/apiextensions-azure-config-e2e-chart/templates/cluster.yaml
```diff
@@ -25,6 +25,7 @@ spec:
       cidr: "{{ .Values.azure.cidr }}"
       masterSubnetCIDR: "{{ .Values.azure.masterSubnetCIDR }}"
       workerSubnetCIDR: "{{ .Values.azure.workerSubnetCIDR }}"
+      vpnSubnetCIDR: "{{ .Values.azure.vpnSubnetCIDR }}"
```
xh3b4sd: V comes before W in the alphabet.

helm/apiextensions-azure-config-e2e-chart/values.yaml
```diff
@@ -13,3 +13,4 @@ azure:
   cidr: "10.42.0.0/16"
   masterSubnetCIDR: "10.42.0.0/24"
   workerSubnetCIDR: "10.42.1.0/24"
+  vpnSubnetCIDR: "10.42.2.0/27"
```
kopiczko: Out of curiosity: why it's /27?

kopiczko (approved):

helm/apiextensions-azure-config-e2e-chart/values.yaml
TheoBrigitte: Good point, this is azure "recommendation" but I did not found so far why we should restrict this. Let's extend it to /24

PR #93, opened 2018-06-26
Remove RoteTableNames field

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/3515


pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -132,11 +132,16 @@ type AWSConfigSpecAWSNode struct {
...
 type AWSConfigSpecAWSVPC struct {
-	CIDR              string   `json:"cidr" yaml:"cidr"`
-	PrivateSubnetCIDR string   `json:"privateSubnetCidr" yaml:"privateSubnetCidr"`
-	PublicSubnetCIDR  string   `json:"publicSubnetCidr" yaml:"publicSubnetCidr"`
-	RouteTableNames   []string `json:"routeTableNames" yaml:"routeTableNames"`
-	PeerID            string   `json:"peerId" yaml:"peerId"`
+	CIDR              string                             `json:"cidr" yaml:"cidr"`
+	PrivateSubnetCIDR string                             `json:"privateSubnetCidr" yaml:"privateSubnetCidr"`
+	PublicSubnetCIDR  string                             `json:"publicSubnetCidr" yaml:"publicSubnetCidr"`
+	RouteTableNames   AWSConfigSpecAWSVPCRouteTableNames `json:"routeTableNames" yaml:"routeTableNames"`
```
kopiczko: We have those values in installations. Isn't it better to pass them as aws-operator flags? We can't have more that a single host cluster per aws-operator, so correct me if I'm wrong but I don't think it makes sense to pass them on CR level.

kopiczko: If it's really required to be here I'd call it `RouteTables` so we can put also other things under this key. Just in case.

fgimenez: I agree, IMO these values should be set as flags and that's it, i don't know why they are included in the CR. I proposed these changes because while working on kubernetesd I noticed it adds the private route table names to the CR, see https://github.com/giantswarm/kubernetesd/blob/master/service/creator/aws.go#L107

WDYT, should we keep it?

kopiczko: I think it should be removed from kubernetsd/CR and set as a flag. That would be work towards https://github.com/giantswarm/giantswarm/issues/2383

fgimenez: Done, will propose the changess to kubernetesd next, PTAL.

kopiczko (approved):
<3

xh3b4sd:
I think this may got stale. Closing.

PR #96, opened 2018-06-29
adds status subresources to provider CRs

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918. This PR aims to provide primitives we can use in https://github.com/giantswarm/statusresource/pull/1. The status related functionality implemented here will be used in the status resource implementation to manage CR statuses across providers. Later provider operators can also manage status independent from the common status resource but this is out of scope here.


kopiczko (approved):

pkg/apis/provider/v1alpha1/status_funcs_test.go
```diff
@@ -0,0 +1,98 @@
...
+			},
+			ExpectedSemver: "",
+		},
+		{
+			Name: "case 1",
+			StatusCluster: StatusCluster{
+				Versions: []StatusClusterVersion{
+					{
+						Date:   time.Unix(10, 0),
+						Semver: "1.0.0",
```
rossf7: TImestamp and semver format here is cool. I think we can extend later for event stream.

rossf7 (approved):
LGTM

pkg/apis/provider/v1alpha1/status_funcs_test.go
xh3b4sd: Yeah so the versions is basically just a little record of history in that sense so we know like the last 5 versions the cluster went through or something. We basically get the history and have a medium to store the current and the next version in that sense while obtaining the state transition of the versions.

PR #97, opened 2018-07-02
Revert vpn cidr

TheoBrigitte:
As VPN CIDR computation is implemented within the azure-operator (see https://github.com/giantswarm/azure-operator/pull/277), it does not longer belong in apiextensions

### DIFF
--- a/pkg/apis/provider/v1alpha1/azure_types.go
+++ b/pkg/apis/provider/v1alpha1/azure_types.go
@@ -96,6 +96,7 @@ type AzureConfigSpecAzureVirtualNetwork struct {
 	// CIDR is the CIDR for the Virtual Network.
 	CIDR string `json:"cidr" yaml:"cidr"`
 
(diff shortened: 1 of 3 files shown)


pkg/apis/provider/v1alpha1/azure_types.go
```diff
@@ -96,10 +96,9 @@ type AzureConfigSpecAzureVirtualNetwork struct {
 	// CIDR is the CIDR for the Virtual Network.
 	CIDR string `json:"cidr" yaml:"cidr"`
 
+	// TODO: remove Master, Worker and Calico subnet cidr after v2 is deleted.
```
kopiczko: We should be clear that `azure-operator v2` is deleted. `cluster-operator` may also be involved here and hence confusion.

kopiczko (approved):

TheoBrigitte:
Oops this went straight into master, but comment is updated.

PR #98, opened 2018-07-03
Add values present in cluster-service storage but not in cluster config CRDs

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/3258

We want to migrate cluster-service to use the clusterconfig CRs rather than the current cluster storage implementation. This PR adds missing values we need to set to be able to get all data from the CRs.


pkg/apis/core/v1alpha1/aws_cluster_types.go
```diff
@@ -62,6 +62,7 @@ type AWSClusterConfigSpecGuest struct {
 	ClusterGuestConfig `json:",inline" yaml:",inline"`
 	CredentialSecret   AWSClusterConfigSpecGuestCredentialSecret `json:"credentialSecret" yaml:"credentialSecret"`
 	Masters            []AWSClusterConfigSpecGuestMaster         `json:"masters,omitempty" yaml:"masters,omitempty"`
+	ResourceTags       map[string]string                         `json:"resourceTags,omitempty" yaml:"resourceTags,omitempty"`
```
rossf7: ResourceTags are only for AWS currently.

pkg/apis/core/v1alpha1/aws_cluster_types.go
```diff
@@ -82,8 +83,11 @@ type AWSClusterConfigSpecGuestWorker struct {
 	Labels                        map[string]string `json:"labels" yaml:"labels"`
 }
 type AWSClusterConfigSpecGuestNode struct {
-	ID           string `json:"id" yaml:"id"`
-	InstanceType string `json:"instanceType,omitempty" yaml:"instanceType,omitempty"`
+	ID            string  `json:"id" yaml:"id"`
+	CPUCores      int     `json:"cpuCores,omitempty" yaml:"cpuCores,omitempty"`
+	InstanceType  string  `json:"instanceType,omitempty" yaml:"instanceType,omitempty"`
+	MemorySizeGB  float64 `json:"memorySizeGB,omitempty" yaml:"memorySizeGB,omitempty"`
+	StorageSizeGB float64 `json:"storageSizeGB,omitempty" yaml:"storageSizeGB,omitempty"`
```
rossf7: These node capabilities were only present for KVM. They also need to be set for AWS and Azure because they are output by `gsctl show cluster`.

pkg/apis/core/v1alpha1/cluster_guest_types.go
```diff
@@ -9,6 +9,7 @@ type ClusterGuestConfig struct {
 	ID             string                            `json:"id" yaml:"id"`
 	Name           string                            `json:"name,omitempty" yaml:"name,omitempty"`
 	Owner          string                            `json:"owner,omitempty" yaml:"owner,omitempty"`
+	ReleaseVersion string                            `json:"releaseVersion,omitempty" yaml:"releaseVersion,omitempty"`
```
rossf7: ReleaseVersion also needs to be set. Currently we just have the version bundles.

pkg/apis/core/v1alpha1/aws_cluster_types.go
rossf7: Yes I wasn't sure on this. These are just look ups based on instance type or VM size. A general GuestNode type would avoid the duplicated fields.

I'll update the PR to do that.

pkg/apis/core/v1alpha1/aws_cluster_types.go
rossf7: AFAIK they are AWS resource tags for cost tracking. This feature was added to the API a long time ago but didn't get implemented in aws-operator yet. 

I think we need to check whether this will get implemented. We could remove from cluster-service but keep in the API to avoid breaking the API spec.

tuommaki: I think worker labels are those referred in API spec. There is an issue to track this: https://github.com/giantswarm/giantswarm/issues/3415

Regarding cost tracking, I'm not sure though if this is enough, but maybe. In a way it makes sense to track resources that scale based on workload when everything else is static (cost wise).

kopiczko: > tags for the cluster, organisation and installation

I think all that fields are in the CR already. Why aws-operator can't take them from there?

pkg/apis/core/v1alpha1/cluster_guest_types.go
kopiczko: What is `ReleaseVersion`?

tuommaki: Version of our release bundle, i.e. version of `IndexRelease`.

pkg/apis/core/v1alpha1/cluster_guest_types.go
```diff
@@ -9,9 +9,16 @@ type ClusterGuestConfig struct {
...
 	VersionBundles []ClusterGuestConfigVersionBundle `json:"versionBundles,omitempty" yaml:"versionBundles,omitempty"`
 }
 
+type ClusterGuestConfigNode struct {
+	CPUCores      int     `json:"cpuCores,omitempty" yaml:"cpuCores,omitempty"`
+	MemorySizeGB  float64 `json:"memorySizeGB,omitempty" yaml:"memorySizeGB,omitempty"`
+	StorageSizeGB float64 `json:"storageSizeGB,omitempty" yaml:"storageSizeGB,omitempty"`
+}
```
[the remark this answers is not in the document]
tuommaki: When this matches fields in `KVMClusterConfigSpecGuestNode`, would it make sense to use this in there too? If not, I'd personally lean towards dropping this from AWS/Azure guest types and just making map lookup on the fly when reading cluster state in `cluster-service`. WDYT?

pkg/apis/core/v1alpha1/aws_cluster_types.go
kopiczko: Cool

kopiczko (approved):
LGTM. We don't support having different node types in the same cluster, but I don't think it can be dealt with in this PR.

pkg/apis/core/v1alpha1/cluster_guest_types.go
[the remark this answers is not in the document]
rossf7: I don't think using the same struct makes sense. On KVM the cpu, memory and storage depend on the underlying properties of the hardware.

Whereas on AWS and Azure they depend on the instance type or VM size and are set by the cloud provider. I'll remove and lets look them up on the fly. That's a better idea, thanks!

rossf7:
Closing in favour of #106

PR #100, opened 2018-07-06
adds version list management for convenience

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918. See also https://github.com/giantswarm/statusresource/pull/4.


pkg/apis/provider/v1alpha1/status_types.go
```diff
@@ -39,8 +52,14 @@ type StatusClusterNetwork struct {
...
 
+// StatusClusterNode holds information about a guest cluster node.
+type StatusClusterNode struct {
+	Name    string `json:"name" yaml:"name"`
+	Version string `json:"version" yaml:"version"`
+}
+
 // StatusClusterVersion expresses the versions in which a guest cluster was and
-// is.
+// may still is.
```
oponder: this is grammatically not right (may still be)

pkg/apis/provider/v1alpha1/status_funcs.go
```diff
@@ -90,3 +99,25 @@ func withCondition(conditions []StatusClusterCondition, search string, replace s
 
 	return newConditions
 }
+
+func withVersion(versions []StatusClusterVersion, date time.Time, version string, limit int) []StatusClusterVersion {
```
oponder: This function could use a comment describing the parameters. I'm not sure what this does right now.

oponder (commented):
LGTM, if you address my comments.

In general I think your PR's could use some introduction besides just a link to a epic.

Explaining what this PR intends to do, and more importantly why.

I get that this adds version list management, and I get that would be convenient. But convenient for who? where is it intended to be used? A sentence or two about that would help.

pkg/apis/provider/v1alpha1/status_funcs.go
xh3b4sd: I think the tests I wrote tell a way better story on the functionality than any comment could express. When you need a comment anyway let me know. I just thought the tests are a good use of documentation already.

pkg/apis/provider/v1alpha1/status_types.go
xh3b4sd: Cool thing. Fixed.

pkg/apis/provider/v1alpha1/status_funcs.go
oponder: The tests are nice! But it's not totally clear about what exactly this param will do. Have to read deeper into it before I understand, and I'm still not 100% of your intention. 

Particularly the `date` param.

So I do think this needs a comment explaining intention, or better naming of params and function.

When I read this I see a function called `withVersion` that takes a bunch of versions, a date, a target version, and a limit, and then returns some version.

So I assume it does some filtering or limiting. But the naming here `withVersion` doesn't really clue me in.

xh3b4sd:
You are right I did not really explain anything here. I messed up wiring the PRs. Like the operator PRs you already approved. In the broader scope this goes towards the update story. It also goes towards the cluster state story. We do lots of preparations here for this because the roadmap stories overlap in certain areas. There is then https://github.com/giantswarm/giantswarm/issues/3663, which is about **delegation and propagation**. This concept is something we want to get straight and establish more since it is pretty important for the overall system design of our operators. My goal is to have the actual version bundle versions we used to provision a specific guest cluster node available in the status of the provider config CR. This status magic thing is done by the `statusresource` implementation. We need the versions to reason about in which state the guest cluster is so we can use these information to implement draining to make upgrades in Azure graceful. We also want to use these version information in the CR status to express cluster states. That way we can say the guest cluster is `Updating` or `Updated`. The cluster states can be used to write generic and proper e2e tests to verify upgrades. You see this is a bit yak shavy and I am grinding through somehow.

pkg/apis/provider/v1alpha1/status_funcs.go
xh3b4sd: Please have a look again.

oponder (approved):
Ok, I get what WithNewVersion does now kinda.

I think it might do a bit too much for what its name implies. But maybe in the context of this domain concern it makes sense.

PR #104, opened 2018-07-10
adds drainer config based on node config

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918. Also see https://github.com/giantswarm/giantswarm/issues/3663.


pkg/apis/core/v1alpha1/drainer_funcs.go
```diff
@@ -0,0 +1,23 @@
+package v1alpha1
+
+const (
+	DrainerConfigStatusStatusTrue  = "True"
+	DrainerConfigStatusTypeDrained = "Drained"
```
xh3b4sd: I copied all of this from the `NodeConfig` that already exists. The idea is to move the draining to dedicated draining types. Then we can use the `NodeConfig` for node specific management and status updates. We just have to do the renaming and wire the operators accordingly.

pkg/apis/core/v1alpha1/drainer_types.go
```diff
@@ -0,0 +1,126 @@
...
+//     spec:
+//       group: core.giantswarm.io
+//       scope: Namespaced
+//       version: v1alpha1
+//       names:
+//         kind: DrainerConfig
+//         plural: drainerconfigs
+//         singular: drainerconfig
+//       subresources:
+//         status: {}
```
xh3b4sd: What I changed though is to add the status sub resource to the new `DrainerConfig` so we use the status as it should be.

rossf7 (approved):
Not deep into the draining changes but this LGTM and I like that we're using the status resource more.

kopiczko (approved):

PR #105, opened 2018-07-11
Add Network status to provider CRDs

tuommaki:
In order to move IPAM implementation to provider specific operators,
part of the state maintenance is to store information about allocated
subnet for specific guest cluster.

Adding Network information to provider types' status struct for this
purpose.


tuommaki:
Towards https://github.com/giantswarm/giantswarm/issues/2148

pkg/apis/provider/v1alpha1/status_types.go
```diff
@@ -32,6 +33,14 @@ type StatusClusterCondition struct {
 	Type string `json:"type" yaml:"type"`
 }
 
+// StatusClusterNetwork expresses the network segment that is allocated for a
+// guest cluster.
+type StatusClusterNetwork struct {
+	CIDR              string `json:"cidr" yaml:"cidr"`
+	PrivateSubnetCIDR string `json:"privateSubnetCidr" yaml:"privateSubnetCidr"`
+	PublicSubnetCIDR  string `json:"publicSubnetCidr" yaml:"publicSubnetCidr"`
```
kopiczko: This won't be generic, we have totally different subnets in azure. Also subnets don't really contribute to IPAM allocation, do they?

xh3b4sd: So what is the common ground then between all the providers?

tuommaki: > we have totally different subnets in azure

How do they look like? There must be top level network segment for the guest cluster that is optionally splitted, right?

I am aware of the reserved subnets in Azure, but I believe this is a property that should be configured to operator and not be present in CRD, if this is what you meant.

> Also subnets don't really contribute to IPAM allocation, do they?

Can you elaborate here what do you mean?

kopiczko: I mean only network CIDR is important. That's what we need to allocate. Correct me if I'm wrong but subnets are just the subnets of the network which is already allocated.

For azure there are:

- calico subnet
- master subnet
- worker subnet
- gateway subnet

But I don't think it's important because as I said before they are just subnets.

tuommaki: > only network CIDR is important

True, though I'm not sure if there's need for cross-resource communication of those other subnets. I'll take iterative approach here and drop them now. Lets add them if needed later.

kopiczko: Yes. I find the selected subnets pretty important to know. I wouldn't be surprised though if we decide to add them in provider specific CR statuses. But iterative approach makes sense here.

kopiczko (approved):

PR #106, opened 2018-07-11
Add ReleaseVersion to cluster guest types

rossf7:
Towards giantswarm/giantswarm#3258

The previous PR got borked and we only need to add a single field. So I've created a fresh PR.


tuommaki (approved):
Yes!

PR #107, opened 2018-07-11
removes status subresource

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918. So due to the lack of the enabled support for the status sub resources on certain installations we have to step back here and do it without, just like the node config magic is done now. I am not willing to bother with this more now. I tested and this seems to do the trick.


rossf7 (commented):
LGTM. A shame we can't use it now but we can use it again in the future.

rossf7 (approved):

PR #108, opened 2018-07-12
Add kvmconfig chart

r7vme:


xh3b4sd:
I think this may got stale. Closing.

PR #110, opened 2018-07-13
Add chart operator e2e chart

MarcelMue:
This is mostly copied over from `chart-operator` here: https://github.com/giantswarm/chart-operator/tree/master/integration/templates
I decided to move this here to align with other operators and to make it possible to reuse this is `cluster` and `release-operator` without importing the `chart-operator`.

towards https://github.com/giantswarm/giantswarm/issues/3390

### DIFF
--- a/helm/apiextensions-chart-config-e2e-chart/templates/chartconfig.yaml
+++ b/helm/apiextensions-chart-config-e2e-chart/templates/chartconfig.yaml
@@ -1,0 +1,21 @@
+apiVersion: "core.giantswarm.io/v1alpha1"
+kind: ChartConfig
+metadata:
+  name: "{{ .Values.chart.name }}"
+  namespace: "{{ .Values.namespace }}"
+spec:
+  chart:
+    channel: "{{ .Values.chart.channel }}"
+    configMap:
+      name: "{{ .Values.chart.configMap.name }}"
+      namespace: "{{ .Values.chart.configMap.namespace }}"
+      resourceVersion: "{{ .Values.chart.configMap.resourceVersion }}"
+    name: "{{ .Values.chart.name }}"
+    namespace: "{{ .Values.chart.namespace }}"
+    release: "{{ .Values.chart.release }}"
+    secret:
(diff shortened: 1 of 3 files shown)


helm/apiextensions-chart-config-e2e-chart/templates/chartconfig.yaml
```diff
@@ -0,0 +1,21 @@
...
+      name: "{{ .Values.chart.configMap.name }}"
+      namespace: "{{ .Values.chart.configMap.namespace }}"
+      resourceVersion: "{{ .Values.chart.configMap.resourceVersion }}"
+    name: "{{ .Values.chart.name }}"
+    namespace: "{{ .Values.chart.namespace }}"
+    release: "{{ .Values.chart.release }}"
+    secret:
+      name: "{{ .Values.chart.secret.name }}"
+      namespace: "{{ .Values.chart.secret.namespace }}"
+      resourceVersion: "{{ .Values.chart.secret.resourceVersion }}"
```
kopiczko: Aren't defaults for `configMap` and `secret` required in `values.yaml`?

MarcelMue: No, in the `cluster-operator`-> `chart-operator`  flow we don't set these values at all because they can be completely empty if a chart does not need any config.

kopiczko: https://docs.helm.sh/chart_best_practices/#flat-or-nested-values

Shouldn't we do something like 

```
  {{ if .Values.chart.secret -}}
    secret:
      name: "{{ .Values.chart.secret.name }}"
      namespace: "{{ .Values.chart.secret.namespace }}"
      resourceVersion: "{{ .Values.chart.secret.resourceVersion }}"
  {{- end -}} 
```

fgimenez (approved):
LGTM <3

kopiczko (approved):

PR #111, opened 2018-07-13
drops unused structure from node config

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918. We use the `DrainerConfig` now for draining. The `NodeConfig` is not used anymore. The `cluster-operator` will emit it for each guest cluster so the `node-operator` can act on behalf of it and manage the node status accordingly. 

screen shot 2018-07-13 at 23 56 07

See also https://docs.google.com/presentation/d/193WiI1HBhVlYezwsd5mme-9F3q47V0TgJc0kPcLdCIM/edit?usp=sharing and its comments to the diagram.


oponder (approved):

pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -55,40 +55,14 @@ type NodeConfig struct {
...
-	Name string `json:"name" yaml:"name"`
-}
-
-type NodeConfigSpecVersionBundle struct {
-	Version string `json:"version" yaml:"version"`
+type NodeConfigSpecPatch struct {
+	// SelfLink is the self link of the provider config CR. This is used by the
+	// node-operator to patch the status of the provider config CR. Without the
+	// self link the node-operator would not know which runtime object to patch.
+	SelfLink string `json:"selfLink" yaml:"selfLink"`
```
kopiczko: It's `.Spec.Patch.SelfLink`? This isn't meaningful name IMO.
I'd prefer something that indicates that it's link to the actual provider specific CR. SelfLink is just a name in the CR metadata which here makes little sense. I was confused with actual SelfLink in NodeConfig and wondered why to repeat it here.
If you need inspiration `Spec.ProviderCR.Link` would work for me. Plus good godoc like you have here.

xh3b4sd:
This is outdated and has to be sorted in a different way.

PR #112, opened 2018-07-16
Fix chartconfig templating

MarcelMue:
Today I learned that the `-` inside helm templating change stuff and whitespaces are fun.

towards https://github.com/giantswarm/giantswarm/issues/3390


kopiczko (approved):
LGTM when tested

PR #114, opened 2018-07-18
adds status structure for node versions to NodeConfig

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1918. Also see https://github.com/giantswarm/node-operator/pull/57. Note that the structure of the `NodeConfig` will further change. Once older clusters moved to newer versions we can drop the old `NodeConfig` structures and only use it for the status magic and its further purposes.


pkg/apis/core/v1alpha1/node_types.go
```diff
@@ -103,6 +104,15 @@ type NodeConfigStatusCondition struct {
...
 }
 
+// NodeConfigStatusVersion expresses the version in which a guest cluster node
+// is.
+type NodeConfigStatusVersion struct {
+	// Name is the node name, e.g. al9qy-worker-000001.
+	Name string `json:"name" yaml:"name"`
+	// Semver is some semver version, e.g. 1.0.0.
+	Semver string `json:"semver" yaml:"semver"`
+}
```
kopiczko: This doesn't read nice to me:

```yaml
status:
  versions:
  - name: al9qy-worker-000001
    semver: 1.0.0
  - name: al9qy-worker-000002
    semver: 1.0.0
```

Looks like our versions have names. I think intention here is to denote what versions our nodes have. Which I would expect to be something like:

```yaml
status:
  nodes:
  - name: al9qy-worker-000001
    version: 1.0.0
  - name: al9qy-worker-000002
    version: 1.0.0
```

rossf7 (approved):
LGTM

TheoBrigitte (commented):
LGTM, when Mr Pawel is happy

xh3b4sd:
I am not sure about this right now and the PR is outdated. Closing.

PR #116, opened 2018-07-24
provider/v1alpha1: add .spec.cluster.baseDomain; deprecate aws DNS related types

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/3861


pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -65,68 +65,66 @@ type AWSConfigSpec struct {
...
+	CredentialSecret CredentialSecret `json:"credentialSecret" yaml:"credentialSecret"`
+
+	Masters []AWSConfigSpecAWSNode `json:"masters" yaml:"masters"`
+	Region  string                 `json:"region" yaml:"region"`
+	VPC     AWSConfigSpecAWSVPC    `json:"vpc" yaml:"vpc"`
+	Workers []AWSConfigSpecAWSNode `json:"workers" yaml:"workers"`
+
+	// TODO delete with: aws-operator@v13
+	// Delete fields below with its types. They are unused from
+	// aws-operator@v14.
```
kopiczko: Should be 

```
// TODO delete with: aws-operator@v14
	// Delete fields below with its types. They are unused from
	// aws-operator@v15.
```

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -65,68 +65,66 @@ type AWSConfigSpec struct {
...
 
-// AWSConfigSpecAWSIngress deprecated since aws-operator v12 resources.
+// AWSConfigSpecAWSIngress deprecated since aws-operator v14 resources.
 type AWSConfigSpecAWSIngress struct {
 	HostedZones string                     `json:"hostedZones" yaml:"hostedZones"`
 	ELB         AWSConfigSpecAWSIngressELB `json:"elb" yaml:"elb"`
 }
 
-// AWSConfigSpecAWSIngressELB deprecated since aws-operator v12 resources.
+// AWSConfigSpecAWSIngressELB deprecated since aws-operator v14 resources.
```
kopiczko: All those should be `v15`

pkg/apis/provider/v1alpha1/cluster_types.go
```diff
@@ -5,6 +5,16 @@ import (
...
+	// BaseDomain is the root zone domain. The operator creates
+	// "{{ .ID }}.k8s.{{ .BaseDomain }}" DNS zone for the cluster and then
+	// create delegation to the newly created zone from the root zone. The
+	// newly created zone contains following records:
+	//
+	//	- ALIAS api.{{ .ID }}.k8s.{{ .BaseDomain }}
+	//	- CNAME etcd.{{ .ID }}.k8s.{{ .BaseDomain }}
+	//	- ALIAS ingress.{{ .ID }}.k8s.{{ .BaseDomain }}
+	//	- CNAME *.{{ .ID }}.k8s.{{ .BaseDomain }}
+	BaseDomain string            `json:"baseDomain" yaml:"baseDomain"`
```
kopiczko: This is the main change. This is what should be used in the newest versions of the operators.

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -65,68 +65,67 @@ type AWSConfigSpec struct {
 }
 
 type AWSConfigSpecAWS struct {
```
kopiczko: Here are only deprecation notes. Normally I should do that after updating aws-operator, but I made it beforehand to make my intentions clearer and easier to understand. There are also Azure and KVM deprecation notes to make. Full list of tasks here https://github.com/giantswarm/giantswarm/issues/3861

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -65,68 +65,67 @@ type AWSConfigSpec struct {
...
+	Masters []AWSConfigSpecAWSNode `json:"masters" yaml:"masters"`
+	Region  string                 `json:"region" yaml:"region"`
+	VPC     AWSConfigSpecAWSVPC    `json:"vpc" yaml:"vpc"`
+	Workers []AWSConfigSpecAWSNode `json:"workers" yaml:"workers"`
+
+	// TODO delete with: aws-operator@v15
+	// Delete fields below with its types. They are unused from
+	// aws-operator@v15.
+	//
+	// Issue: https://github.com/giantswarm/giantswarm/issues/3862
```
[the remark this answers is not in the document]
xh3b4sd: Above it still says this. 

```
TODO delete with: aws-operator@v15
```

It is also weird that something we introduce in 15 has to be removed in 14. How does that work?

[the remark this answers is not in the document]
kopiczko: Yes this was mind bending for me yesterday as you can see here. I mixed things up.
The idea is to make sure to stop using this values with v15, which means we can delete that with v14. And TBH it's hard to say when they were introduced. So I will write `Introduced with: ???` and `Remove with: aws-operator@v14`. Does that make sense?

Thanks for helping with making that straight <3

xh3b4sd (approved):
Would be cool to get a clearer comment on the TODO magic. Fine with me otherwise.

pkg/apis/provider/v1alpha1/cluster_types.go
TheoBrigitte: This is not so easy to understand, I think more verbosity and details are needed. 
Correct me if I'm wrong:
`BaseDomain` is the host cluster zone with NS record (delegation) to guest cluster zone `{{ .ID }}.k8s.{{ .BaseDomain }}`

kopiczko: This is the value from there https://github.com/giantswarm/installations/blob/80d92c9a87d3726cf37771d8dfdbc49809824432/gauss/cluster.yaml#L1

I'm going to align it in the cluster-operator. Tracked here https://github.com/giantswarm/giantswarm/issues/3861

What details you would need to understand it better?

TheoBrigitte: I got confused here because the comment metions DNS zone, records, delegation which are architecture specific. I recall seeing good comments about this on the aws-operator implementation. So I think we can keep it simple and high level here by only describing what does `baseDomain` and other sub domains looks like.
Something like:
```
	// BaseDomain is the root zone domain.
        // e.g. gauss.eu-central-1.aws.gigantic.io
        //
        // It is used to build guest cluster domains:
        // {{ .ID }}.k8s.{{ .BaseDomain }}
        // api.{{ .ID }}.k8s.{{ .BaseDomain }}
        // etcd.{{ .ID }}.k8s.{{ .BaseDomain }}
        // ingress.{{ .ID }}.k8s.{{ .BaseDomain }}
        // *.{{ .ID }}.k8s.{{ .BaseDomain }}
```

kopiczko:
This is stale. I will get back to it after we have release-operator.

PR #117, opened 2018-08-02
extends cluster status types with deleted and deleting status

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2209.

### DIFF
--- a/pkg/apis/provider/v1alpha1/status_funcs_test.go
+++ b/pkg/apis/provider/v1alpha1/status_funcs_test.go
@@ -98,6 +98,57 @@ func Test_Provider_Status_LatestVersion(t *testing.T) {
 	}
 }
 
+func Test_Provider_Status_withCondition(t *testing.T) {
+	testCases := []struct {
+		Name               string
+		Conditions         []StatusClusterCondition
+		Search             string
+		Replace            string
+		Status             string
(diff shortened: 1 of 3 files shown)


pkg/apis/provider/v1alpha1/status_funcs_test.go
```diff
@@ -98,6 +98,57 @@ func Test_Provider_Status_LatestVersion(t *testing.T) {
 	}
 }
 
+func Test_Provider_Status_withCondition(t *testing.T) {
```
xh3b4sd: Starting some first tests. I plan to extend the design of `withCondition` later on in separate steps to make it more intuitive and reliable.

pkg/apis/provider/v1alpha1/status_types.go
```diff
@@ -16,6 +16,11 @@ const (
 	StatusClusterTypeCreating = "Creating"
 )
 
+const (
+	StatusClusterTypeDeleted  = "Deleted"
+	StatusClusterTypeDeleting = "Deleting"
+)
```
xh3b4sd: Going to manage these in the `statusresource`.

kopiczko (approved):

PR #118, opened 2018-08-03
adds support for resource specific status conditions

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/3908.


pkg/apis/provider/v1alpha1/status_types.go
```diff
@@ -63,6 +66,25 @@ type StatusClusterNode struct {
...
 
+// Resource is structure holding arbitrary conditions of operatorkit resource
+// implementations. Imagine an operator implements an instance resource. This
+// resource may operates sequentially but has to operate based on a certain
+// system state it manages. So it tracks the status as needed here specific to
+// its own implementation and means in order to fulfil its premise.
+type StatusClusterResource struct {
+	Conditions []StatusClusterResourceCondition `json:"conditions" yaml:"conditions"`
+	Name       string                           `json:"status" yaml:"status"`
+}
```
xh3b4sd: I am thinking about to address the issue we have in Azure with such a resource condition. 

```
{
    "conditions": [
        {
            "status": "False",
            "type": "LatestARMTemplatesUploaded"
        }
    ],
    "name": "instancev3"
}
```

kopiczko (approved):
If that solves the problem let's go with it. What worries me is this information isn't very useful for the user.

xh3b4sd:
What information do you mean? I think I am coming back to you on Monday. Need a bit of a brainstorm eventually. Lets see.

kopiczko:
I mean this isn't useful for humans reading the CR. It's just a storage for the resource AIUI. If it isn't possible to rely on the provider API to get that information I think that's ok. But we should try hard to not use that.

Unfortunately I'm afk till Tuesday next week. I'll try to work when I have access to internet but it isn't guaranteed.

xh3b4sd:
I see. Yes the resource states are quite internal but in Azure we seem to need this kind of "storage" in some cases now.

PR #119, opened 2018-08-07
Add port mappings into kvmconfig

r7vme:
part-of: https://github.com/giantswarm/ingress-operator/issues/96

related: https://github.com/giantswarm/kvm-operator/pull/525

# Why?

See the issue above. In short, we want to use kubernetes default `nodePort` to expose TCP ports from guest vms instead of current implmentation that uses nginx-ingress-controller tcp piping for that.

Here is example of kvmconfig part with new field "portMappings".
```
  kvm:
    portMappings:
    - name: ingress-http
      targetPort: 30010
      nodePort: 30083
    - name: ingress-https
      targetPort: 30011
      nodePort: 30084
    endpointUpdater:
      docker:
        image: quay.io/giantswarm/k8s-endpoint-updater:d92b6f452add9c4694b27141cf3c2578de1442f1
```


pkg/apis/provider/v1alpha1/kvm_types.go
```diff
@@ -71,6 +71,7 @@ type KVMConfigSpecKVM struct {
 	Network         KVMConfigSpecKVMNetwork         `json:"network" yaml:"network"`
 	// NOTE THIS IS DEPRECATED
 	NodeController KVMConfigSpecKVMNodeController `json:"nodeController" yaml:"nodeController"`
+	PortMappings   []KVMConfigSpecKVMPortMappings `json:"portMappings yaml:"portMappings"`
```
r7vme: I never did changes in this repo before, please let me know if i'm missing anything.

xh3b4sd: You need to run `./scripts/gen.sh` from project root to generate the go clients. Then  you see if something changes or not.

r7vme: Thanks. Updated the PR with generated code.

xh3b4sd (approved):

PR #125, opened 2018-08-20
reduce-worker-count-kvm7

calvix:
`e2e-harness`  expect that cluster has only 2 workers as it hardcheck against 3 nodes (1 master + 2 workers) and if the number doesn't match exactly it will return error.

So setting KVM cluster to 2 workers instead of 3.


xh3b4sd (approved):

fgimenez (approved):
LGTM

PR #130, opened 2018-09-04
Add configuration option for docker volume size

tuommaki:
Allows configuration of volume size that is mounted to /var/lib/docker.


calvix (approved):

teemow:
Adding this to the node config feels wrong if this is something that is only configurable per cluster. But yeah the structure is somehow borked anyway. 

There is an array of workers that aren't managed individually anyway (because they are all in one ASG) and also not configurable individually in our interfaces.

PR #131, opened 2018-09-04
Add custom configmap to chartconfig

indyfree:
Towards part of: https://github.com/giantswarm/giantswarm/pull/4093

Adds a custom configmap that can be configured per cluster, to support custom configurations from customers.

### DIFF
--- a/pkg/apis/core/v1alpha1/chart_types.go
+++ b/pkg/apis/core/v1alpha1/chart_types.go
@@ -70,6 +70,9 @@ type ChartConfigSpecChart struct {
 	// ConfigMap references a config map containing values that should be
 	// applied to the chart.
 	ConfigMap ChartConfigSpecConfigMap `json:"configMap" yaml:"configMap"`
+	// CustomConfigMap references a config map containing custom values that should be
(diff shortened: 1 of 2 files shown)


pkg/apis/core/v1alpha1/chart_types.go
```diff
@@ -70,6 +70,9 @@ type ChartConfigSpecChart struct {
 	// ConfigMap references a config map containing values that should be
 	// applied to the chart.
 	ConfigMap ChartConfigSpecConfigMap `json:"configMap" yaml:"configMap"`
+	// CustomConfigMap references a config map containing custom values that should be
+	// applied to the chart.
```
rossf7: I think godoc should mention these are values overridden by the user.

rossf7 (approved):
LGTM

tuommaki (approved):
The naming of this field somehow bothers me, but I cannot really suggest anything better either. IIWFYILGTM.

indyfree:
Spoke with @ross and we decided that `UserConfigMap` is a better naming.

PR #133, opened 2018-09-11
fixes version label in KVM chart

xh3b4sd:
This is necessary to have the chart dynamic and relative to the tested version so e2e tests do test the right version at all. This is also necessary to make the e2e update tests work. Otherwise we never see the `Created` cluster status condition.


calvix (approved):

kopiczko:
Is it fine for corresponding AWS and Azure charts?

kopiczko:
I meant if we don't have something like `aws-operator.giantswarm.io/version=2.5.0` in other charts. I checked we don't.

xh3b4sd:
Oh, I commented on something else and noticed this here is a different PR. Sorry. :D I get your question now. All good.

PR #138, opened 2018-09-18
Add e2e chart for release

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/3313

also created the related quay repo: https://quay.io/application/giantswarm/apiextensions-release-e2e-chart

### DIFF
--- a/helm/apiextensions-release-e2e-chart/values.yaml
+++ b/helm/apiextensions-release-e2e-chart/values.yaml
@@ -1,0 +1,2 @@
+namespace: "giantswarm"
+versionBundleVersion: "0.1.0"
--- a/helm/apiextensions-release-e2e-chart/values.yaml
+++ b/helm/apiextensions-release-e2e-chart/values.yaml
@@ -1,2 +1,5 @@
 namespace: "giantswarm"
+operator:
+  name: "test-operator"
+  version: "0.1.0"
 versionBundleVersion: "0.1.0"
--- a/helm/apiextensions-release-e2e-chart/values.yaml
+++ b/helm/apiextensions-release-e2e-chart/values.yaml
@@ -2,4 +2,5 @@ namespace: "giantswarm"
 operator:
   name: "test-operator"
   version: "0.1.0"
-versionBundleVersion: "0.1.0"
+versionBundle:
+  version: "0.1.0"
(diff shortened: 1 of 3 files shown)


helm/apiextensions-release-e2e-chart/values.yaml
```diff
@@ -0,0 +1,2 @@
+namespace: "giantswarm"
+versionBundleVersion: "0.1.0"
```
xh3b4sd: I would like to have more structure here. See https://github.com/giantswarm/apiextensions/blob/6206a841872f34408a645afddd7bae4162984bf1/helm/apiextensions-kvm-config-e2e-chart/values.yaml.

xh3b4sd: Also, why is there no operator name and version there? How is that used?

MarcelMue: My idea here was to just set some defaults, AFAIK we overwrite these anyway in `e2etemplates` (like we did here https://github.com/giantswarm/e2etemplates/blob/master/pkg/e2etemplates/apiextensions_chart_config_e2e_chart_values.go ). This whole thing isn't super clear to me tbh, so I might be wrong.

xh3b4sd: I find it super weird to only set some things and assume others are set magically. We should always set everything? Don't know how this is supposed to be.

xh3b4sd: @tuommaki WDTY?

MarcelMue: I added some defaults PTAL
We basically do the same in all our helm charts in operators, we don't have any defaults set and expect them to be set through our deployment pipeline with draughtsman config.

You have a point though, that it isn't very nice in general.

helm/apiextensions-release-e2e-chart/values.yaml
```diff
@@ -0,0 +1,5 @@
+namespace: "giantswarm"
+operator:
+  name: "test-operator"
+  version: "0.1.0"
+versionBundleVersion: "0.1.0"
```
xh3b4sd: I still would like to see more structure here. I shared a link for an example. When you disagree, lets talk.

MarcelMue: Fixed this now, thinking about the defaulting too much and forgot to change this :man_facepalming:

xh3b4sd (approved):

helm/apiextensions-release-e2e-chart/values.yaml
tuommaki: I don't have strong opinions here. I think we need to experiment over time and if htere's pain, then fix appropriately.

PR #139, opened 2018-09-19
adds initial Cluster CRD and CR

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/3313.


pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+		},
+	}
+}
+
+func NewTypeMeta() metav1.TypeMeta {
+	return metav1.TypeMeta{
+		APIVersion: version,
+		Kind:       kindCluster,
+	}
+}
```
xh3b4sd: I missed this in several places so far. Just adding now because I think it is useful in certain places.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+//       scope: Namespaced
+//       version: v1alpha1
+//       names:
+//         kind: Cluster
+//         plural: clusters
+//         singular: cluster
+//       subresources:
+//         status: {}
+//
+func NewClusterCRD() *apiextensionsv1beta1.CustomResourceDefinition {
```
xh3b4sd: This is the usual magic we use for everything else. Only difference is the new and clean naming like we already did with the `Release` CR and the usage of a couple of constants internally.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+// request a tenant cluster creation by providing necessary configurations.
+// Fields here are either mandatory or optional. Optional fields left blank will
+// be filled with appropriate default values which are then propagated into the
+// CR status.
+type ClusterSpec struct {
+	// Cluster holds cluster specific configuration.
+	Cluster ClusterSpecCluster `json:"cluster" yaml:"cluster"`
+	// Release holds release specific configuration.
+	Release ClusterSpecRelease `json:"release" yaml:"release"`
+}
```
xh3b4sd: Since we aim to use CRs as public interface we could already make more use of some godoc explaining how things work. I do not claim completeness here. Just a first catch and whoever wants to improve can do it later in further iterations.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+
+// ClusterSpecCluster holds cluster specific configuration users provide.
+type ClusterSpecCluster struct {
+	// Description is the optional cluster description users can provide. If left
+	// blank a cluster description will be generated. The cluster description is
+	// propagated into the CR status.
+	Description string `json:"description" yaml:"description"`
```
xh3b4sd: This what we call cluster name right now. IMO name is wrong. People use it as a description field to describe a cluster using multiple words. I would like to take the opportunity for the CR to improve this. API spec in the `api` microservice can stay as it is. I would just map it accordingly. There is anyway a lot of api spec work we still have to do in the scope of the update story.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+// the tenant cluster description will be generated if left blank upon cluster
+// creation and made available here.
+type ClusterStatus struct {
+	// Cluster holds cluster specific status information.
+	Cluster ClusterStatusCluster `json:"cluster" yaml:"cluster"`
+	// Conditions is a list of status conditions.
+	Conditions []ClusterStatusCondition `json:"conditions" yaml:"conditions"`
+	// Release holds release specific status information.
+	Release ClusterStatusRelease `json:"release" yaml:"release"`
+}
```
xh3b4sd: As of this proposal there are certain fields present in the spec as well as in the status. The idea here is to provide the user interface with the spec. There are optional fields though, which we then generate automatically. The release version is such a candidate. We then propagate such fields from the spec to the status and set defaults if necessary.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+		Kind:       kindCluster,
+	}
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type Cluster struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
```
xh3b4sd: Below you will notice that I did not define any cluster ID. So far the ID is exclusively generated internally by us. I want to keep this for now. The `Cluster` CR will for now be created within our `api` microservice, where we need to generate the tenant cluster ID. We set the ID as the object meta data name. No need for having it redundantly in the spec as well.

xh3b4sd (commented):
The proposal here describes an idea on how to manage the new `Cluster` CR. In the first step we only track and manage `Description`, `ID`, `Organization` and `Version`. This is done by `Spec` and `Status` working together. More information below.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// ClusterSpec is the part of the interface available to users in order to
+// request a tenant cluster creation by providing necessary configurations.
+// Fields here are either mandatory or optional. Optional fields left blank will
+// be filled with appropriate default values which are then propagated into the
+// CR status.
+type ClusterSpec struct {
+	// Cluster holds cluster specific configuration.
+	Cluster ClusterSpecCluster `json:"cluster" yaml:"cluster"`
```
tuommaki: I feel this is a bit repetative: `Cluster.Spec.Cluster...` and same in yaml:
```yaml
apiVersion: v1alpha1
kind: Cluster
metadata:
  name: foobar
spec:
  cluster:
    description: "Cluster foobar"
    organization: giantswarm
  release:
    version: 5.1.1
```

Maybe remove extra level from here? It would then be more in line with upstream types (https://github.com/kubernetes/api/blob/master/core/v1/types.go). WDYT?

kopiczko: +1 on that.

kopiczko: I would also skip `release` part. I don't see what else can be added under this key.

```
apiVersion: v1alpha1
kind: Cluster
metadata:
  name: foobar
spec:
  description: "Cluster foobar"
  organization: giantswarm
  version: 5.1.1
```

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+	Spec              ClusterSpec   `json:"spec"`
+	Status            ClusterStatus `json:"status"`
+}
+
+// ClusterSpec is the part of the interface available to users in order to
+// request a tenant cluster creation by providing necessary configurations.
+// Fields here are either mandatory or optional. Optional fields left blank will
+// be filled with appropriate default values which are then propagated into the
+// CR status.
+type ClusterSpec struct {
```
kopiczko: I think `provider` will be a required field.

kopiczko (commented):
Godocs are very nice <3

pkg/apis/core/v1alpha1/cluster_types.go
xh3b4sd: Later yes. Not of interest now.

pkg/apis/core/v1alpha1/cluster_types.go
teemow: "Release holds release specific configuration" is a really useless sentence. How and why does this add any benefit?

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// Release holds release specific status information.
+	Release ClusterStatusRelease `json:"release" yaml:"release"`
+}
+
+// ClusterStatusCluster holds cluster specific status information.
+type ClusterStatusCluster struct {
+	// Description is the propagated cluster description users can provide or the
+	// system generates automatically if left blank.
+	Description string `json:"description" yaml:"description"`
+}
```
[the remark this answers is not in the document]
teemow: Yes I also think it doesn't make sense to distinguish desired and actual state of the description. It doesn't need to be propagated elsewhere.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,148 @@
...
+				Singular: "cluster",
+			},
+			Subresources: &apiextensionsv1beta1.CustomResourceSubresources{
+				Status: &apiextensionsv1beta1.CustomResourceSubresourceStatus{},
+			},
+		},
+	}
+}
+
+func NewTypeMeta() metav1.TypeMeta {
```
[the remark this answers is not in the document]
xh3b4sd: Good shout. Done.

pkg/apis/core/v1alpha1/cluster_types.go
xh3b4sd: Good input. Done. Also did it with the status, but there we have an extra level with conditions so I kept the cluster block there.

pkg/apis/core/v1alpha1/cluster_types.go
xh3b4sd: I removed it already. The guys thought this extra level is not cool anyway. Please check again.

pkg/apis/core/v1alpha1/cluster_types.go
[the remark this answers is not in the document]
xh3b4sd: Then we have to talk about the problem at hand. Because not doing it makes no sense to me. Lets better hang about this

puja108 (commented):
I believe this is a good effort and a step forward to have more user-friendly CRs without having to change much of our internal ways of doing things. As such this LGTM.

Two points we should consider for future steps:
1. Can we re-use the upstream defined cluster object?
2. Could we rethink our concepts to be more K8s native?

For 2 I'm specifically thinking of:
- Can org be a namespace instead?
- Can we get rid of ClusterID and have the user give the cluster a name just like with any K8s object and the only requirement is that it is unique within the namespace? (especially if orgs are namespaces)

teemow:
@puja108 we had a longer chat about this and i think this would be good to discuss in the umbrella issue. Marcel already added a short summary of our discussion:

https://github.com/giantswarm/giantswarm/issues/4135#issuecomment-423106779

For now we need to keep it and thats why we move the generation from cluster-service to the api service itself.

puja108:
Copied my comment there. And I agree that for now it's ok to move and keep things.

pkg/apis/core/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,130 @@
...
+	// blank a cluster description will be generated. The cluster description is
+	// propagated into the CR status.
+	Description string `json:"description" yaml:"description"`
+	// Organization is the mandatory cluster organization in which a tenant
+	// cluster will be scoped into.
+	Organization string `json:"organization" yaml:"organization"`
+	// Version is the optional release version users can provide. If left blank
+	// the current default release version will be used. The release version is
+	// propagated into the CR status.
+	Version string `json:"version" yaml:"version"`
```
puja108: I just realized that we have a bit of an issue here, as our release version is provider specifc, but this new object is supposed to be generic. I have no good idea how to fix this, @xh3b4sd did you have some thoughts around this in your team? You seem to also have kind of a split in the thinking here, as you mention both a version field, as well as defaulting and well as propagation into status field.

xh3b4sd: For the record. We talked and agreed that we have to use the status for certain defaults because of the current lack of mutating admission controllers and the current state of our infrastructure. We are going into this direction anyway and will work on certain technical debt down the road.

xh3b4sd:
Fishing for several approvals. I think we agree on going with this for now and iterate.

MarcelMue (approved):
We talked about this quite extensively, we should get going like this and learn / improve as we go.

kopiczko (approved):

PR #140, opened 2018-09-21
modifies Release CR to match GS release structure

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/3313.


marians:
Can you provide a YAML representation of an example CR? That would help me reviewing a great deal.

xh3b4sd:
You saw the test buddy? JSON or YAML? What do you prefer?

xh3b4sd:
I mean, what is missing? You want to have a fully fletched CR as godoc or what?

marians:
I didn't see the PR. What I'd like would be: An entire YAML representation of this CR pasted into one of the comments, like in the main comment:

```yaml
some: yaml
key: value
```

marians:
In the godoc I see a YAML representation of the `CustomResourceDefinition`. What I'd like: YAML of an example `Release` CR.

xh3b4sd:
Please have another look.

pkg/apis/core/v1alpha1/release_types.go
```diff
@@ -59,13 +103,18 @@ type Release struct {
 }
 
 type ReleaseSpec struct {
-	Operator      ReleaseSpecOperator      `json:"operator" yaml:"operator"`
+	Active        bool                     `json:"active" yaml:"active"`
+	Authorities   []ReleaseSpecAuthority   `json:"authorities" yaml:"authorities"`
+	Date          DeepCopyTime             `json:"date" yaml:"date"`
+	Version       string                   `json:"node" yaml:"version"`
```
marians: `json:"node"` but `yaml:"version"` - that doesn't look as if that's intended

pkg/apis/core/v1alpha1/release_types.go
```diff
@@ -59,13 +103,18 @@ type Release struct {
 }
 
 type ReleaseSpec struct {
-	Operator      ReleaseSpecOperator      `json:"operator" yaml:"operator"`
+	Active        bool                     `json:"active" yaml:"active"`
```
MarcelMue: Would it make sense for us to add the WIP flag here as well now? That way we have it here already.

pkg/apis/core/v1alpha1/release_types.go
```diff
@@ -48,9 +52,49 @@ func NewReleaseCRD() *apiextensionsv1beta1.CustomResourceDefinition {
...
+//         provider: azure
+//         version: 0.7.0
+//       date: "0001-01-01T00:00:00Z"
+//       version: "2.0.0"
+//       versionBundle:
+//         version: "0.1.0"
+//     status:
+//       conditions:
+//       - status: True
+//         type: InUse
```
marians: According to a comment in the `ReleaseStatusCondition` definition, valid values here may be `Pending, Ready, Draining, Drained`. That comment should be fixed.

marians:
That's a big step ahead. So the CR can look like

```
typemeta:
  apiversion: "v1alpha1"
  kind: "Release"
objectmeta:
  name: "2.0.0"
spec:
  active: false
  date: "0001-01-01T00:00:00Z"
  version: "2.0.0"
  versionBundle:
    version: "0.1.0"
  authorities:
  - endpoint: http://azure-operator:8000
    name: azure-operator
    version: 2.0.0
  - endpoint: http://cert-operator:8000
    name: cert-operator
    version: 0.1.0
  - endpoint: http://chart-operator:8000
    name: chart-operator
    version: 0.3.0
  - endpoint: http://cluster-operator:8000
    name: cluster-operator
    provider: azure
    version: 0.7.0
status:
  conditions:
  - status: True
    type: InUse
```

The introduced items seem okay to me. But I am lacking some background regarding whether the wording of attributes is ideal, or what alternatives there are.

With `type: InUse`, I am unsure what `InUse` means exactly, whether there could be a better name for that and what the other names are.

xh3b4sd:
Please don't bother too much with details on the status conditions. You wanted an example and I made up one. We are not having the `InUse` status condition. I just made it up for you because you asked for an example.

pkg/apis/core/v1alpha1/release_types.go
xh3b4sd: The comment was wrong. Copy/paste issue. Sorted.

pkg/apis/core/v1alpha1/release_types.go
xh3b4sd: Thanks. Done.

pkg/apis/core/v1alpha1/release_types.go
xh3b4sd: My thinking was to put more brainpower into it later. I would like to address these improvements separately. Lets please not do too many things at once.

MarcelMue (approved):
I am fine with this. Let's wait for PO / UX feedback and then go.

marians:
@xh3b4sd Then I suggest to replace `InUse` with `ExampleCondition` or something like that. Then nobody has to think about it.

pkg/apis/core/v1alpha1/release_types.go
```diff
@@ -48,9 +52,49 @@ func NewReleaseCRD() *apiextensionsv1beta1.CustomResourceDefinition {
...
+//         name: chart-operator
+//         version: 0.3.0
+//       - endpoint: http://cluster-operator:8000
+//         name: cluster-operator
+//         provider: azure
+//         version: 0.7.0
+//       date: "0001-01-01T00:00:00Z"
+//       version: "2.0.0"
+//       versionBundle:
+//         version: "0.1.0"
```
puja108: Can you clarify what this is? I understand there's GS release versions, which is both the name and the version field higher up. and there's versions of components (or here called authorities), but what is this version bundle version now? and why is it needed?

marians (approved):

pkg/apis/core/v1alpha1/release_types.go
```diff
@@ -48,9 +52,49 @@ func NewReleaseCRD() *apiextensionsv1beta1.CustomResourceDefinition {
...
+//
+//     typemeta:
+//       apiversion: "v1alpha1"
+//       kind: "Release"
+//     objectmeta:
+//       name: "2.0.0"
+//     spec:
+//       active: false
+//       authorities:
+//       - endpoint: http://azure-operator:8000
```
puja108: what is the reason we define endpoints here? especially as it looks like endpoints can be generated from the name field

xh3b4sd:
Makes it worse for me because the other information are rather specific as well, but ok. I changed the example condition.

pkg/apis/core/v1alpha1/release_types.go
```diff
@@ -48,9 +52,49 @@ func NewReleaseCRD() *apiextensionsv1beta1.CustomResourceDefinition {
...
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
+// Release represents a Giant Swarm release used to describe the managed
+// versiones of tenant clusters. This might look something like the following.
+//
+//     typemeta:
+//       apiversion: "v1alpha1"
+//       kind: "Release"
+//     objectmeta:
+//       name: "2.0.0"
```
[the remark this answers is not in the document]
xh3b4sd: Please outline the examples you have in mind. I cannot think of a single one. The releases live inside an installation and will never leave them. There will never be provider confusion within an installation because one installation is only about a single provider and all releases present are only relevant to this single provider. A release as such is also not interested in anything related to AWS or something and this should stay that way IMO.

pkg/apis/core/v1alpha1/release_types.go
xh3b4sd: That is basically the operator version which the `release-operator` in this case uses in order to route and consume the CR. Without this information the operator would not know which code path to apply. The name version bundle version is rather reasonless legacy and just the term/structure we use in every other CR out there.

puja108: Understood, ok with leaving it like this, but it sounds quite confusing, as we're playing with multiple release/version naming schemes now and rely on lots of tacit knowledge about GS.

pkg/apis/core/v1alpha1/release_types.go
```diff
@@ -59,13 +103,18 @@ type Release struct {
 }
 
 type ReleaseSpec struct {
-	Operator      ReleaseSpecOperator      `json:"operator" yaml:"operator"`
+	Active        bool                     `json:"active" yaml:"active"`
+	Authorities   []ReleaseSpecAuthority   `json:"authorities" yaml:"authorities"`
+	Date          DeepCopyTime             `json:"date" yaml:"date"`
+	Version       string                   `json:"version" yaml:"version"`
```
teemow: Compared to the releases in the installation repo the only thing that is missing is the provider. I've thought about this earlier this morning already and I somehow would prefer to have the provider in here too. 

First I thought that due to federation of control planes we could also have all releases in one control plane (because of federation or byoc etc). But then i realized that this doesn't even work because of possible version collisions. 

Quick questions to shed some more light:
* Would the release operator need to know on which provider it runs? Eg to select the right configmaps for the operator chartconfigs or so? If uncertain, where could the operator get the information from? 
* If the operator knows the provider is there any other reason why we could have releases from different providers on the same control plane?

pkg/apis/core/v1alpha1/release_types.go
```diff
@@ -59,13 +103,18 @@ type Release struct {
...
 	VersionBundle ReleaseSpecVersionBundle `json:"versionBundle" yaml:"versionBundle"`
 }
 
-type ReleaseSpecOperator struct {
-	Name    string `json:"cluster" yaml:"name"`
-	Version string `json:"node" yaml:"version"`
+type ReleaseSpecAuthority struct {
+	Endpoint DeepCopyURL `json:"endpoint" yaml:"endpoint"`
+	Name     string      `json:"name" yaml:"name"`
+	Provider string      `json:"provider" yaml:"provider"`
```
teemow: Ooops. Why do we have the provider in here? Is there a chance we have different authorities on different providers?

pkg/apis/core/v1alpha1/release_types.go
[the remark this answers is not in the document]
puja108: My issue is quite general with this. We have a CR which is generic between providers, and we are using names that could overlap between providers in their identifying information (i.e. their name and version) and we are relying solely on context information to differentiate the objects. That means in our config management we keep CRs that are named exactly the same, but only based on folders that they are in or clusters they exist in can be distinctly identified. This for me just opens up a room for errors and mistakes. Concrete use cases might rather be in future features we might want to have.

See also @teemow's comment below https://github.com/giantswarm/apiextensions/pull/140#discussion_r219491929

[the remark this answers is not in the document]
puja108: Especially if we think about at some point opening up our control plane towards the customer these kind of things might lead to confusion.

pkg/apis/core/v1alpha1/release_types.go
teemow: Yes I also think that the naming is rather unfortunate. 

Imo this shouldn't be part of the spec. Why aren't we using the kubernetes resource versioning?

https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/

xh3b4sd: Well, this is just the implementation we have right now and we cannot just remove it. This is not my opinion but a matter of fact. As I described above after we transitioned all operators to only manage one single version and having `release-operator` working properly we can get rid of the version bundle version and use the release information to configure the operator so it knows what to do. The reason so far why we did not use the k8s API version was simplicity and overhead reduction. With each new API version you need a new CRD and all this implies more leg work when crafting new operator versions. This needs more thought and has to be implemented if we want to have it in the future but this does not happen now.

pkg/apis/core/v1alpha1/release_types.go
```diff
@@ -48,9 +52,49 @@ func NewReleaseCRD() *apiextensionsv1beta1.CustomResourceDefinition {
...
+//         version: 0.1.0
+//       - endpoint: http://chart-operator:8000
+//         name: chart-operator
+//         version: 0.3.0
+//       - endpoint: http://cluster-operator:8000
+//         name: cluster-operator
+//         provider: azure
+//         version: 0.7.0
+//       date: "0001-01-01T00:00:00Z"
+//       version: "2.0.0"
```
kopiczko: WDYT about dropping `version` when we have that in the object name?

pkg/apis/core/v1alpha1/release_types.go
teemow: :+1: i now this is the implementation we use and a change would be out of scope. just wanted to understand it better. i think i got it now. thanks!

for the future we should probably write a doc how we do the versioning in resources and how to version apiextensions as best practice. would be something for sig-operator to think about imo.

xh3b4sd: Good shout. Noted: https://github.com/giantswarm/giantswarm/issues/4245.

pkg/apis/core/v1alpha1/release_types.go
xh3b4sd: This is still heavily controversial but I try to consider it once we settle with something reasonable. Thanks for your feedback.

pkg/apis/core/v1alpha1/release_types.go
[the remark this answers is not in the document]
MarcelMue: To get more behind the motivation of this:
Would it be enough to have a label or something similar that shows which provider it is? I don't quite follow the argument that we only separate these by folders, folders are a separating structure and changing the name to something like `2.0.0-AWS` doesn't change that angle too much IMO. 

With a label the user could easily recognize which provider it is related to and the provider would match exactly to the provider of the installation, unless something is wrong.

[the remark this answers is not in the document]
MarcelMue: The encoding of what is at its core metadata about the provider of the version, inside the identifier seems off.

xh3b4sd:
I added the provider as discussed in our meet. The version is kept explicitly because the object meta data name should never be used to derive any specific information from the desired state. The name stays to be the version for now but this is likely to change. 

PO please approve if everything is fine now.

teemow (approved):
Pretty neat! :shipit:

puja108 (approved):

PR #141, opened 2018-09-21
Fix postMappings typo

r7vme:
I've just noticed that portmapping look weird in resulted kvmconfig :)

```
      'portMappings yaml:':
      - name: ingress-http
        nodePort: 30032
        targetPort: 30010
```


xh3b4sd (approved):
So is this now broken everywhere? What is the impact? How can we fix it? What about the existing CRs?

kopiczko (approved):

MarcelMue (approved):

r7vme:
I'm going to vendor everyqwhere and manually fix kvmconfig CRs (seems only few clusters are new with this CR)

PR #143, opened 2018-09-21
adds key funcs to release authority type

xh3b4sd:
Follow up of https://github.com/giantswarm/release-operator/pull/33#discussion_r219552964.


MarcelMue (approved):
Let's go with this here, we should talk about this in sig-operator IMO

PR #144, opened 2018-10-02
adds proper cluster label to e2e charts

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/4214.

### DIFF
--- a/helm/apiextensions-aws-cluster-config-e2e-chart/templates/aws-cluster-config.yaml
+++ b/helm/apiextensions-aws-cluster-config-e2e-chart/templates/aws-cluster-config.yaml
@@ -2,6 +2,8 @@ apiVersion: "core.giantswarm.io/v1alpha1"
 kind: AWSClusterConfig
 metadata:
   name: {{ .Values.guest.name }}
+  labels:
+    giantswarm.io/cluster: "{{ .Values.guest.id }}"
 spec:
   guest:
     dnsZone: "{{ .Values.guest.dnsZone }}"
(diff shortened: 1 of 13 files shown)


helm/apiextensions-aws-cluster-config-e2e-chart/templates/aws-cluster-config.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: "core.giantswarm.io/v1alpha1"
 kind: AWSClusterConfig
 metadata:
   name: {{ .Values.guest.name }}
+  labels:
+    giantswarm.io/cluster: "{{ .Values.guest.id }}"
```
xh3b4sd: Well, what is the difference between `.Values.guest.name` and `.Values.guest.id`?

helm/apiextensions-aws-config-e2e-chart/templates/cluster.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: "provider.giantswarm.io/v1alpha1"
 kind: AWSConfig
 metadata:
   name: {{.Values.clusterName}}
+  labels:
+    giantswarm.io/cluster: "{{ .Values.clusterName }}"
```
xh3b4sd: Later we have to rename this to make the cluster ID more clear. But that has to happen separately and in multiple iterations.

helm/apiextensions-aws-cluster-config-e2e-chart/templates/aws-cluster-config.yaml
MarcelMue: IMO the difference should be: The name is just the name we create, it can change in the future to something different. The id should always be the id we generate. 
I don't know if that is the difference in reality as well :man_shrugging:

MarcelMue (approved):
I think this is fine, but you are right - the naming does not seem straight yet.

PR #145, opened 2018-10-03
add-networksetup-image

calvix:
fix missing network image

### DIFF
--- a/helm/apiextensions-kvm-config-e2e-chart/templates/cluster.yaml
+++ b/helm/apiextensions-kvm-config-e2e-chart/templates/cluster.yaml
@@ -50,6 +50,9 @@ spec:
         domain: worker.{{.Values.cluster.id}}.{{.Values.baseDomain}}
         labels: kvm-operator.giantswarm.io/version={{.Values.versionBundle.version}},giantswarm.io/provider=kvm
         port: 10250
+      networkSetup:
+        docker:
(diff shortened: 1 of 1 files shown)


xh3b4sd:
Why do we need this actually? I do not see where it is used in the other operators. In case we really need it for anything, we should hard code it in the key package where it is used and then we should have an issue for this so we do not forget to do it. Can somebody check and verify=

calvix:
its used here https://github.com/giantswarm/k8scloudconfig/blob/master/v_3_6_1/master_template.go#L2067

 since its k8scloudconfig that means all providers need to set it up somewhere

xh3b4sd:
@giantswarm/team-batman works on managed services. When you take more and more services out, do you see any chance to abstract the network setup here? With this we would get this refactored and the image pinned somewhere.

puja108:
2 things:

1. managed services as an area does currently (and as planned) not include core components that are needed for general functionality (e.g. Calico and I guess this networking setup is not part of it). That said, chart-operator could be used for such things, but would need some work to have reliable "phases" or "order of execution" (latter might somehow be already in). As we said core functionality stays in KaaS this would be in team spirit.

2. What does this unit do actually? If it's provider specific stuff, I think in the past we have kept that in provider operators (like with storage classes, or am I mistaken there?). If it's really used on all providers than I don't see anything against the place it is right now.

xh3b4sd:
AFAIK it is for all providers. @giantswarm/sig-infrastructure can you confirm? I think in this case we should pin the image in the `k8scloudconfig` repo.

PR #147, opened 2018-10-09
removes Vault token management

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/4152.


kopiczko (approved):
When it's not used, that's beautiful <3

PR #148, opened 2018-10-11
adds status condition times to align more with upstream

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/3988.


pkg/apis/core/v1alpha1/drainer_types.go
```diff
@@ -111,20 +109,17 @@ type DrainerConfigStatus struct {
 
 // DrainerConfigStatusCondition expresses a condition in which a node may is.
 type DrainerConfigStatusCondition struct {
-	// Date is the time of the given condition being updated.
-	Date time.Time `json:"date" yaml:"date"`
```
xh3b4sd: That is not used at all. Removing it should be safe.

tuommaki (approved):
Makes sense

kopiczko:
This is nice.

PR #151, opened 2018-10-14
remove deprecated aws types

teemow:
all old clusters have been upgraded. the code that handled the legacy clusters has been removed already. so we can clean up the resources as well.

i'll do a corresponding PR in the aws-operator too.


kopiczko:
This is cool. Please also run `scripts/gen.sh`.

teemow:
@kopiczko do you know already how the pruning works in 1.11? how do we apply schema changes is this done in kubernetesd or aws-operator?

tuommaki:
> how do we apply schema changes is this done in kubernetesd or aws-operator?

AFAIK there has been two kind of cases here:
* When new field has been introduced then the field has been populated by CR creator and / or the operation that uses this field has been defaulting to something over the transition period.
* When field is removed, one must first make sure that nothing uses it anymore. When component that creates the CR object has updated `apiextensions` dependency and it updates existing CR object by reading it from k8s and then updating it with `Update` (on contrary to `Patch`) the removed field will disappear (as k8s object is replacted on update).

Did this answer to your question?

tuommaki (approved):
When removed fields aren't used anymore this LGTM.

kopiczko:
As @tuommaki says. `prune` has nothing to do with vendoring those changes in particular (also it's independent from Go version).

First you should remove all unused packages from `aws-operator` https://github.com/giantswarm/aws-operator/tree/master/service/controller.

Then vendor this change by running `dep ensrue -update github.com/giantswarm/apiextensions`.

For `kubernetesd` you just vendor. There are no `v*` packages there.

We don't apply schema changes ATM. Old CRs stay as they are :(
This is to be implemented in `cluster-operator`.

kopiczko (approved):

teemow:
My question was not about vendoring. My question was when and how the schema of the k8s resources will be changed. Not sure if your answers are helping me. 

Does that mean that pruning is done once we `Update` but we don't do that at the moment? And if I understand this correctly we would need to iterate through all the resources? I thought we would update the schema and then kubernetes will do the pruning for us. But I might be wrong.

> This is to be implemented in cluster-operator

`cluster-operator` would be the wrong place imo. it should only be allowed to read its resources and not to change them. obviously if cluster-operator creates more resources then updating their schema would be the job of cluster-operator.

kopiczko:
Sorry I relate `pruning` too string with `dep` :D 

> Does that mean that pruning is done once we `Update` but we don't do that at the moment?

Correct we don't do that.

>> This is to be implemented in cluster-operator
>
> cluster-operator would be the wrong place imo.

I meant `AWSConfig` (which is managed by `cluster-opreator`) and not `AWSClusterConfig`. I agree `AWSClusterConfig` should be changed during the `Update` in `cluster-service`. It isn't implemented. I'm not sure in scope of what we should implement it. Releasing operators?

I agree this is all messy ATM. I also tried to clean those objects a bit before but I gave up because with `release-operator` it will be so much easier I decided I don't want to invest more energy into that.

tuommaki:
`Update` of resources happens when cluster is scaled or updated. In order to "prune" old fields from existing CRs, all of them must be processed.

From k8s 1.11 release notes (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#sig-api-machinery-1):
> Metadata of CustomResources is now pruned and schema-checked during deserialization of requests and when read from etcd. In the former case, invalid meta data is rejected, in the later it is dropped from the CustomResource objects. 

So only metadata is automatically pruned for now.

As Pawel already noted, your changes are about `AWSConfig` which is created by `cluster-operator` and reconciled by `aws-operator` so in this case it would be `cluster-operator's` responsibility to prune old fields, but as said, it's not implemented yet and for now only `Update` of CR object works (and this is performed by `kubernetesd`).

teemow:
This also fixes https://github.com/giantswarm/giantswarm/issues/4507

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -88,9 +82,8 @@ type AWSConfigSpecAWS struct {
 	// done in order to provide more HA during single availability zone failures.
 	// In case a specific availability zone fails, not all tenant clusters will be
 	// affected due to the described selection process.
-	AvailabilityZones int                  `json:"availabilityZones" yaml:"availabilityZones"`
-	CredentialSecret  CredentialSecret     `json:"credentialSecret" yaml:"credentialSecret"`
-	Etcd              AWSConfigSpecAWSEtcd `json:"etcd" yaml:"etcd"`
```
xh3b4sd: Is it safe to remove this?

xh3b4sd (approved):
When this works for you go ahead. I am just not sure about the etcd config.

pkg/apis/provider/v1alpha1/aws_types.go
teemow: Yes the PR was about this initially. This lives in `HostedZones` now

teemow:
I am closing this. I forgot that there is no way to drop fields in our CRs at the moment :sob:

xh3b4sd:
Why? When the code is not used anymore we can remove it.

teemow:
Let's better keep the code if the schema is still there and the fields will stay in the CR.

PR #152, opened 2018-10-15
sets default namespace for CRs

xh3b4sd:


MarcelMue (approved):
LGTM

tuommaki (approved):
FMI: This is for the Helm weirdness?

xh3b4sd:
Somehow yes. We install via our `e2e-harness/pkg/release` package and this has the `giantswarm` namespace configured. So right now the CR is in this namespace but the e2e tests expect it in `default`.

kopiczko:
Why `KVMConfig` isn't changed as well?

xh3b4sd:
On KVM we have no encapsulation on the CI builds and they are namespaced very differently. 😕

xh3b4sd:
The direction should rather be to get rid of the namespacing weirdness I feel like. for now we only need this because of the isolation issues on KVM which we want to address in the future with more dynamic control plane management.

MarcelMue:
I think I was wrong here, the `kvmconfig` is also in the default namespace - we use the labelselector to choose the right one instead of the namespace. I will provide a PR to add the `default` namespace to kvm.

PR #157, opened 2018-10-23
adds list of AZs to AWSConfig CR

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2202.


pkg/apis/core/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -303,6 +303,7 @@ func (in *AzureClusterConfigSpec) DeepCopy() *AzureClusterConfigSpec {
 func (in *AzureClusterConfigSpecGuest) DeepCopyInto(out *AzureClusterConfigSpecGuest) {
 	*out = *in
 	in.ClusterGuestConfig.DeepCopyInto(&out.ClusterGuestConfig)
+	out.CredentialSecret = in.CredentialSecret
 	if in.Masters != nil {
```
pipo02mix: why it has been changed if the credential secret was already defined?

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -65,10 +65,26 @@ type AWSConfigSpec struct {
...
+	// useful usage of availability zones.
+	//
+	//     1 element means no AZ HA. All workers run within the same AZ.
+	//
+	//     2 elements mean normal AZ HA. All workers run within 2 different AZs.
+	//
+	//     4 elements mean high AZ HA. All workers run within 4 different AZs.
+	//
+	AvailabilityZones []string             `json:"availabilityZones" yaml:"availabilityZones"`
+	CredentialSecret  CredentialSecret     `json:"credentialSecret" yaml:"credentialSecret"`
```
pipo02mix: Just wondering if we don't need a struct instead of a single string array. For example, if we will need different min/max nodes or lifecycle hooks by AZ.

pipo02mix (commented):
A couple of questions

pkg/apis/core/v1alpha1/zz_generated.deepcopy.go
xh3b4sd: I can imagine the code generator was not invoked when adding the last code. Uncertain.

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -65,10 +65,26 @@ type AWSConfigSpec struct {
...
-	CredentialSecret CredentialSecret     `json:"credentialSecret" yaml:"credentialSecret"`
-	Etcd             AWSConfigSpecAWSEtcd `json:"etcd" yaml:"etcd"`
+	API AWSConfigSpecAWSAPI `json:"api" yaml:"api"`
+	// TODO remove the deprecated AZ field due to AvailabilityZones.
+	//
+	//     https://github.com/giantswarm/giantswarm/issues/4507
+	//
+	AZ string `json:"az" yaml:"az"`
+	// AvailabilityZones is a list of AWS availability zone references defining
+	// where to run the tenant cluster's worker nodes. Must hold 1, 2 or 4
```
xh3b4sd: @giantswarm/sig-ux this requires your attention here. We have to sort the validation here a bit. The thing is that some AZs apparently do not even have 4 options to chose from. So this is already wrong and seems to vary from case to case. How do we want to handle that? At which level do we want to validate this?

xh3b4sd:
Some more questions that came up during SIG UX sync. 

- Is it possible to add and remove AZs in existing clusters?
- When having 3 AZs and 30 worker nodes, how many worker nodes run in each AZ?
- Will AZs be the very same or any different? Will there always be the very same amount of nodes in each AZ?

pkg/apis/provider/v1alpha1/aws_types.go
teemow: @giantswarm/sig-ux this is an implementation detail leaking into the interface somehow.

If we split up ip ranges for clusters we will always end up having binary slices. So if the user chooses 3 AZs we will have one unused ip range slice that is "wasted". I am not 100% sold that "1, 2, 4" here is a good solution. There are definitely regions in which there are only 3 availability zones. So it will be weird for the user to choose 4 and then anyway having the 4th slice wasted.

@xh3b4sd regarding the value of each element. That will be eg `eu-west-1a` etc? you should probably add this as an example in the comment.

teemow: I wrote down some considerations regarding IPAM in the spec: https://github.com/giantswarm/giantswarm/pull/2202/files#diff-195efbf97630852fcc8ee4886d9a1760R27

xh3b4sd: I will rephrase it and not mention specific numbers as this depends on the specific case. The validation we have to implement in the `api` has to be aware of the provided AZs in order to verify user input properly.

pkg/apis/provider/v1alpha1/aws_types.go
xh3b4sd: We talked and decided to keep it as it is as there is not a real use case on the horizon at the level of `aws-operator`. We might want to structure it differently later at the `cluster-operator` level or something. Thank you for the feedback.

pkg/apis/provider/v1alpha1/aws_types.go
xh3b4sd: I added an example with some AZ names.

xh3b4sd:
Anything left? Fishing for LGTM.

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -65,10 +65,26 @@ type AWSConfigSpec struct {
...
+	//     https://github.com/giantswarm/giantswarm/issues/4507
+	//
+	AZ string `json:"az" yaml:"az"`
+	// AvailabilityZones is a list of AWS availability zone references defining
+	// where to run the tenant cluster's worker nodes. There are limitations on
+	// availability zones settings due to binary IP range splitting. When for
+	// instance choosing 3 availability zones, the configured IP range will be
+	// split into 4 ranges and thus one of it will not be able to be utilized.
+	// Such limitations have to be considered when designing the network topology
+	// and configuring tenant cluster HA via Azs. The elements of the list might
```
kopiczko: Nit: `Azs` -> `AZs`.

xh3b4sd: Mr. Eagle Eye.

kopiczko (approved):
LGTM, we can change godoc any time.

xh3b4sd:
I would also like to have one from my PO and one from SIG UX.

teemow (approved):

pipo02mix (approved):
LGTM

pkg/apis/core/v1alpha1/zz_generated.deepcopy.go
marians: Yes, my bad in https://github.com/giantswarm/apiextensions/pull/155

PR #158, opened 2018-10-24
Ensure code generation through CI check

marians:
This addition to the CircleCI config should ensure that code generation has been executed.

It executes the generator script and then checks if the code has changed through that.


JosephSalisbury:
@marians please create an issue in sig-releng

marians:
Issue is in https://github.com/giantswarm/giantswarm/issues/4578. Closing.

PR #160, opened 2018-10-26
makes AZs a number

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2202. Team Spirit reconsidered the approach of configuring and managing multiple availability zones. Consequence is to simply use a number defined in the `AWSConfig` CR as the `api-spec` already suggests as user input. Background is that we think now the `aws-operator` should do the calculation and randomization of AZs internally. It only gets the available list configured at startup via the `installations` repo and the user input desired for a specific tenant cluster. Validation is also happening on the `api` level.

### DIFF
--- a/pkg/apis/provider/v1alpha1/aws_types.go
+++ b/pkg/apis/provider/v1alpha1/aws_types.go
@@ -79,6 +79,15 @@ type AWSConfigSpecAWS struct {
 	// it will not be able to be utilized. Such limitations have to be considered
 	// when designing the network topology and configuring tenant cluster HA via
 	// AvailabilityZones.
+	//
+	// The selection and usage of the actual availability zones for the created
+	// tenant cluster is randomized. In case there are 4 availability zones
+	// provided in the used region and the user selects 2 availability zones, the
+	// actually used availability zones in which tenant cluster workload is put
+	// into will tend to be different across tenant cluster creations. This is
+	// done in order to provide more HA during single availability zone failures.
+	// In case a specific availability zone fails, not all tenant clusters will be
+	// affected due to the described selection process.
 	AvailabilityZones int                  `json:"availabilityZones" yaml:"availabilityZones"`
 	CredentialSecret  CredentialSecret     `json:"credentialSecret" yaml:"credentialSecret"`
 	Etcd              AWSConfigSpecAWSEtcd `json:"etcd" yaml:"etcd"`
--- a/pkg/apis/provider/v1alpha1/aws_types.go
+++ b/pkg/apis/provider/v1alpha1/aws_types.go
@@ -71,18 +71,15 @@ type AWSConfigSpecAWS struct {
 	//     https://github.com/giantswarm/giantswarm/issues/4507
 	//
 	AZ string `json:"az" yaml:"az"`
-	// AvailabilityZones is a list of AWS availability zone references defining
-	// where to run the tenant cluster's worker nodes. There are limitations on
-	// availability zones settings due to binary IP range splitting. When for
-	// instance choosing 3 availability zones, the configured IP range will be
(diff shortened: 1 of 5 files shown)


pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -71,18 +71,15 @@ type AWSConfigSpecAWS struct {
...
-	AvailabilityZones []string             `json:"availabilityZones" yaml:"availabilityZones"`
+	// AvailabilityZones is the number of AWS availability zones used to spread
+	// the tenant cluster's worker nodes across. There are limitations on
+	// availability zone settings due to binary IP range splitting and provider
+	// specific region capabilities. When for instance choosing 3 availability
+	// zones, the configured IP range will be split into 4 ranges and thus one of
+	// it will not be able to be utilized. Such limitations have to be considered
+	// when designing the network topology and configuring tenant cluster HA via
+	// AvailabilityZones.
+	AvailabilityZones int                  `json:"availabilityZones" yaml:"availabilityZones"`
```
xh3b4sd: There is some migration code already in the WIP resources and `kubernetesd` but this should be easily manageable. I find a way.

MarcelMue (approved):
LGTM :+1:

pkg/apis/provider/v1alpha1/aws_types.go
kopiczko: Are AZ selected at random? Should that be described in godoc as well?

xh3b4sd: Good shout. I added something to the godoc.

kopiczko: It's cool now.

pkg/apis/provider/v1alpha1/aws_types.go
```diff
@@ -71,18 +71,24 @@ type AWSConfigSpecAWS struct {
...
-	AvailabilityZones []string             `json:"availabilityZones" yaml:"availabilityZones"`
+	// The selection and usage of the actual availability zones for the created
+	// tenant cluster is randomized. In case there are 4 availability zones
+	// provided in the used region and the user selects 2 availability zones, the
+	// actually used availability zones in which tenant cluster workload is put
+	// into will tend to be different across tenant cluster creations. This is
+	// done in order to provide more HA during single availability zone failures.
+	// In case a specific availability zone fails, not all tenant clusters will be
+	// affected due to the described selection process.
+	AvailabilityZones int                  `json:"availabilityZones" yaml:"availabilityZones"`
```
kopiczko: I forgot again. We need to change chart as well. And e2etemplates.

xh3b4sd: Addressed.

tuommaki (dismissed):
I need some clarification on why we decided to use number in `AWSConfig` instead of list of specified AZs. I wrote something in our team Slack so lets discuss a bit before approval.

teemow:
We decided to go with a number because we wanted to move the decision which AZ is picked into the provider specific operator itself. From your comments I see now that this might be a problem if we would like to have AZs chosen randomly but a deterministic way to reconcile it too. It might be easier to have the AZs in the spec vs in the status. How would we make the AZ selection deterministic?

xh3b4sd:
It may just be a pattern that you define something specific indirectly using some settings which cause the operator to compute arbitrary or randomized content once and puts it into the status, which is then used as desired state. Makes sense somehow to me. We can further break this down and think about who should provide the specific list to the operator in the first place. To keep moving here now I think this is still reasonable. And then we iterate. Anything against this? I would like to have more approvals since this is new and complicated.

kopiczko (approved):
LGTM

PR #161, opened 2018-10-26
adds aws specific tenant cluster status structures for AZ management

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2202.


kopiczko (approved):
LGTM

tuommaki (approved):
Something is bugging me when defining AZ subnets like this. I don't have better proposal, but some background from IPAM: Currently tenant cluster network allocation is stored in `StatusCluster.Network.CIDR`: https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/provider/v1alpha1/status_types.go#L34 This works for all cases as it's not provider dependent in any way.

Now related information is separated in different hierarchy. I don't have better proposal for this and I'm not blocking on this, but just bringing this up if someone has better ideas. My initial idea here was to have provider specific subnet computations implemented with deterministic function so given the same input subnet the output is always the same. In that case there wouldn't be any need to persist it anywhere, but I understand the desire to have visibility for human observer in here. I'm just afraid if there is possibility to get these two related fields out of sync and still keep the code running with invalid state.

teemow:
I think this is a good point. The code to calculate the AZs and subnets needs to be deterministic. Otherwise we wouldn't be able to reconsile it. The structure in this PR is only the actual state. What would happen if suddendly the calculated subnets do not match the actual state?

But the data structure makes sense for AWS imo.

kopiczko:
My understanding is the operator takes `StatusCluster.Network.CIDR`, splits it randomly and writes AZ subnets into status. If AZ subnets are present in the status, the operator just reads them.

If there is a need to recreate exact the same cluster we can create CR with AZ subnets set in status - tricky but I think there are reasons to do that should very rare if any. 

Am I missing something?

PR #162, opened 2018-11-08
removes unused hyperkube docker image

xh3b4sd:
I checked the provider operators and couldn't find code using this. So I removed it. Feel free to double check so I did not miss something.


helm/apiextensions-aws-config-e2e-chart/templates/cluster.yaml
```diff
@@ -42,9 +42,6 @@ spec:
       dns:
         ip: "172.31.0.10"
       domain: "cluster.local"
-      hyperkube:
-        docker:
-          image: "quay.io/giantswarm/hyperkube:v1.7.5_coreos.0"
```
kopiczko: I anything in the tests would use this value we'd probably see it :)

kopiczko (approved):
feelsgood

tuommaki (approved):
Yeah, when even some e2e tests (that spin tenant clusters) succeed, this should be all fine.

PR #164, opened 2018-11-15
Add AZs to cluster_guest_types

MarcelMue:
As I understand it, this is the right way to put this information for `cluster-service` to write it into the CR. Cluster-service then checks the different cluster CRs (for aws, azure and kvm) to recover that information when the cluster searcher is used.

towards https://github.com/giantswarm/giantswarm/pull/2202

### DIFF
--- a/pkg/apis/core/v1alpha1/cluster_guest_types.go
+++ b/pkg/apis/core/v1alpha1/cluster_guest_types.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 type ClusterGuestConfig struct {
+	AvailabilityZones int `json:"availability_zones,omitempty" yaml:"availability_zones,omitempty"`
 	// DNSZone for guest cluster is supplemented with host prefixes for
 	// specific services such as Kubernetes API or Etcd. In general this DNS
 	// Zone should start with `k8s` like for example


xh3b4sd:
I am so not sure about `cluster-operator` right now and where this is going or what we should do there now.

MarcelMue:
I am not sure either, the way I see the current state:
- We use this CR in cluster-service to get information about guestclusters: https://github.com/giantswarm/cluster-service/blob/790e0ab5a6664e55b8cea4d004a2d712eee6b452/service/cluster/searcher/aws.go#L43-L53
- We fall back to the storage service if the CR doesn't exist (for older clusters) : https://github.com/giantswarm/cluster-service/blob/790e0ab5a6664e55b8cea4d004a2d712eee6b452/service/cluster/searcher/service.go#L87-L96
- `cluster-operator` basically does nothing with these CRs AFAIK
- We can't move this logic directly into the api because of the storage fallback for older clusters

So my idea is:
- Add the AZ value to this CR and set / retrieve it in `cluster-service` for now
- We can move the code easily to API once all old clusters are gone
- Aligning these CRs with the `cluster-api` concept in the future

xh3b4sd:
I am not sold anymore the indirect with yet another CR, which are these cluster config CRs, are really good. We moved IPAM to the provider operators, or are basically in the mids of doing that. Vault magics got resolved in other operators. The ingress ports and rangepool magics are soon gone just like `ingress-operator`. The VNI part in KVM is left but that can maybe just go into the `kvm-operator` or something like that.

kopiczko (approved):
LGTM. `cluster-operator` doesn't reconcile `awsconfigs` yet. Just remember to set this information in both places in `kubernetesd`.

Maybe we change the approach here but for now we have what we have.

MarcelMue:
Also: I ran `./gen.sh` but nothing changed here - is this correct?

xh3b4sd:
When there is only a string field added, yes.

tuommaki (approved):
I'm ok with the change itself. Unsure about discussion on `cluster-operator` right now. That would be something to discuss later, in different context I'd say.

PR #165, opened 2018-11-15
fixes JSON and YAML tags

xh3b4sd:
AFAIK the resource status is only used in the `azure-operator` right now. When this change here is applied the resource status is reset and the reconciliation starts from the beginning again, which it does anyway, because it operates in circles and goes through three different stages. This should not do harm.

### DIFF
--- a/pkg/apis/provider/v1alpha1/status_types.go
+++ b/pkg/apis/provider/v1alpha1/status_types.go
@@ -86,7 +86,7 @@ type StatusClusterNode struct {
 // its own implementation and means in order to fulfil its premise.
 type StatusClusterResource struct {
 	Conditions []StatusClusterResourceCondition `json:"conditions" yaml:"conditions"`
-	Name       string                           `json:"status" yaml:"status"`
+	Name       string                           `json:"name" yaml:"name"`
 }
 
 // StatusClusterResourceCondition expresses the conditions in which an


xh3b4sd:
I would like to get a couple of eyes on this and multiple approvals.

kopiczko (commented):
I have no idea. But if you need a second pair of eyes we can sit down together.

tuommaki:
> This should not do harm.

What about that recent issue with cluster-mss (or something alike) deployment? Would that deployment recreation be idempotent when status condition is "missed" due to change in name?

Also, when this gets changed, does it mean that old CRs have orphaned status fields? I don't think it hurts that much, but just wondering.

xh3b4sd:
The deployment creation/update is idempotent anyway. Nothing changes. The operator only goes back to stage one and continues from there. 

I merged a fix for the duplication yesterday already. The fix is in the wip version. At some point all duplicates will be gone. What happens with this change here is that the resource statuses we have with the wrong names get orphaned and should be deleted manually. I could try to figure a more graceful way by adding a new tag and keeping the old and support some migration code or something.

tuommaki (approved):

tuommaki:
> I could try to figure a more graceful way by adding a new tag and keeping the old and support some migration code or something.

I think this is fine when everything works correctly. Amount of potentially orphaned data is so small that it won't affect anything and will eventually go away anyway.

PR #168, opened 2018-11-22
adds status subresource for NodeConfig and DrainerConfig

xh3b4sd:
As it seems we do not have this enabled yet. I want to get it in for some e2e tests in `operatorkit`. It would also be beneficial to get this straight in the `node-operator` and provider operators where they work against the `DrainerConfig` status.


tuommaki (approved):

xh3b4sd:
Works on `godsmack`.

PR #181, opened 2018-12-12
Remove the ASG Name again

MarcelMue:
This was brought up here:
https://github.com/giantswarm/giantswarm/issues/4543#issuecomment-446606430

With autodiscovery we can simply use the cluster name which we already have in `cluster-operator`. The ASG Name is not needed therefor.

We can simply remove this field because it was just added and nothing uses it.


xh3b4sd (approved):
That makes sense to me.

PR #189, opened 2019-01-14
 Update to kubernetes 1.13.2

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4709


pkg/clientset/versioned/typed/application/v1alpha1/app.go
```diff
@@ -76,23 +78,33 @@ func (c *apps) Get(name string, options v1.GetOptions) (result *v1alpha1.App, er
 
 // List takes label and field selectors, and returns the list of Apps that match those selectors.
 func (c *apps) List(opts v1.ListOptions) (result *v1alpha1.AppList, err error) {
+	var timeout time.Duration
+	if opts.TimeoutSeconds != nil {
+		timeout = time.Duration(*opts.TimeoutSeconds) * time.Second
+	}
```
tuommaki: I was wondering if it would make sense to extract this pattern in to a function since it's repeating a lot?

Something like this:
```go
func timeoutFromListOpts(opts v1.ListOptions) time.Duration {
    var timeout time.Duration
    if opts.TimeoutSeconds != nil {
        timeout = time.Duration(*opts.TimeoutSeconds) * time.Second
    }
    return timeout
}
```

...and then below it could be just `Timeout(timeoutFromListOpts(opts))`

tuommaki (approved):
In general LGTM. Small note on repeating pattern.

pkg/clientset/versioned/typed/application/v1alpha1/app.go
kopiczko: This is generated :D We only change `pkg/apis` here.

tuommaki: :joy: I somehow completely missed the filename etc. Yeah... Mornings... All good then :+1:

MarcelMue (approved):

rossf7:
⚠️ The merge fest begins.

PR #190, opened 2019-01-16
Add extra status fields for apps and charts CRDs

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4949

See https://github.com/giantswarm/chart-operator/pull/183#issue-245609254 for more details.


pkg/apis/application/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -523,3 +525,13 @@ func (in *ChartStatus) DeepCopy() *ChartStatus {
 	in.DeepCopyInto(out)
 	return out
 }
+
+// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new DeepCopyTime.
```
rossf7: The deep copy magic for time.Time fields.

xh3b4sd (approved):

MarcelMue (approved):

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -158,8 +158,17 @@ type AppSpecUserConfigSecret struct {
...
+	// e.g. 0.21.0.
+	AppVersion string `json:"appVersion" yaml:"appVersion"`
+	// LastDeployed is the time when the app was last deployed.
+	LastDeployed DeepCopyTime `json:"lastDeployed" yaml:"lastDeployed"`
 	// Status is the status of the deployed app.
+	// e.g. DEPLOYED.
 	Status string `json:"status" yaml:"status"`
+	// Version is the version of the deployed app,
+	// e.g. 1.0.0.
+	Version string `json:"version" yaml:"version"`
```
kopiczko: What is the difference between `Version` and `AppVersion`?

rossf7: Its the version of the app being deployed rather than the version of the chart. The naming is a bit confusing but its useful to have both versions. Godoc is updated so it should be clearer now.

https://docs.helm.sh/developing_charts/#the-chart-yaml-file

TheoBrigitte (approved):

pkg/apis/application/v1alpha1/app_types.go
kopiczko: Ack. So this is just an arbitrary value pulled from Chart.yaml?

rossf7: Yes its an optional value set in the Chart.yaml.

PR #191, opened 2019-01-22
App Catalog CRD changes

rossf7:
Updates CRDs with changes from latest spec PR.

- AppCatalog CRD now also has config.
- Release field in App CRD is renamed to Version.

See https://github.com/giantswarm/giantswarm/pull/5099


pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -60,6 +60,9 @@ type AppCatalogSpec struct {
 	// CatalogStorage references a map containing values that should be
 	// applied to the appcatalog.
 	CatalogStorage AppCatalogSpecCatalogStorage `json:"catalogStorage" yaml:"catalogStorage"`
+	// Config is the config to be applied when apps belonging to this
+	// catalog are deployed.
+	Config AppCatalogSpecConfig `json:"config" yaml:"config"`
```
rossf7: AppCatalog CRD now also has a configmap and secret.

- We use this for VOO as a replacement for the draughtsman configmap and secret.
- For App Catalog the configmap will configure the Image Registry so we use Aliyun for AWS China.

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -82,11 +82,11 @@ type AppSpec struct {
...
 	// e.g. monitoring
 	Namespace string `json:"namespace" yaml:"namespace"`
-	// Release is the version of the app that should be deployed.
-	// e.g. 1.0.0
-	Release string `json:"release" yaml:"release"`
 	// UserConfig is the user config to be applied when the app is deployed.
 	UserConfig AppSpecConfig `json:"userConfig" yaml:"userConfig"`
+	// Version is the version of the app that should be deployed.
+	// e.g. 1.0.0
+	Version string `json:"version" yaml:"version"`
```
rossf7: Release is renamed to Version. This makes the naming clearer and avoids confusion with the Release CRD.

tomahawk28: I need to apply this change in app-operator since it's using `Release` to make tarball url.

tomahawk28:
After these changes, should we copy all the configMap & secret from appCatalog to app CR?

pkg/apis/application/v1alpha1/app_types.go
rossf7: No you were right the first time. We'll need to update the code that generates the tarball URL but its just renaming a field.

rossf7:
> After these changes, should we copy all the configMap & secret from appCatalog to app CR

We will merge the values from the appCatalog CR and the app CR. The configmap and secret resources in app-operator will do this.

kopiczko (approved):

tomahawk28 (approved):

xh3b4sd (commented):
@giantswarm/sig-ux should always be included in `apiextensions` for knowledge share and feedback as the k8s types are going to be user facing. We also want to have more godoc with an example of how the YAML structure looks like in the end. That way we can reason about the changes more easily. It would be nice to have the example for `AppCatalog`.

puja108:
true, we forgot to ass the example CRD in the godoc, we should add that

rossf7:
@xh3b4sd Good catch. Whitespace should be sorted now. PTAL.

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -43,6 +43,30 @@ func NewAppCatalogCRD() *apiextensionsv1beta1.CustomResourceDefinition {
...
+//        URL: “https://giantswarm.github.com/app-catalog/”
+//      config:
+//        configMap:
+//          name: "app-catalog-values"
+//          namespace: "giantswarm"
+//        secret:
+//          name: "app-catalog-secrets"
+//          namespace: "giantswarm"
+//      logoURL: “https://s.giantswarm.io/...”
+
```
xh3b4sd: The godoc has to be put directly above the type it documents. The example is not displayed right now. 

screenshot 2019-01-23 at 15 06 29

You can check it locally with `godoc -http=:6060 -goroot=$HOME/go` and go to http://localhost:6060/pkg/github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1/#AppCatalog.

rossf7: Thanks. Fixed.

xh3b4sd: Nice. 

screenshot 2019-01-23 at 15 54 07

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -46,6 +46,29 @@ func NewAppCatalogCRD() *apiextensionsv1beta1.CustomResourceDefinition {
...
+//    kind: AppCatalog
+//    metadata:
+//      name: “giantswarm”
+//      labels:
+//        app-operator.giantswarm.io/version: "1.0.0"
+//
+//    spec:
+//      title: “Giant Swarm”
+//      description: “Catalog of Apps by Giant Swarm”
+//      catalogStorage:
```
kopiczko: @rossf7 Can we rename that to `storage:`? Having catalog prefix stutters.

rossf7: Sure good point. I'll propose a PR.

PR #195, opened 2019-01-31
Add e2e chart for chart CRD

rossf7:
Adds the e2e chart for the new chart CRD. 

See https://github.com/giantswarm/e2etemplates/pull/52 for the related template.


helm/apiextensions-chart-e2e-chart/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+      name: "mychart-secrets"
+      namespace: "default"
+  tarballURL: "https://giantswarm.github.io/sample-catalog/kubernetes-test-app-chart-0.6.8.tgz"
+
+chartOperator:
+  version: "1.0.0"
+
+configMap:
+  values:
+    Installation:
```
kopiczko: I wonder why Installations isn't a top level key. Is it copied from other charts?

kopiczko (approved):

helm/apiextensions-chart-e2e-chart/values.yaml
MarcelMue: Seems like a mistake to me too, `configmap` should have the user data and installations should take our overwrites from installations.

puja108: no, this is right.

I think you are misunderstanding something. This configmap (and below secret), are generated outside by app-operator based on config from installations, this is different from the current user configmap with resides inside tenant clusters and overwrites this. as explained in the app operator spec there'll be 3 levels of config: installations (global config per installation), cluster, user. currently the first 2 are in CP and the last in tenant cluster, but in the future all 3 should be in CP and app-operator merges them into a single CM and Secret which is depicted here.

kopiczko: It was too early for me. I was thinking about collisions between the charts and all other sorts of things you don't want to know :D 

It makes sense.

rossf7: Yes this `values` key replaces the `values.json` we used for chartconfig CRDs.

We needed to switch to YAML to be compatible with the installations repo. It's also nicer for managed apps which is cool.

PR #197, opened 2019-02-01
drop date and use last transition time instead

xh3b4sd:
Fixes https://github.com/giantswarm/giantswarm/issues/3988.


xh3b4sd:
This has to stay for a little while longer. Not all clusters are prepared yet.

xh3b4sd:
Irrelevant

PR #199, opened 2019-02-01
release, scripts: add Release and ReleaseCycle CRDs

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5249


pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -0,0 +1,196 @@
...
+								"changelog",
+								"components",
+								"parentVersion",
+								"version",
+							},
+						},
+						"status": {
+							Type: "object",
+							Properties: map[string]apiextensionsv1beta1.JSONSchemaProps{
+								"cycle": NewReleaseCycleCRD().Spec.Validation.OpenAPIV3Schema.Properties["spec"],
```
kopiczko: I took a shortcut here.

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -0,0 +1,196 @@
...
+	// Name of the release component.
+	Name string `json:"name" yaml:"name"`
+	// Version of the release component.
+	Version string `json:"version" yaml:"version"`
+}
+
+type ReleaseStatus struct {
+	// Cycle is the most recent observed copy of the specification of the
+	// ReleaseCycle CR referencing this Release CR.
+	Cycle ReleaseCycleSpec `json:"cycle,omitempty" yaml:"cycle,omitempty"`
```
kopiczko: I also took a shortcut here.

pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -0,0 +1,131 @@
...
+								"phase",
+								"release",
+							},
+						},
+						"status": {
+							Type: "object",
+						},
+					},
+				},
+			},
```
kopiczko: I added `Validation` above. I haven't tested that yet. But I'll do before merging.

pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -0,0 +1,131 @@
...
+										{Raw: []byte("disabled")},
+										{Raw: []byte("eol")},
+									},
+								},
+								"release": {
+									Type: "object",
+									Properties: map[string]apiextensionsv1beta1.JSONSchemaProps{
+										"name": {
+											Type:      "string",
+											MinLength: toInt64P(3),
```
xh3b4sd: Why 3? `v0.0.0` would be 6.

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -0,0 +1,196 @@
...
+			Version: "v1alpha1",
+			Names: apiextensionsv1beta1.CustomResourceDefinitionNames{
+				Kind:     "Release",
+				Plural:   "releases",
+				Singular: "release",
+			},
+			Subresources: &apiextensionsv1beta1.CustomResourceSubresources{
+				Status: &apiextensionsv1beta1.CustomResourceSubresourceStatus{},
+			},
+			Validation: &apiextensionsv1beta1.CustomResourceValidation{
```
xh3b4sd: Would it make sense to put the validation into a separate method and file? Getting an idea about the scheme is really hard now.

pkg/apis/release/v1alpha1/to_funcs.go
```diff
@@ -0,0 +1,5 @@
+package v1alpha1
+
+func toInt64P(i int64) *int64 {
```
xh3b4sd: What about `giantswarm/cast` or something like that? We have similar code in `api`.

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -0,0 +1,196 @@
...
+	return metav1.TypeMeta{
+		APIVersion: version,
+		Kind:       kindRelease,
+	}
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type Release struct {
```
xh3b4sd: Please add godoc with an example output of the CR YAML. That helps SIG UX a lot to reason about future changes.

pkg/apis/release/v1alpha1/release_cycle_types.go
kopiczko: According to the spec it should be `0.0.0` which is `5`. The thing is we can set it to any arbitrary value while testing. This is described in the spec. I thought `3` is ok because this is a sane minimum length. If this was strictly semver I'd go with regexp like here: https://github.com/giantswarm/apiextensions/pull/199/files#diff-cde3a23593317e018cc324672345dc8bR104

kopiczko: If you have a reason to change that knowing what I wrote above let me know.

pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -0,0 +1,134 @@
...
+					"release",
+				},
+			},
+			"status": {
+				Type: "object",
+			},
+		},
+	},
+}
+
```
rossf7: The usual comment with what the CRD should look like it missing. Although I'm not sure how useful that is. The CR example is more important IMO.

pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -0,0 +1,134 @@
...
+	CyclePhaseUpcoming ReleaseCyclePhase = "upcoming"
+	CyclePhaseEnabled  ReleaseCyclePhase = "enabled"
+	CyclePhaseDisabled ReleaseCyclePhase = "disabled"
+	CyclePhaseEOL      ReleaseCyclePhase = "eol"
+)
+
+type ReleaseCyclePhase string
+
+var releaseCycleValidation = &apiextensionsv1beta1.CustomResourceValidation{
+	// See http://json-schema.org/learn.
```
rossf7: Having the schema validation is really cool. We should use this more.

@puja108 @oponder FYI. Could be useful for the front end changes for creating apps CRs?

pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -0,0 +1,134 @@
...
+func NewReleaseCycleTypeMeta() metav1.TypeMeta {
+	return metav1.TypeMeta{
+		APIVersion: version,
+		Kind:       kindReleaseCycle,
+	}
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
```
rossf7: I also think we need the CR example here. 

It makes it easier for sig UX to review. But also easier for everyone to scan when reviewing changes.

pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -0,0 +1,134 @@
...
+	// Release contains information about Release CR referenced by this ReleaseCycle CR.
+	Release ReleaseCycleSpecRelease `json:"release" yaml:"release"`
+}
+
+type ReleaseCycleSpecRelease struct {
+	// Name is the name of the Release CR referenced by this ReleaseCycle CR
+	Name string `json:"name" yaml:"name"`
+}
+
+type ReleaseCycleStatus struct {
```
rossf7: Do we know what will go in the status? Or is this just a placeholder?

pkg/apis/release/v1alpha1/release_types.go
rossf7: +1

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -0,0 +1,199 @@
...
+	// Version is the version of the release. Releases with semver version
+	// (without the "v" prefix) are taken from control-plane AppCatalog.
+	// All other releases are taken from control-plane-test AppCatalog.
+	Version string `json:"version" yaml:"version"`
+}
+
+type ReleaseSpecChangelogEntry struct {
+	// Component name.
+	Component string `json:"component" yaml:"component"`
+	// Description of the component changes expressed in full sentence.
```
rossf7:
```suggestion
	// Description of the component changes expressed as a full sentence.
```

rossf7 (commented):
Direction is cool but I'd like the CR examples to be included. Otherwise just minor stuff.

pkg/apis/release/v1alpha1/release_cycle_types.go
kopiczko: Placeholder. Nothing is specified so far.

pkg/apis/release/v1alpha1/release_cycle_types.go
kopiczko: I don't want to deal with that. Especially with validation. I thought about using some tool like https://github.com/gobuffalo/packr/tree/master/v2 and then having yamls which would be converted in `init` functions. But I don't want to introduce/propose such things within this PR.

rossf7: It does add overhead but I think its hard to review structure changes just with the fields GoDoc.

`packr` looks cool and I agree tooling would be good here. Also to be clear I don't care much about the CRD example YAML and I'd be fine with dropping that. It's the CR example YAML that I think is important.

rossf7 (approved):
LGTM. 

Schema validation is very nice. I'd still like to see the CR example YAML but I've said my piece.

pkg/apis/release/v1alpha1/release_cycle_types.go
puja108: yes, we should use this everywhere, I'd say all CRDs should have schema validation as minimum req, topic for @giantswarm/sig-operator ?

pkg/apis/release/v1alpha1/release_cycle_types.go
puja108: not sure if that's the same you are talking about but we said that we want to include a full example YAML in the docs and in the same PR where we add the CRDs cause it's much easier to discuss a CRD in that style than the types here

pkg/apis/release/v1alpha1/release_cycle_types.go
puja108: not relevant for this review, but maybe we wanna start thinking about some default status stuff we want to have, I've seen things like last recon, or last change or idk, another topic for @giantswarm/sig-operator ?

pkg/apis/release/v1alpha1/release_cycle_types.go
kopiczko: I check it works fine. I had to change enums values from `"value"` to ``` `"value"` ``` but that's a detail.

The most disappointing thing is you still can add unspecified fields. I tried to set `"additionalProperties": false` but I got `additionalProperties and properties are mutual exclusive` error.

More details here https://kubernetes.slack.com/archives/C0EG7JC6T/p1549374425866800

kopiczko:
PTALA

pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -0,0 +1,196 @@
...
+//	apiVersion: "release.giantswarm.io/v1alpha1"
+//	kind: "ReleaseCycle"
+//	metadata:
+//	  name: "aws.v6.1.0"
+//	  labels:
+//	    giantswarm.io/managed-by: "opsctl"
+//	    giantswarm.io/provider: "aws"
+//	spec:
+//	  disabledDate: 2019-01-12
+//	  enabledDate: 2019-01-08
```
xh3b4sd: Should the date contain the time of the day? How is the format there? We can have multiple releases being changed at the same day.

xh3b4sd (approved):
Pretty neat.

rossf7 (approved):
@kopiczko LGTM. Thanks for adding the examples. ❤️ 

Yes we should start using the validation. Especially in the `application` CRDs since they are new and customer facing.

pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -0,0 +1,196 @@
...
+							{Raw: []byte(`"disabled"`)},
+							{Raw: []byte(`"eol"`)},
+						},
+					},
+					"release": {
+						Type: "object",
+						Properties: map[string]apiextensionsv1beta1.JSONSchemaProps{
+							"name": {
+								Type:      "string",
+								MinLength: to.Int64P(3),
```
JosephSalisbury: For my curiosity, why 3?

pkg/apis/release/v1alpha1/release_cycle_types.go
kopiczko: I don't see a problem with having multiple releases being changed the same day. It serves only informational purpose. Is having exact time useful? So far we always had 12:00 so I decided to skip that. This is also validated as a date.

PR #201, opened 2019-02-08
release: fix date marshalling

kopiczko:
This is based on @TheoBrigitte findings here https://github.com/giantswarm/apiextensions/pull/200:

> This PR solves an issue where an operator throw the following error when watching ReleaseCycle CR.
>
> ```
> E0207 14:39:35.302146   31346 streamwatcher.go:109] Unable to decode an event from the watch stream: unable to decode watch event: v1alpha1.ReleaseCycle.Spec: v1alpha1.ReleaseCycleSpec.EnabledDate: DisabledDate: u
> nmarshalerDecoder: parsing time ""2019-01-12"" as ""2006-01-02T15:04:05Z07:00"": cannot parse """ as "T", error found in #10 byte of ...|019-01-12","enabledD|..., bigger context ...|8c89a5c143ce"},"spec":{"disable
> dDate":"2019-01-12","enabledDate":"2019-01-08","phase":"enabled","rel|...
> ```
> This is due to the fact that by default `time` object are decoded using `time.RFC3339` in [`time.UnmarshalJSON`](https://golang.org/src/time/time.go?s=37288:37335#L1230) which is incompatible with the json-schema [date](http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3.1) format.
>
> `time.RFC3339` expects : `2019-01-12T08:00:00Z`
> json-schema `date` expects : `2019-01-12`

Closes #200


pkg/apis/release/v1alpha1/deep_copy.go
```diff
@@ -4,12 +4,27 @@ import (
...
+func (d *DeepCopyDate) UnmarshalJSON(data []byte) error {
+	// Ignore null, like in the main JSON package.
+	if string(data) == "null" {
+		return nil
+	}
+
+	// Error masking is skipped here as this will go trough generated
+	// unmarshaling code.
+	var err error
+	d.Time, err = time.Parse(`"2006-01-02"`, string(data))
```
xh3b4sd: I do not comprehend the back and forth. What Theo tried to do sounded pretty reasonable to me. Now you seem to change that again. I actually had a last question in the spec that did not get answered I guess. I think there should be a day time as well. We are likely to have multiple releases at the same day. Why is it not important to have a specific time present?

kopiczko: There is no back and forth. This should be a date according to the spec. I don't know why we should change the format when there is a reasonable solution to the problem.

I remember myself answering this question but let me do that again. We can have multiple releases on the same date. No problem. `enabledDate` and `disabledDate` serve only informational purposes. They are not used with any business logic so far. I think it's clear that time we assigned to releases was artificial hour (most releases we have are released at noon UTC). Why should we keep that? It's only misleading IMO.

kopiczko: BTW I'd talk to @TheoBrigitte about that if you think it's a back and forth between us (which it isn't) but he's sick today.

xh3b4sd: All good. Thanks for the explanation again.

xh3b4sd (approved):
When this is how it should be, let's go.

pkg/apis/release/v1alpha1/deep_copy_test.go
```diff
@@ -0,0 +1,61 @@
...
+			type JSONObject struct {
+				TestDate DeepCopyDate `json:"testDate"`
+			}
+
+			var jsonObject JSONObject
+			err := json.Unmarshal([]byte(tc.json), &jsonObject)
+
+			switch {
+			case err == nil && tc.errorMatcher == nil:
+				// correct; carry on
```
tomahawk28: Why don't we compare the data in here?

kopiczko: Because I was lazy. PTALA.

tomahawk28: 👍

tomahawk28 (approved):

TheoBrigitte (approved):
LGTM

PR #202, opened 2019-02-11
add app and appCatalog

tomahawk28:
- Adding e2e testing structure for `app`, `appCatalog` CR. 
I'm not sure how to test this structure before run inside e2e testing.


helm/apiextensions-app-e2e-chart/values.yaml
```diff
@@ -0,0 +1,23 @@
...
+appCatalog:
+  title: "myapp-catalog"
+  description: "giantswarm app catalog"
+  logoUrl: "http://giantswarm.com/catalog-logo.png"
+  storage:
+    type: "helm"
+    url: "https://giantswarm.github.com/sample-catalog/"
+
+namespace: "default"
+appOperator:
```
rossf7: Let's move this above namespace so all the values are ordered.

helm/apiextensions-app-e2e-chart/values.yaml
```diff
@@ -0,0 +1,23 @@
+app:
+  name: "myapp"
+  namespace: "giantswarm"
+  config:
+    configMap:
+      name: "app-values"
+      namespace: "giantswarm"
```
rossf7: I have a worry about this as we're setting these values in the spec but the configmap and secret are not created as part of the chart.

Can you add templates for these? As we will need them to check that the values support works correctly.

rossf7 (commented):
Direction is good just a couple of comments. Also to fix the CI error you need to create the CNR application in Quay.

helm/apiextensions-app-e2e-chart/values.yaml
tomahawk28: Thanks, I forgot that

helm/apiextensions-app-e2e-chart/values.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+appCatalog:
+  title: "myapp-catalog"
+  description: "giantswarm app catalog"
+  logoUrl: "http://giantswarm.com/catalog-logo.png"
+  storage:
+    type: "helm"
+    url: "https://giantswarm.github.com/sample-catalog/"
+
+configMap:
```
tomahawk28: @rossf7 I'm not sure how to contain it so copy from ChartConfig.

rossf7: Yes this can be anything really. I just based it on the installations repo structure.  So we test we handle the YAML correctly.

Looks good and the values can be overridden in the e2e test if we want to test other values.

rossf7 (approved):
LGTM

tomahawk28:
I'm getting this error, 
```
Error: plugin "registry" exited with error
2019/02/11 13:47:41 retrying task 'helm-push' due to error (exit status 2)
usage: cnr [-h]
           {show,inspect,logout,plugins,helm,pull,list,delete-package,version,push,login,config,channel}
           ...
cnr: error: 403 Client Error: FORBIDDEN for url: https://quay.io/cnr/api/v1/packages/giantswarm/apiextensions-app-e2e-chart?force=true
```
Should I create quay application repo?

rossf7:
@tomahawk28 Yes please create the application repo. That should fix the CI error.

PR #209, opened 2019-02-15
Compare version.LastTransitionTime and only fallback to Date

tuommaki:
Version date has been deprecated and the real comparable value is
LastTransitionTime. In case last transition time is not present, fallback to
compare Date.

### DIFF
--- a/pkg/apis/provider/v1alpha1/status_funcs.go
+++ b/pkg/apis/provider/v1alpha1/status_funcs.go
@@ -74,7 +74,7 @@ func (s StatusCluster) LatestVersion() string {
 	latest := s.Versions[0]
 
 	for _, v := range s.Versions {
-		if latest.Date.Before(v.Date) {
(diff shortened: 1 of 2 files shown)


xh3b4sd (approved):

pkg/apis/provider/v1alpha1/status_funcs.go
```diff
@@ -74,7 +74,7 @@ func (s StatusCluster) LatestVersion() string {
 	latest := s.Versions[0]
 
 	for _, v := range s.Versions {
-		if latest.Date.Before(v.Date) {
+		if latest.LastTransitionTime.Time.Before(v.LastTransitionTime.Time) || latest.Date.Before(v.Date) {
```
tomahawk28: if we compare time already, why we need to compare `v.date`?

tuommaki: There is possibility that some clusters haven't been migrated yet to a version of an operator that provides `LastTransitionTime` field so that's why there's still a fallback to date field. There's another PR that completely removes it but we are not there yet.

PR #210, opened 2019-02-20
Add KVM.NodeIndexes field to KVM Status object

tuommaki:
Unique [unsigned] integer number / index is required for each node for
different use cases. Right now the most pressing one is for iSCSI initiator
name.

As this is very platform specific at the moment, it is only added for KVM. It
is also added as a separate map distinct from Spec masters/workers and status
nodes because this information is needed before the nodes created and it must
be maintained in the cluster. It should also not affect future work when
refactoring the structuring of worker nodes.


tuommaki:
Towards https://github.com/giantswarm/giantswarm/issues/5327

kopiczko (approved):

pkg/apis/provider/v1alpha1/kvm_types.go
```diff
@@ -128,7 +128,13 @@ type KVMConfigSpecVersionBundle struct {
 }
 
 type KVMConfigStatus struct {
-	Cluster StatusCluster `json:"cluster" yaml:"cluster"`
+	Cluster StatusCluster      `json:"cluster" yaml:"cluster"`
+	KVM     KVMConfigStatusKVM `json:"kvm" yaml:"kvm"`
+}
+
+type KVMConfigStatusKVM struct {
+	// NodeIndexes is a map from nodeID -> nodeIndex.
```
kopiczko:
```suggestion
	// NodeIndexes is a map from nodeID -> nodeIndex. This is used to create deterministic iSCSI initiator names.
```

xh3b4sd (approved):

pkg/apis/provider/v1alpha1/kvm_types.go
tuommaki: Can be used for other stuff as well, but accepted :slightly_smiling_face:

PR #212, opened 2019-02-22
Add context and inCluster to kubeconfig

rossf7:
Implementation of spec PR https://github.com/giantswarm/giantswarm/pull/5313.

Adds the context name and the `InCluster` flag. For VOO the app CRs just need in cluster to be true. The `kubeconfig` library will then load the credentials of the `app-operator` service account.


pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -156,10 +157,20 @@ type AppSpecConfigSecret struct {
 }
 
 type AppSpecKubeConfig struct {
+	// Context is the kubeconfig context.
+	Context AppSpecKubeConfigContext `json:"context" yaml:"context"`
```
rossf7: This should have been added before but got missed somehow.

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -84,6 +84,7 @@ func NewAppTypeMeta() metav1.TypeMeta {
 //        kubeConfig:
 //          context:
 //            name: "giantswarm-12345"
+//          inCluster: false
```
pipo02mix: Just a question, when it is `true` the context and secret does not have sense?

pipo02mix (approved):
LGTM

pkg/apis/application/v1alpha1/app_types.go
TheoBrigitte: If yes then it would be nice to comment about this behavior.

TheoBrigitte (approved):
LGTM

pkg/apis/application/v1alpha1/app_types.go
puja108: I somehow cannot review, but I was not aware that we agreed to add an `InCluster` field there. I thought we had said we create a kubeconfig that points to incluster config so it works the same and we do not have 2 ways of doing things. Anyway, I'm not super against this option, it should just be documented that you either set InCluster to true or you give it a kubeconfig and context, and I think it would make sense then to move the inCluster field above context as context and secret go together

rossf7: Yes if its true then the context and secret are not needed. It does need better docs and I'll move the field up.

I think the flag makes the integration easier. There is less config needed on the app CRs and the kubeconfig library can be smart enough to choose the right credentials. Otherwise we'd need a kubeconfig resource in release-operator as well as cluster-operator.

@puja108 It was the final change to the spec PR. I did ask for more feedback but I could have pinged more.

rossf7:
Should be good now.

pkg/apis/application/v1alpha1/app_types.go
puja108: nah, it's ok, my fault for not looking at that PR close enough, checked today and it was definitely in your comment to me, so entirely my fault ;)

and InCluster functionality is not a very uncommon case so having an option for it is nice

PR #216, opened 2019-02-28
Remove quote for bool type

tomahawk28:
- To remove errors from deployment as below
```
E0226 16:11:16.349363       1 streamwatcher.go:109] Unable to decode an event from the watch stream: unable to decode watch event: v1alpha1.App.Spec: v1alpha1.AppSpec.KubeConfig: v1alpha1.AppSpecKubeConfig.InCluster: ReadBool: expect t or f, but found ", error found in #10 byte of ...|Cluster":"false","se|..., bigger context ...|e":"","namespace":""}},"kubeConfig":{"inCluster":"false","secret":{"name":"kubeconfig-secret","names|...
```


rossf7 (approved):
LGTM, good catch.

PR #218, opened 2019-03-01
extend AWSConfig status to track route table IDs

xh3b4sd:
We want to separate the CF resource magic in the `aws-operator`. One thing that is weird is that we lookup the necessary route table IDs multiple times, where we can actually make only one request. It is also that the implementation design right now is now nice because fetching the current state is mixed with the template adapters. This does not make sense and is not good. I want to suggest a separate status resource that simply fetches route table IDs and write them into the CR status.


kopiczko:
Why this information shouldn't be propagated with controllercontext?

xh3b4sd:
I think I will do this. I crafted the PR before we talked.

PR #222, opened 2019-03-28
release: fix example parentVersion

kopiczko:


pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -254,7 +254,7 @@ func NewReleaseTypeMeta() metav1.TypeMeta {
 //	      version: "0.5.0"
 //	    - name: "cluster-operator"
 //	      version: "0.10.0"
-//	  parentVersion: "6.2.1"
+//	  parentVersion: "6.0.1"
```
kopiczko: It should be lower than the `version`.

xh3b4sd (approved):

PR #223, opened 2019-03-28
release: make types cluster scoped

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5278

### DIFF
--- a/pkg/apis/release/v1alpha1/release_cycle_types.go
+++ b/pkg/apis/release/v1alpha1/release_cycle_types.go
@@ -128,6 +128,7 @@ func NewReleaseCycleTypeMeta() metav1.TypeMeta {
 }
 
 // +genclient
+// +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // ReleaseCycle CRs might look something like the following.
--- a/pkg/apis/release/v1alpha1/release_types.go
+++ b/pkg/apis/release/v1alpha1/release_types.go
@@ -229,6 +229,7 @@ func NewReleaseTypeMeta() metav1.TypeMeta {
 }
 
 // +genclient
+// +genclient:nonNamespaced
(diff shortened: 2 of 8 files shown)


rossf7 (approved):

xh3b4sd:
Not too related to this PR but I found this. Why aren't the release types all scoped under the same domain? 

```
kubectl get crd
NAME                                     CREATED AT
...
releasecycles.release.giantswarm.io      2019-02-12T08:48:54Z
releases.core.giantswarm.io              2019-02-11T19:31:39Z
releases.release.giantswarm.io           2019-02-14T08:18:12Z
...
```

kopiczko:
`releases.core.giantswarm.io` was removed from this repo a while ago. This is what `release-operator` was previously creating. We create CRDs while running operators but never delete them when they stop watching them (which makes sense because we couldn't redeploy).

kopiczko:
I'm deleting this from all installations.

xh3b4sd:
Ok, but why are not `Release` and `ReleaseCycle` kind in `release.giantswarm.io`?

kopiczko:
> Ok, but why are not `Release` and `ReleaseCycle` kind in `release.giantswarm.io`?

They are. I'm deleting `releases.core.giantswarm.io`. This is the object that was removed.

xh3b4sd:
Well yeah, but `ReleaseCycle` is in `releasecycles.release.giantswarm.io` and `Release` is in `releases.release.giantswarm.io`. I am asking because we try to figure the scope out for Node Pools. See https://github.com/giantswarm/giantswarm/pull/5683#discussion_r269762185.

kopiczko:
The first part is name of the resource `releasecycles` or `releases`. The rest is the group `release.giantswarm.io` which is common for both.

In general we (magic) thought objects that should be versioned together should be in the same group.

This probably should be stated in some internal docs.

xh3b4sd:
Uh ok. We will notice that at some point in Team Spirit then I guess.

kopiczko:
Deleted `releases.core.giantswarm.io` on all installations.

PR #228, opened 2019-04-05
Remove unused part in charts

tomahawk28:
- remove `kubeconfig` from a spec of the chart.


pkg/apis/application/v1alpha1/chart_types.go
```diff
@@ -103,9 +103,6 @@ type ChartSpec struct {
 	// Name is the name of the Helm chart to be deployed.
 	// e.g. kubernetes-prometheus
 	Name string `json:"name" yaml:"name"`
-	// KubeConfig is the kubeconfig to connect to the cluster when deploying
-	// the app.
-	KubeConfig ChartSpecKubeConfig `json:"kubeConfig" yaml:"kubeConfig"`
```
kopiczko: So how is the kubeconfig referenced now?

tomahawk28: In `app` spec, we have `kubeconfig` information so app-operator could use it to access to tenant cluster. But `chart` do not need `kubeconfig` since it's installed already inside of target destination.

kopiczko: Thanks.

kopiczko (approved):

PR #230, opened 2019-04-22
Add MD5 checksum for values to the chart CRD status

rossf7:
Towards giantswarm/giantswarm#4949

In the `release` resource in `chart-operator` we don't get back exactly the same values that we provide.

e.g. `memoryInMb: 17179.869184` is rounded down to `memoryInMb: 17179.87`

So this adds an MD5 checksum to the status of the chart CR. This will be used to detect when new values need to be passed to Tiller.


MarcelMue (approved):

pkg/apis/application/v1alpha1/chart_types.go
```diff
@@ -156,6 +156,8 @@ type ChartStatus struct {
 	Reason string `json:"reason,omitempty" yaml:"reason,omitempty"`
 	// Release is the status of the Helm release for the deployed chart.
 	Release ChartStatusRelease `json:"release" yaml:"release"`
+	// Values is the status of the values for the deployed chart.
+	Values ChartStatusValues `json:"values" yaml:"values"`
```
tomahawk28: if we only need to show it when its deployed chart, can we put `omitempty` in here?

rossf7: Good point but I'm not sure that will work with the code generation. AIUI so far we haven't used `omitempty` in apiextensions.

But the empty fields in CRs are a bit ugly and might be something we care about when exposing CRs to customers. I think this is a topic for sig-operator and maybe sig-ux.

rossf7:
Going with this to get the VOO changes in.

pkg/apis/application/v1alpha1/chart_types.go
kopiczko: I think this is against the latest guidelines. We shouldn't use status information for reconciliation. It's also useless from the user perspective. I'd put that information into annotations.

rossf7: @kopiczko Ack. I know there have been fmt PRs but I'd missed that.

> It's also useless from the user perspective. I'd put that information into annotations.

I'm fine with using an annotation for this. I like that we're being more sophisticated with how we use metadata. I'll do a PR to remove this field.

kopiczko: It isn't in fmt. It's here if you missed that https://github.com/giantswarm/giantswarm/blob/master/sigs/operator/controller_context.md#cr-scope

PR #231, opened 2019-04-23
add types for node pool provider configuration

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/5833. Follow up of https://github.com/giantswarm/giantswarm/pull/5683.


MarcelMue (approved):
I have a limited understanding here, as I did not join the latest discussions. I don't see anything obviously  wrong here.

pkg/apis/provider/aws/v1alpha1/cluster_spec_types.go
```diff
@@ -0,0 +1,82 @@
...
+//       region: eu-central-1
+//
+type ClusterSpec struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Cluster           ClusterSpecCluster  `json:"cluster" yaml:"cluster"`
+	Provider          ClusterSpecProvider `json:"provider" yaml:"provider"`
+}
+
+type ClusterSpecCluster struct {
```
kopiczko: Isn't this type supposed to be shared between providers?

pkg/apis/provider/aws/v1alpha1/doc.go
```diff
@@ -0,0 +1,4 @@
+// +k8s:deepcopy-gen=package,register
+
+// +groupName=aws.provider.giantswarm.io
```
kopiczko: I said that once already and I still dislike we have separate groups for all providers. This means that again we have separate CRDs for all providers. Why then we can reuse existing types?

kopiczko (dismissed):

pkg/apis/provider/aws/v1alpha1/cluster_spec_types.go
tuommaki: Information as such is common between different providers, but for the beginning I'd maybe still keep these under provider specific type. There aren't that many fields and implementation is always provider specific anyway.

pkg/apis/provider/aws/v1alpha1/cluster_spec_types.go
```diff
@@ -0,0 +1,82 @@
...
+//         version: 4.9.0
+//     provider:
+//       credentialSecret:
+//         name: credential-default
+//         namespace: giantswarm
+//       master:
+//         instanceType: m4.large
+//       region: eu-central-1
+//
+type ClusterSpec struct {
```
tuommaki: Hmm... Thinking about this now, maybe the `AWSProviderSpec` would actually make sense here (as is the case with upstream implementations as well)? That would then be `v1alpha1.AWSProviderSpec` that's embedded in upstream `v1alpha1.Cluster` type's `ProviderSpec` field.

pkg/apis/provider/aws/v1alpha1/cluster_status_types.go
```diff
@@ -0,0 +1,57 @@
...
+	LastTransitionTime DeepCopyTime `json:"lastTransitionTime" yaml:"lastTransitionTime"`
+	Version            string       `json:"version" yaml:"version"`
+}
+
+type ClusterStatusProvider struct {
+	Network ClusterStatusProviderNetwork `json:"network" yaml:"network"`
+}
+
+type ClusterStatusProviderNetwork struct {
+	CIDR string `json:"cidr" yaml:"cidr"`
```
tuommaki: Upstream type has `NetworkRanges` and `CIDRBlocks` - would it make sense to align naming with that?

https://github.com/kubernetes-sigs/cluster-api/blob/release-0.1/pkg/apis/cluster/v1alpha1/cluster_types.go#L77-L81

I'm also thinking here: Does it make sense to track all cluster's network ranges in this status or shall node pools' ranges be separated? Technically those node pools' subnets must be configured to control plane VPCs network ranges so it wouldn't be that wrong to have them here.

pkg/apis/provider/aws/v1alpha1/doc.go
tuommaki: We don't create CRDs here. Only Go types with generated deep copy functions. This type gets serialized and embedded into Cluster API's CR objects. The provider specific extension that is embedded in e.g. Cluster CR, is always provider specific and therefore cannot be shared.

pkg/apis/provider/aws/v1alpha1/cluster_spec_types.go
xh3b4sd: Also, we do not implement this for KVM nor Azure now.

pkg/apis/provider/aws/v1alpha1/doc.go
xh3b4sd: What Tuomas said. The group for these types is less important at this point.

pkg/apis/provider/aws/v1alpha1/cluster_spec_types.go
xh3b4sd: We can name it like that. No strong feelings.

pkg/apis/provider/aws/v1alpha1/cluster_status_types.go
xh3b4sd: No strong feelings on naming. The whole ranges are configured already in the Cluster type AFAIK. This here is only the selected status for the cluster. The status for the node pools should be kept in the other types and when we need information we collect it on demand.

pkg/apis/provider/aws/v1alpha1/cluster_spec_types.go
xh3b4sd: Done

pkg/apis/provider/aws/v1alpha1/cluster_status_types.go
xh3b4sd: Let's keep that for now and then we iterate.

xh3b4sd:
Heads up. 

* Discussed with Tuomas. 
* Renamed structures. 
* Renamed API groups to have it nicer even though it is not critical.
* Moved files around. 

Please have another look.

pkg/apis/provider/aws/v1alpha1/doc.go
kopiczko: Okay understood. Do we need groupName directive here at all?

tuommaki (approved):
LGTM. Let's iterate when something comes to our mind.

pkg/apis/cluster/v1alpha1/aws_provider_spec_types.go
```diff
@@ -0,0 +1,82 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// AWSProviderSpec is the structure put into the provider spec of the Cluster
+// API's Cluster type. There it is tracked as serialized raw extension.
+//
+//     kind: AWSProviderSpec
+//     apiVersion: cluster.giantswarm.io/v1alpha1
```
kopiczko: I think this is misleading. Why do we have `kind` and `apiVersion` in here? Same applies to the status.

pkg/apis/cluster/v1alpha1/aws_provider_spec_types.go
```diff
@@ -0,0 +1,82 @@
...
+//       name: 8y5kc
+//     cluster:
+//       description: my fancy cluster
+//       dns:
+//         domain: gauss.eu-central-1.aws.gigantic.io
+//       oidc:
+//         claims:
+//           username: email
+//           groups: groups
+//         clientId: foobar-dex-client
```
kopiczko:
```suggestion
//         clientID: foobar-dex-client
```

pkg/apis/cluster/v1alpha1/aws_provider_spec_types.go
xh3b4sd: We can drop the type meta if that makes you happy. We want to keep the object metadata for labels and such.

kopiczko: Yes it makes me happy. Is there a reason why labels shouldn't be set on the "main" object?

xh3b4sd: I tried to drop it but it is required by some k8s magic. 

```
# github.com/giantswarm/apiextensions/pkg/apis/cluster/v1alpha1
pkg/apis/cluster/v1alpha1/register.go:18:18: cannot use AWSProviderSpec literal (type *AWSProviderSpec) as type runtime.Object in array or slice literal:
	*AWSProviderSpec does not implement runtime.Object (missing GetObjectKind method)
pkg/apis/cluster/v1alpha1/register.go:19:20: cannot use AWSProviderStatus literal (type *AWSProviderStatus) as type runtime.Object in array or slice literal:
	*AWSProviderStatus does not implement runtime.Object (missing GetObjectKind method)
pkg/apis/cluster/v1alpha1/zz_generated.deepcopy.go:30:5: out.TypeMeta undefined (type *AWSProviderSpec has no field or method TypeMeta)
pkg/apis/cluster/v1alpha1/zz_generated.deepcopy.go:30:19: in.TypeMeta undefined (type *AWSProviderSpec has no field or method TypeMeta)
pkg/apis/cluster/v1alpha1/zz_generated.deepcopy.go:50:3: cannot use c (type *AWSProviderSpec) as type runtime.Object in return argument:
	*AWSProviderSpec does not implement runtime.Object (missing GetObjectKind method)
pkg/apis/cluster/v1alpha1/zz_generated.deepcopy.go:192:5: out.TypeMeta undefined (type *AWSProviderStatus has no field or method TypeMeta)
pkg/apis/cluster/v1alpha1/zz_generated.deepcopy.go:192:19: in.TypeMeta undefined (type *AWSProviderStatus has no field or method TypeMeta)
pkg/apis/cluster/v1alpha1/zz_generated.deepcopy.go:212:3: cannot use c (type *AWSProviderStatus) as type runtime.Object in return argument:
	*AWSProviderStatus does not implement runtime.Object (missing GetObjectKind method)
?   	github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1	[no test files]
ok  	github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1	(cached)
ok  	github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1	(cached)
ok  	github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1	(cached)
```

xh3b4sd: I am fine with that.

pkg/apis/cluster/v1alpha1/register.go
```diff
@@ -0,0 +1,40 @@
...
+	group   = "cluster.giantswarm.io"
+	version = "v1alpha1"
+)
+
+// knownTypes is the full list of objects to register with the scheme. It
+// should contain pointers of zero values of all custom objects and custom
+// object lists in the group version.
+var knownTypes = []runtime.Object{
+	&AWSProviderSpec{},
+	&AWSProviderStatus{},
```
kopiczko: I believe when we drop the meta those objects shouldn't be registered. I also find that misleading.

pkg/apis/provider/aws/v1alpha1/doc.go
xh3b4sd: I removed the comment.

pkg/apis/cluster/v1alpha1/register.go
xh3b4sd: I think the registration is necessary for some `RawExtension` encoding magic. I want to keep it as it is right now.

pkg/apis/cluster/v1alpha1/aws_provider_spec_types.go
kopiczko: I'm not. This is the spec. Spec doesn't have TypeMeta or ObjectMeta. I don't understand how this is useful. I also don't understand why we need DeepCopy on this object at all. AIUI it will be used only for unmarshaling purposes.

tuommaki: I don't know specifically, but guess that DeepCopy magic etc. is needed in order to make k8s `runtime.RawExtension` work correctly in all cases. This is what all the upstream projects also do with same type (i.e. the provider extensions used in cluster api implementations).

kopiczko: It's the first time I hear about `runtime.RawExtension`. If this is the case then fine. Can you point me somewhere code so I can learn more?

pkg/apis/cluster/v1alpha1/aws_provider_status_types.go
```diff
@@ -0,0 +1,57 @@
...
+//     metadata:
+//       name: 8y5kc
+//     cluster:
+//       conditions:
+//       - lastTransitionTime: "2019-03-25T17:10:09.333633991Z"
+//         type: Created
+//       id: 8y5kc
+//       versions:
+//       - lastTransitionTime: "2019-03-25T17:10:09.995948706Z"
+//         version: 4.9.0
```
kopiczko: Why `version` needs to be an array? Will we keep a history of version upgrades in here?

pkg/apis/cluster/v1alpha1/aws_provider_spec_types.go
```diff
@@ -0,0 +1,82 @@
...
+//     cluster:
+//       description: my fancy cluster
+//       dns:
+//         domain: gauss.eu-central-1.aws.gigantic.io
+//       oidc:
+//         claims:
+//           username: email
+//           groups: groups
+//         clientID: foobar-dex-client
+//         issuerUrl: https://dex.8y5kc.fr-east-1.foobar.example.com
```
kopiczko:
```suggestion
//         issuerURL: https://dex.8y5kc.fr-east-1.foobar.example.com
```

pkg/apis/cluster/v1alpha1/aws_provider_status_types.go
xh3b4sd: Yes, we want to keep the version history.

pkg/apis/cluster/v1alpha1/aws_provider_spec_types.go
xh3b4sd: Maybe one of these here. 

* https://github.com/kubermatic/machine-controller/tree/master/pkg/providerconfig
* https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/master/pkg/apis/awsprovider/v1alpha1

xh3b4sd:
Going with this now. We can iterate.

PR #234, opened 2019-04-24
Revert Add MD5 checksum for values to the chart CRD status

rossf7:
See https://github.com/giantswarm/apiextensions/pull/230#discussion_r278166144 we will use an annotation for storing the checksum.


kopiczko (approved):
That makes sense to me, thanks.

MarcelMue (approved):
To labels we go :D

kopiczko:
> To labels we go :D

Annotations

MarcelMue:
> > To labels we go :D
> 
> Annotations

To annotations we go :D

rossf7:
No problem and it improves the design. I did consider annotations but I turned right when I should have turned left.

PR #239, opened 2019-05-14
add AZ to master spec

xh3b4sd:


tuommaki (approved):
Yup. Needed so that we can define master AZ.

MarcelMue (approved):

PR #240, opened 2019-05-16
Add New*CRD() functions for Cluster API types

tuommaki:
Provide Go structs for Cluster & MachineDeployment types so that these can be
passed to corresponding controller implementation for registration.

### DIFF
--- a/pkg/apis/cluster/v1alpha1/cluster_api_crds.go
+++ b/pkg/apis/cluster/v1alpha1/cluster_api_crds.go
@@ -1,0 +1,95 @@
+package v1alpha1
+
+import (
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NewClusterCRD returns a new custom resource definition for Cluster (from
+// Cluster API). This might look something like the following.
+//
+//     apiVersion: apiextensions.k8s.io/v1beta1
+//     kind: CustomResourceDefinition
+//     metadata:
+//       name: clusters.cluster.k8s.io
+//     spec:
+//       group: cluster.k8s.io
+//       scope: Namespaced
+//       version: v1alpha1
+//       names:
+//         kind: Cluster
+//         plural: clusters
+//         singular: cluster
+//       subresources:
+//         status: {}
+//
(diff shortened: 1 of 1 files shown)


pkg/apis/cluster/v1alpha1/cluster_api_crds.go
```diff
@@ -0,0 +1,95 @@
...
+//
+func NewClusterCRD() *apiextensionsv1beta1.CustomResourceDefinition {
+	return &apiextensionsv1beta1.CustomResourceDefinition{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: apiextensionsv1beta1.SchemeGroupVersion.String(),
+			Kind:       "CustomResourceDefinition",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: map[string]string{
+				"controller-tools.k8s.io": "1.0",
```
xh3b4sd: What is with these labels? Anything that has to be updated later on? Should we drop a comment somewhere?

xh3b4sd (approved):

pkg/apis/cluster/v1alpha1/cluster_api_crds.go
tuommaki: They come from upstream. I guess they are for compatibility with specific versions of `clusterctl` etc. I'm not sure if we need a comment though? In the long run I think we need to find a better approach to maintain these CRDs and for now they are probably more or less updated every now and then.

kopiczko (approved):

pkg/apis/cluster/v1alpha1/cluster_api_crds.go
xh3b4sd: I was also wondering about this. We should support some auto update behaviour when the CRD changes.

kopiczko: We already ran into problems with release-operator because validation was updated.

xh3b4sd: Was or wasn't? Do you think we should have CRD update support or not?

kopiczko: Validation in the CRD definition _was_ updated and the CRD wasn't updated by the operatorkit. I think we'll have to support CRD update.

PR #241, opened 2019-05-17
Add missing Version to Cluster & MachineDeployment CRDs

tuommaki:


tuommaki:
Human error, already verified this to fix the problem. Going with this on my own.

PR #242, opened 2019-06-05
Be a bit specific about the namespace

oponder:
If I understood it right, the convention we're taking with the userconfig configmap is to put it in the tenant cluster namespace on the control plane. So I adjust the comment to save a future soul some searching.

Let me know if I got it wrong @puja108 @rossf7


puja108:
I would make it even more concrete saying it can be either cluster namespace or a customer owned namespace than can be shared between clusters of a single or even of all orgs as that will be sth that we will need to support at some point soon

PR #244, opened 2019-06-14
Extract Cluster status type to be shared among different providers

tuommaki:
Extract provider independent cluster status type so that it can be shared among
different provider extension status types. This allows provider independent
status resource implementation.


pkg/apis/cluster/v1alpha1/common_cluster_status_types.go
```diff
@@ -0,0 +1,19 @@
+package v1alpha1
+
+// StatusCluster is shared type to contain provider independent cluster status
+// information.
+type StatusCluster struct {
```
xh3b4sd: Can we name the type according to the file? I am not happy with the naming as we have it for the old types. It is super confusing because there is no symmetry whatsoever. 

```suggestion
type CommonClusterStatus struct {
```

pkg/apis/cluster/v1alpha1/common_cluster_status_types.go
```diff
@@ -0,0 +1,19 @@
+package v1alpha1
+
+// CommonStatusCluster is shared type to contain provider independent cluster status
+// information.
+type CommonStatusCluster struct {
```
xh3b4sd: This was not what I suggested. But almost there buddy. When we have `AWSClusterStatus` we should have `CommonClusterStatus`. That was the idea.

tuommaki: :man_facepalming: Right. One minute. :slightly_smiling_face:

xh3b4sd (approved):

PR #246, opened 2019-06-17
Fix example issuerURL from AWS cluster provider extension spec

tuommaki:
IssuerURL cannot point to to-be-created cluster at this point and when it
points to another domain, there shouldn't be cluster ID from to-be-created
cluster present.

### DIFF
--- a/pkg/apis/cluster/v1alpha1/aws_cluster_spec_types.go
+++ b/pkg/apis/cluster/v1alpha1/aws_cluster_spec_types.go
@@ -25,7 +25,7 @@ import (
 //           username: email
 //           groups: groups
 //         clientID: foobar-dex-client
-//         issuerURL: https://dex.8y5kc.fr-east-1.foobar.example.com
+//         issuerURL: https://dex.gatekeeper.fr-east-1.aws.example.com
 //     provider:
 //       credentialSecret:
 //         name: credential-default
--- a/pkg/apis/cluster/v1alpha1/aws_cluster_spec_types.go
+++ b/pkg/apis/cluster/v1alpha1/aws_cluster_spec_types.go
@@ -25,7 +25,7 @@ import (
 //           username: email
 //           groups: groups
 //         clientID: foobar-dex-client
-//         issuerURL: https://dex.gatekeeper.fr-east-1.aws.example.com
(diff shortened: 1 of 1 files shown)


pkg/apis/cluster/v1alpha1/aws_cluster_spec_types.go
```diff
@@ -25,7 +25,7 @@ import (
 //           username: email
 //           groups: groups
 //         clientID: foobar-dex-client
-//         issuerURL: https://dex.8y5kc.fr-east-1.foobar.example.com
+//         issuerURL: https://dex.gatekeeper.fr-east-1.aws.example.com
```
rossf7:
```suggestion
//         issuerURL: https://dex.gatekeeper.eu-central-1.aws.example.com
```

rossf7 (approved):
LGTM but being pedantic about the URL ;)

kopiczko:
Is it that by default it will point to `https://dex.8y5kc.fr-east-1.foobar.example.com`?

tuommaki:
> Is it that by default it will point to https://dex.8y5kc.fr-east-1.foobar.example.com?

No. We haven't yet defined *default* for this, but if it was self-pointing, it would be `https://dex.8y5kc.gauss.eu-central-1.aws.gigantic.io` (see the base domain above there).

kopiczko:
Yeah I only meant `gatekeeper` part. Why it shouldn't be the cluster ID. I don't follow. Is it because the ID will generated and isn't available at the time of creation? Will it be possible to deploy dex inside the cluster?

tuommaki:
> Yeah I only meant gatekeeper part. Why it shouldn't be the cluster ID. I don't follow. Is it because the ID will generated and isn't available at the time of creation? Will it be possible to deploy dex inside the cluster?

Two things:
1. In most of the cases it appears that customer has separate systems (or maybe clusters) that are running OIDC and that is not the cluster that is being created.
2. Right now it is not possible to know cluster ID in call cases when you create a cluster from CR (this might / can be changed).

It is possible to configure OIDC to use dex deployed on the same cluster, but this must be done in two steps then. First create the cluster and then configure OIDC & recreate master.

kopiczko (approved):
Thanks for the explanations!

PR #252, opened 2019-06-26
drop machine deployment status types

xh3b4sd:
A [`MachineDeployment`](https://godoc.org/sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1#MachineDeploymentStatus) does not have a provider status as we anticipated when designing our internal types. We found ways around this asymmetric limitation and can drop the unused types now. @tuommaki please double check.


tuommaki (approved):
Yep. We can live without this.

* Cluster & Node Pool IDs and allocated Node Pool subnet are labels on `MachineDeployment`.
* AZ specific subnets are computed with deterministic algorithm so there's no need to persist this.
* ASG name can be queried from AWS with Cluster ID and Node Pool ID tags.
* Versions are managed as labels on tenant cluster nodes.
* Scaling parameters / numbers are present in upstream `MachineDeployment` CR spec & status.

PR #256, opened 2019-06-28
sort status versions to have the latest version on top

xh3b4sd:
Follow up of https://github.com/giantswarm/apiextensions/pull/254.

### DIFF
--- a/pkg/apis/cluster/v1alpha1/common_cluster_status_funcs.go
+++ b/pkg/apis/cluster/v1alpha1/common_cluster_status_funcs.go
@@ -156,27 +156,32 @@ func withCondition(conditions []CommonClusterStatusCondition, search string, rep
 }
 
 // withVersion computes a list of version history using the given list and new
-// version structure to append. withVersion also limits total amount of elements
-// in the list by cutting off the tail with respect to the limit parameter.
+// version structure to append. withVersion also limits the total amount of
+// elements in the list by cutting off the tail with respect to the limit
+// parameter.
 func withVersion(versions []CommonClusterStatusVersion, version CommonClusterStatusVersion, limit int) []CommonClusterStatusVersion {
 	if hasVersion(versions, version.Version) {
 		return versions
 	}
 
+	// Create a copy to not manipulate the input list.
 	var newVersions []CommonClusterStatusVersion
-
-	start := 0
-	if len(versions) >= limit {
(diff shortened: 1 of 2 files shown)


pkg/apis/cluster/v1alpha1/common_cluster_status_funcs.go
```diff
@@ -156,27 +156,32 @@ func withCondition(conditions []CommonClusterStatusCondition, search string, rep
...
 
-	sort.Sort(sortClusterStatusVersionsByDate(versions))
+	// Sort the versions in a way that the newest version, namely the one with the
+	// highest timestamp, is at the top of the list.
+	sort.Sort(sort.Reverse(sortClusterStatusVersionsByDate(newVersions)))
 
-	for i := start; i < len(versions); i++ {
-		newVersions = append(newVersions, versions[i])
+	// Calculate the index for capping the list in the next step.
+	l := limit - 1
```
MarcelMue: Pretty sure there is an edge case here when the limit is 1 or 0. Should never happen in reality to be fair.

pkg/apis/cluster/v1alpha1/common_cluster_status_funcs_test.go
```diff
@@ -399,12 +402,12 @@ func Test_Provider_Status_withVersion(t *testing.T) {
...
 
-	for _, tc := range testCases {
-		t.Run(tc.Name, func(t *testing.T) {
+	for i, tc := range testCases {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
 			versions := withVersion(tc.Versions, tc.Version, tc.Limit)
 
 			if !reflect.DeepEqual(versions, tc.ExpectedVersions) {
-				t.Fatalf("expected %#v got %#v", tc.ExpectedVersions, versions)
+				t.Fatalf("\n\n%s\n", cmp.Diff(versions, tc.ExpectedVersions))
```
MarcelMue: Love the `cmp` diff we do now <3

MarcelMue (approved):

pkg/apis/cluster/v1alpha1/common_cluster_status_funcs.go
xh3b4sd: When anyone ever wants that, the unit tests will cause a panic or just fail. Good enough for me now.

PR #257, opened 2019-06-29
rework status condition list management

xh3b4sd:
Follow up of https://github.com/giantswarm/apiextensions/pull/254.

### DIFF
--- a/pkg/apis/cluster/v1alpha1/common_cluster_status_funcs.go
+++ b/pkg/apis/cluster/v1alpha1/common_cluster_status_funcs.go
@@ -74,19 +74,39 @@ func (s CommonClusterStatus) LatestVersion() string {
 }
 
 func (s CommonClusterStatus) WithCreatedCondition() []CommonClusterStatusCondition {
-	return withCondition(s.Conditions, ClusterStatusConditionCreating, ClusterStatusConditionCreated, time.Now())
+	newCondition := CommonClusterStatusCondition{
+		LastTransitionTime: DeepCopyTime{time.Now()},
+		Condition:          ClusterStatusConditionCreated,
+	}
+
+	return withCondition(s.Conditions, newCondition, ClusterConditionLimit)
 }
 
 func (s CommonClusterStatus) WithCreatingCondition() []CommonClusterStatusCondition {
-	return withCondition(s.Conditions, ClusterStatusConditionCreated, ClusterStatusConditionCreating, time.Now())
+	newCondition := CommonClusterStatusCondition{
+		LastTransitionTime: DeepCopyTime{time.Now()},
+		Condition:          ClusterStatusConditionCreating,
+	}
+
+	return withCondition(s.Conditions, newCondition, ClusterConditionLimit)
 }
 
 func (s CommonClusterStatus) WithDeletedCondition() []CommonClusterStatusCondition {
-	return withCondition(s.Conditions, ClusterStatusConditionDeleting, ClusterStatusConditionDeleted, time.Now())
+	newCondition := CommonClusterStatusCondition{
+		LastTransitionTime: DeepCopyTime{time.Now()},
+		Condition:          ClusterStatusConditionDeleted,
+	}
+
+	return withCondition(s.Conditions, newCondition, ClusterConditionLimit)
 }
 
 func (s CommonClusterStatus) WithDeletingCondition() []CommonClusterStatusCondition {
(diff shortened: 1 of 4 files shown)


pkg/apis/cluster/v1alpha1/common_cluster_status_funcs_test.go
```diff
@@ -125,52 +128,358 @@ func Test_Provider_Status_LatestVersion(t *testing.T) {
...
+					Condition:          ClusterStatusConditionCreated,
+				},
+				{
+					LastTransitionTime: DeepCopyTime{time.Unix(20, 0)},
+					Condition:          ClusterStatusConditionCreating,
+				},
+			},
+		},
+		{
+			name: "case 7: the third update of the tenant cluster ends",
```
MarcelMue: From what I see we unit test the same functionality 3 times in a row here - any reason why this is tested for 3 consecutive upgrades?

Wouldn't it be more interesting to test some edge case behaviour - like an upgrade starting before the previous one is finished?

xh3b4sd: The tests go step by step through the lifecycle of a tenant cluster. 3 upgrades are tested to ensure the tracking and capping works properly. The tests here represent the edge cases. What else do you have in mind?

MarcelMue: Like I said, I think having two upgrades start before the first one is finished could be interesting. No blocker though.

MarcelMue (approved):

pkg/apis/cluster/v1alpha1/common_cluster_status_funcs_test.go
xh3b4sd: > I think having two upgrades start before the first one is finished could be interesting

That should never happen but I can add a test case for it.

MarcelMue: I am aware that it never should happen, but it has happened before with updates that got stuck and went unfixed until the next update.
It would therefor be nice to see that this works in a sane manner if it happens (so debugging is easy).

xh3b4sd: I found a cool way to inject missing conditions. Please check if this is sufficient.

pkg/apis/cluster/v1alpha1/common_cluster_status_funcs.go
```diff
@@ -136,23 +179,137 @@ func hasVersion(versions []CommonClusterStatusVersion, search string) bool {
...
+		}
+
+		conditionPairs = append(conditionPairs, p)
+	}
+
+	// Here the list is finally flattened again and its pairs are limitted to the
+	// input parameter.
+	var limittedList []CommonClusterStatusCondition
+	for _, p := range conditionPairs {
+		// We cmpute the pair limit here for the total number of items. This is why
```
tuommaki:
```suggestion
		// We compute the pair limit here for the total number of items. This is why
```

pkg/apis/cluster/v1alpha1/common_cluster_status_funcs.go
```diff
@@ -136,23 +179,137 @@ func hasVersion(versions []CommonClusterStatusVersion, search string) bool {
...
+		}
 	}
 
-	for _, c := range conditions {
-		if c.Condition == search {
-			continue
+	return false
+}
+
+func withCondition(conditions []CommonClusterStatusCondition, condition CommonClusterStatusCondition, limit int) []CommonClusterStatusCondition {
```
tuommaki: I'd like to have a comment here to explain what this function is expected to do. It has several nested loops that require quit a lot of effort to decompile in head so summary of what is expected from this function would be great.

pkg/apis/cluster/v1alpha1/common_cluster_status_funcs_test.go
MarcelMue: I reread this and it looks good to me. I already approved previously :+1:

pkg/apis/cluster/v1alpha1/common_cluster_status_funcs.go
xh3b4sd: I made such a great effort to add comments inline and forgot this one. 😄

xh3b4sd: Done

tuommaki (approved):

PR #258, opened 2019-07-01
Add cluster availability zones into provider spec

tuommaki:
In order to have clean separation for management of cluster's availability
zones, they are defined in the spec as desired state. Intention here is then
that when MachineDeployment controller reconciles a new CR, it ensures that
corresponding Cluster has all required availability zones defined. This also
makes implementation of dynamic AZ / public subnet infrastructure management
easier and more straightforward while having the on-demand aspect of it.


pkg/apis/cluster/v1alpha1/aws_cluster_spec_types.go
```diff
@@ -67,9 +70,10 @@ type AWSClusterSpecClusterOIDCClaims struct {
 }
 
 type AWSClusterSpecProvider struct {
-	CredentialSecret AWSClusterSpecProviderCredentialSecret `json:"credentialSecret" yaml:"credentialSecret"`
-	Master           AWSClusterSpecProviderMaster           `json:"master" yaml:"master"`
-	Region           string                                 `json:"region" yaml:"region"`
+	AvailabilityZones []string                               `json:"availabilityZones" yaml:"availabilityZones"`
```
MarcelMue: Just a note - nothing to do here:
We should somehow ensure that this list is deduplicated.

MarcelMue (approved):

xh3b4sd (changes requested):
I am not sure this is the right thing to do. We already have the AZs in the Machine Deployments. See the example of https://godoc.org/github.com/giantswarm/apiextensions/pkg/apis/cluster/v1alpha1#AWSMachineDeploymentSpec.

tuommaki:
> I am not sure this is the right thing to do. We already have the AZs in the Machine Deployments.

Yes, we do and that is requirement in order to be able to know what is needed for cluster as well. As we discussed this with Marcel, the logic is following:

**Cluster controller**
- Reconciled `Cluster` CR and ensures that tenant cluster control plane exists and holds required infrastructure.
- Fully functional tenant cluster control plane can exist without a single `MachineDeployment` and therefore it doesn't need to "care" about `MachineDeployments`.

**MachineDeployment controller**
- In order to have a valid `MachineDeployment` CR, a corresponding `Cluster` must exist.
- `MachineDeployment` required tenant cluster control plane infrastructure in order to be able to create node pool CF stack.
- A resource run for `MachineDeployment` CR ensures that specified AZs for given node pool exists in desired state of corresponding `Cluster` CR. Vice versa on deletion.

This design keeps communication / dependence between these two CRs and controllers one-way only (MD requires Cluster but not other way round) and also allows simple and straightforward implementation of dynamic public subnet infrastructure management.

kopiczko:
> A resource run for MachineDeployment CR ensures that specified AZs for given node pool exists in desired state of corresponding Cluster CR. Vice versa on deletion.

I understand it isn't a big concern now but how can we ensure that this won't be set to some random value by the user in the future?

tuommaki:
> I understand it isn't a big concern now but how can we ensure that this won't be set to some random value by the user in the future?

My idea here was that either we don't guard this and `MachineDeployment` controller would reconcile it into desired [correct] state eventually or alternatively if there's a need, we add an admission controller.

tuommaki:
We discussed with Tim about this and he had a strong concern on duplication of information here and wrong direction of effects. After some discussion back and forth, the outcome was that for now we just list `MachineDeployments` for given cluster in `Cluster` controller and gather the list of required AZs from there instead of adding this information into the cluster spec. Therefore I'm dropping this.

PR #260, opened 2019-07-22
Add OpenAPI V3 validation to app CRD

rossf7:

Adds validation for the app CRD and aims to align with the release group CRDs.

### DIFF
--- a/pkg/apis/application/v1alpha1/app_types.go
+++ b/pkg/apis/application/v1alpha1/app_types.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	"github.com/giantswarm/to"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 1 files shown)


pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -9,21 +10,214 @@ const (
...
+	OpenAPIV3Schema: &apiextensionsv1beta1.JSONSchemaProps{
+		Properties: map[string]apiextensionsv1beta1.JSONSchemaProps{
+			"spec": {
+				Type: "object",
+				Properties: map[string]apiextensionsv1beta1.JSONSchemaProps{
+					"catalog": {
+						Type:      "string",
+						MinLength: to.Int64P(3),
+					},
+					"config": {
```
rossf7: I have a doubt about optional properties. The config and secret blocks are both optional. Do I need to define that somehow?

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -9,21 +10,214 @@ const (
...
+//   group: application.giantswarm.io
+//   names:
+//     kind: App
+//     listKind: AppList
+//     plural: apps
+//     singular: app
+//   scope: Namespaced
+//   subresources:
+//     status: {}
+//   validation:
```
rossf7: The example CRD is pretty huge now. Should we include the validation?

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -9,21 +10,214 @@ const (
...
+//                     name:
+//                       minLength: 3
+//                       type: string
+//                     namespace:
+//                       minLength: 3
+//                       type: string
+//                   type: object
+//               type: object
+//             version:
+//               pattern: ^\d+\.\d+\.\d+$
```
kopiczko: I think this is too strict. What about `1.2.3-mywodnerfulgitsha12345`?

kopiczko (commented):
I think in general mid term it would be good to parse text in runtime instead of generating text from structs. Those structs for validation are nightmare.

pkg/apis/application/v1alpha1/app_types.go
rossf7: Good catch. I'll make it more permissive.

rossf7 (commented):
> Those structs for validation are nightmare.

Agreed. So the master copy would be in YAML? Where would that be stored? Could we use a constant like we do for the current struct?

kopiczko:
> > Those structs for validation are nightmare.
> 
> Agreed. So the master copy would be in YAML? Where would that be stored? Could we use a constant like we do for the current struct?

I think go string const is by far more readable. We have that in the comment already which we can replace with the const. If the const is parsed in `init` functions we'll catch errors/panics in the unit test phase of the CI. WDYT?

rossf7:
OK cool thanks. I'll update this to use a string const. Yes I think using an init function works.

We started working on this during SWH but it's also part of our team story.

cc @sslavic

kopiczko (approved):

kopiczko:
LGTM when version validation is relaxed.

pkg/apis/application/v1alpha1/app_types.go
sslavic: From https://swagger.io/docs/specification/data-models/data-types/#required

> By default, all object properties are optional. You can specify the required properties in the required list:

rossf7: @sslavic Ack and thanks that link is super useful.

rossf7:
Closing in favor of #279.

PR #261, opened 2019-07-23
Define schema for AppCatalog CRD

sslavic:

### DIFF
--- a/pkg/apis/application/v1alpha1/app_catalog_types.go
+++ b/pkg/apis/application/v1alpha1/app_catalog_types.go
@@ -8,55 +8,56 @@ import (
 )
 
 const appCatalogCRDValidationSchema = `
-  openAPIV3Schema:
-	type: object
-	properties:
-	  spec:
-		type: object
-		properties:
-		  title:
-			type: string
-		  description:
-			type: string
-		  config:
-			type: object
-			properties:
-			  configMap:
-				type: object
-				properties:
-				  name:
-					type: string
-				  namespace:
-					type: string
-				required: ["name", "namespace"]
-			  secret:
-				type: object
-				properties:
-				  name:
-					type: string
-				  namespace:
-					type: string
-				required: ["name", "namespace"]
-		  logoURL:
-			type: string
-			format: uri
-		  storage:
-			type: object
-			properties:
-			  type:
-				type: string
-				enum:
-				  - helm
-			  URL:
-				type: string
-				format: uri
-		required: ["title", "description", "logoURL", "storage"]
+openAPIV3Schema:
+  type: object
+  properties:
+	spec:
+	  type: object
+	  properties:
+		title:
+		  type: string
+		description:
+		  type: string
+		config:
+		  type: object
+		  properties:
+			configMap:
+			  type: object
+			  properties:
+			    name:
+				  type: string
+				namespace:
+				  type: string
+			  required: ["name", "namespace"]
+			secret:
+			  type: object
+			  properties:
+				name:
+				  type: string
+				namespace:
+				  type: string
+			  required: ["name", "namespace"]
+		logoURL:
+		  type: string
+		  format: uri
+		storage:
+		  type: object
+		  properties:
+			type:
+			  type: string
+			  enum:
+				- helm
+			URL:
+			  type: string
+			  format: uri
+	  required: ["title", "description", "logoURL", "storage"]
 `
 
 var appCatalogCRDValidation *apiextensionsv1beta1.CustomResourceValidation
 
 func init() {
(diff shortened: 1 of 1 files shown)


pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -1,10 +1,66 @@
...
+			type: object
+			properties:
+			  configMap:
+				type: object
+				properties:
+				  name:
+					type: string
+				  namespace:
+					type: string
+				required: ["name", "namespace"]
```
rossf7: A query on the required as I have gaps in my knowledge here.

The configMap and secret blocks should be optional. But this is to ensure if they are set then both the name and namespace are present?

rossf7 (commented):
Direction is great here <3.

Just a query on the required fields. Please also ask SIG Operator for a review.

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -1,10 +1,66 @@
...
 
 import (
+	"encoding/json"
+
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const appCatalogCRDValidationSchema = `
+  openAPIV3Schema:
```
rossf7:
```suggestion
openAPIV3Schema:
```

Yes the formatting looks strange but two spaces should be OK. Please remove the leading whitespace as that's how we usually format these constants.

pkg/apis/application/v1alpha1/app_catalog_types.go
sslavic: Yes. `config` object definition doesn't have it's own `required` specification, so all of it's properties are optional by default. `config` itself is optional property too. Note one edge case is supported - `config` can be present without any inner properties there (neither `configMap` nor `secret`).

rossf7: Ack, thanks. I think that's OK as an edge case for `config`. Having the validation that both name and namespace are present is nice.

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -1,10 +1,66 @@
...
+			  URL:
+				type: string
+				format: uri
+		required: ["title", "description", "logoURL", "storage"]
+`
+
+var appCatalogCRDValidation *apiextensionsv1beta1.CustomResourceValidation
+
+func init() {
+	if err := json.Unmarshal([]byte(appCatalogCRDValidationSchema), &appCatalogCRDValidation); err != nil {
```
rossf7: Nesting the error handling like this is something we try to avoid. This is a simple case but when we use it we also need to remember when we can use it.

```
err := json.Unmarshal([]byte(appCatalogCRDValidationSchema, &appCatalogCRDValidation)
if err != nil {
  panic(err)
}
```

rossf7 (approved):
LGTM once whitespace issue is resolved and error handling is updated.

pkg/apis/application/v1alpha1/app_catalog_types.go
sslavic: Fixed formatting - there were tabs, replaced them with two spaces.

pkg/apis/application/v1alpha1/app_catalog_types.go
sslavic: Fixed

tomahawk28 (approved):
LGTM :)

rossf7:
@sslavic I think we can go with this. The review request from sig-operator is mainly important for knowledge sharing.

I'm going to align my PR #260 with this. Could you add the validation for the chart CRD? We can then close giantswarm/giantswarm#5407.

sslavic:
@rossf7 clear, thanks, merging.

Sure, I will add schema for chart CRD too.

rossf7:
@sslavic Thx much appreciated! ❤️

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -1,10 +1,67 @@
 package v1alpha1
 
 import (
+	"encoding/json"
+
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const appCatalogCRDValidationSchema = `
```
kopiczko: Why not define whole CRD here?

PR #263, opened 2019-07-25
Define schema for AppCatalog CRD

sslavic:
Towards giantswarm/giantswarm#5407

### DIFF
--- a/pkg/apis/application/v1alpha1/app_catalog_types.go
+++ b/pkg/apis/application/v1alpha1/app_catalog_types.go
@@ -1,9 +1,105 @@
 package v1alpha1
 
 import (
+	"encoding/json"
+
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+const appCatalogCRDJSON = `
+{
+	"apiVersion": "apiextensions.k8s.io/v1beta1",
+	"kind": "CustomResourceDefinition",
+	"metadata": {
+		"name": "appcatalogs.application.giantswarm.io"
+	},
+	"spec": {
+		"group": "application.giantswarm.io",
+		"scope": "Cluster",
+		"version": "v1alpha1",
+		"names": {
+			"kind": "AppCatalog",
+			"plural": "appcatalogs",
+			"singular": "appcatalog"
+		},
+		"validation": {
+			"openAPIV3Schema": {
+				"type": "object",
+				"properties": {
+					"spec": {
+						"type": "object",
+						"properties": {
+							"title": {
+								"type": "string"
+							},
+							"description": {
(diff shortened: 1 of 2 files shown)


pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -1,10 +1,106 @@
 package v1alpha1
 
 import (
+	"encoding/json"
+
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const appCatalogCRDJSON = `
```
kopiczko: Why not yaml?

rossf7 (changes requested):
What I care about here is maintainability.  This JSON will be hard to maintain and is a step backwards from what we have now IMO.

I would like us to take a step back and just add the OpenAPI validation rules as YAML. That is what this issue is about. I agree that the Go structs for the rules are difficult to work with.

If we want to change how we manage CRDs that should be a separate issue and we should have consensus in sig-operator.

kopiczko:
I'm personally fine with defining whole CRD in the const. I'm even biased that this the way to go ATM. But choosing JSON as a format is controversial.

Having that in a separate issue/PR makes sense though.

rossf7:
Yes my main point here is we should make smaller iterations and have consensus before making major changes. This started out as a SWH task to add some validation rules and has escalated a lot. 

I also don't think there is consensus on using JSON for this.

pkg/apis/application/v1alpha1/app_catalog_types.go
sslavic: There are no yaml tags in the types, only json.
More details on slack thread: https://gigantic.slack.com/archives/C2MS2VB37/p1564049998054900

kopiczko: Ack. Yes that is a problem to some degree. We'd probably have to unmarshal to `map[string]interface{}` then marshal to json and then unmarshal to struct.

rossf7:
@sslavic I think this can be closed?

sslavic:
@rossf7 we discussed yesterday on Slack, options were
- to delete this PR/branch and created new one, or
- use this PR/branch, rebase it and overwrite the commit with new one that's aligned with chart CRD 

I'm fine with either, @tomahawk28 suggested he will chose latter (see https://gigantic.slack.com/archives/C76JX6YLQ/p1567415625013000?thread_ts=1567415386.011000&cid=C76JX6YLQ )

rossf7:
@sslavic But I think the validation was added here? 😕 https://github.com/giantswarm/apiextensions/pull/280

@tomahawk28 Do you still need this PR?

tomahawk28:
@sslavic @rossf7 Sorry, I started with another branch. We can remove this branch.

PR #264, opened 2019-07-25
Define schema for Chart CRD

sslavic:
Towards giantswarm/giantswarm#5407

### DIFF
--- a/pkg/apis/application/v1alpha1/chart_types.go
+++ b/pkg/apis/application/v1alpha1/chart_types.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	"github.com/ghodss/yaml"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -8,6 +9,67 @@ import (
 const (
 	kindChart = "Chart"
 )
+
+const chartCRDYAML = `
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: charts.application.giantswarm.io
+spec:
+  group: application.giantswarm.io
+  scope: Namespaced
+  version: v1alpha1
+  names:
+    kind: Chart
+    plural: charts
+    singular: chart
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          type: object
+          properties:
+            name:
+              type: string
+            namespace:
+              type: string
+            config:
+              type: object
+              properties:
+                configMap:
+                  type: object
+                  properties:
+                    name:
+                      type: string
+                    namespace:
+                      type: string
+                    resourceVersion:
+                      type: string
+                  required: ["name", "namespace"]
+                secret:
+                  type: object
+                  properties:
+                    name:
+                      type: string
+                    namespace:
+                      type: string
+                    resourceVersion:
+                      type: string
+                  required: ["name", "namespace"]
+            tarballURL:
+              type: string
+              format: uri
+          required: ["name", "namespace", "tarballURL"]
+`
+
+var chartCRD *apiextensionsv1beta1.CustomResourceDefinition
+
+func init() {
+	err := yaml.Unmarshal([]byte(chartCRDYAML), &chartCRD)
+	if err != nil {
+		panic(err)
+	}
+}
 
 // NewChartCRD returns a new custom resource definition for Chart.
 // This might look something like the following.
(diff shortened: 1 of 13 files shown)


oponder:
@sslavic I'm not 100% up to date here, but this looks ok.. have we tested it anywhere yet?

sslavic:
@oponder thanks for review; sorry, but I still have to update this and related CRD schema definition PRs to align them with most recent agreement made on sig-operator sync with @xh3b4sd and @kopiczko - we want to give it a try to define whole CRD with schema in yaml, with likely intermediary parsing to map, before getting structs; upon updating will ping once more for another round of review.

sslavic:
@kopiczko @xh3b4sd @oponder et al PTAL

pkg/apis/application/v1alpha1/chart_types.go
```diff
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	"github.com/ghodss/yaml"
```
kopiczko: We already have `gopkg.in/yaml.v2` in vendor. Why do we have to pull another one?

pkg/apis/application/v1alpha1/chart_types.go
```diff
@@ -9,6 +10,68 @@ const (
...
+            tarballURL:
+              type: string
+              format: uri
+          required: ["name", "namespace", "tarballURL"]
+`
+
+var chartCRD *apiextensionsv1beta1.CustomResourceDefinition
+
+func init() {
+	err := yaml.Unmarshal([]byte(chartCRDYAML), &chartCRD)
```
kopiczko: So no need for extra intermediate marshaling?

sslavic: github.com/ghodss/yaml builds on top of gopkg.in/yaml.v2 to implement that patter.
See http://ghodss.com/2014/the-right-way-to-handle-yaml-in-golang/

pkg/apis/application/v1alpha1/chart_types.go
sslavic: See https://github.com/giantswarm/apiextensions/pull/264#discussion_r313266854

pkg/apis/application/v1alpha1/chart_types.go
kopiczko: Nice I like that 👍

kopiczko (approved):
I like that way more than what we have so far. LGTM

sslavic:
Thanks for review @kopiczko!

Won't merge yet - to further test I've update apiextensions dependency in chart-operator branch to this branch, and chart-operator e2e integration tests started hanging and failing after timeout; still have to determine if it's due to minikube (old k8s version and/or CRD related configuration) or really due to some validation failing now that validation schema is in place; if former, migration to e2ectl/KIND might help.

pkg/apis/application/v1alpha1/chart_types_test.go
```diff
@@ -0,0 +1,10 @@
+package v1alpha1
+
+import "testing"
+
+func Test_NewChartCRD(t *testing.T) {
```
tomahawk28: Can we add a unittest for `openAPIV3Schema` in action?

tomahawk28 (approved):

pkg/apis/application/v1alpha1/chart_types_test.go
sslavic: Unit test is there for exported `NewChartCRD` function and it gives us some confidence that the Chart CRD yaml definition loading during initialization works, no yaml syntax errors etc. I'm not aware, if you know please share - is there a simple interface for testing just validation as it would be done when processing request in k8s, but without spinning up k8s cluster?

If not, I guess you're suggesting we should add an e2e integration test where with given k8s cluster and Chart CRD with validation schema installed we can have validation passing and failing depending on submitted (in)valid Chart CRs. Will seek for more opinions on this idea, how to best implement and organize, considering fact that there are already existing related e2e integration tests e.g. in `chart-operator`.

I actually relied on those existing e2e integration tests. In a branch of `chart-operator` repository I updated `apiextensions` dependency to this branch to test existing e2e integration tests still pass. Since those are still using minikube with k8s 1.11 or older, I got hit by https://github.com/kubernetes/kubernetes/issues/65293 so I removed top level `type: object` (btw thanks for helping uncover this earlier today). Also through one of the e2e integration tests in `chart-operator` I found that `resourceVersion` field of `configMap`+`secret` is not actually required so I updated the Chart CRD validation schema definition accordingly. Might have to migrate `chart-operator` e2e integration tests to use KIND, to verify Chart CRD works also for newer k8s.

tomahawk28: >  is there a simple interface for testing just validation as it would be done when processing request in k8s, but without spinning up k8s cluster?

I have no idea of it, sorry. Perhaps it's better to have it on e2e test as you said.

kopiczko: Please don't overdo this.

sslavic:
Update apiextensions dependency on branches, and now KIND based chart-operator integration tests still pass https://circleci.com/workflow-run/2120c259-7e0b-4c10-9542-95b500dce383

rossf7 (approved):
LGTM and nice to have this verified in chart-operator e2e.

PR #268, opened 2019-08-12
bumping k8s patch version to 4

tobbbles:
Since #266 was an invalid approach, I'm bumping the patch version here directly.

Comparrison between tags:
https://github.com/kubernetes/apimachinery/compare/kubernetes-1.13.2..kubernetes-1.13.4

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -46,19 +46,19 @@ required = [
 
 [[constraint]]
   name = "k8s.io/apiextensions-apiserver"
-  version = "kubernetes-1.13.2"
+  version = "kubernetes-1.13.4"
 
 [[constraint]]
   name = "k8s.io/apimachinery"
-  version = "kubernetes-1.13.2"
+  version = "kubernetes-1.13.4"
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "kubernetes-1.13.2"
+  version = "kubernetes-1.13.4"
 
 [[constraint]]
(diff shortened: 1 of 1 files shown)


kopiczko (approved):
That should be fine. It will have big implications though because we'll have to update Gopkg.toml in many projects. We should update aws-operator at least and announce in news-dev in Slack. Please get at least one more approval.

kopiczko:
It would be also good to hear from @rossf7

tuommaki (approved):
The diff in the upstream should be pretty much nobrainer here. As mentioned by Pawel though, this might require changes to many other projects so prepare for a sweep to update everything (to be aligned).

Also as mentioned, @rossf7 (or someone working with Helm right now) should probably check this in case this has any implications to helm client dependency (we recently faced many issues there).

kopiczko:
I'm afraid @rossf7 is on vacation.

xh3b4sd:
I mean, why are we doing this here anyway? Feels like a waste of time and I do not see which problem this solves.

rossf7:
Yes, I'd like to understand why we need to upgrade the 1.13 patch version.

I'm assigned to upgrading the k8s version to 1.15 and the related helm upgrade in https://github.com/giantswarm/giantswarm/issues/6302. I spent a lot of time there but I can't get it working. 😞 

We still need to do the 1.15 upgrade but I don't know when I'll get time to work on it again.

kopiczko:
I guess we can close that?

PR #269, opened 2019-08-22
Add chart-operator app CR into app operator e2e test

tomahawk28:
To support https://github.com/giantswarm/app-operator/pull/120. 

We now need to add chart-operator app CR, and related appCatalog CR inside e2e test.


helm/apiextensions-app-e2e-chart/templates/app.yaml
```diff
@@ -1,33 +1,38 @@
...
-  namespace: "{{ .Values.app.namespace }}"
-  {{- if .Values.app.kubeconfig}}
+  catalog: "{{ .catalog }}"
+  name: "{{ .name }}"
+  namespace: "{{ .namespace }}"
+  {{- if .kubeconfig}}
   kubeConfig:
-    inCluster: {{ .Values.app.kubeconfig.inCluster }}
+    inCluster: {{ .kubeconfig.inCluster }}
+    {{- if not .kubeconfig.inCluster }}
```
tomahawk28: If it supports inCluster, no need to create secrets.

helm/apiextensions-app-e2e-chart/templates/configmap.yaml
```diff
@@ -1,10 +1,13 @@
-{{ if .Values.app.config.configMap.name }}
+{{- range .Values.apps }}
+  {{- if .config }}
+    {{ if (.config.configMap.name) and (index $.Values.configMaps .config.configMap.name) }}
```
tomahawk28: `(index $.Values.configMaps .config.configMap.name)` is for data sanity checks.

helm/apiextensions-app-e2e-chart/templates/configmap.yaml
```diff
@@ -1,10 +1,13 @@
...
 metadata:
-  name: "{{ .Values.app.config.configMap.name }}"
-  namespace: "{{ .Values.app.config.configMap.namespace }}"
-data:
-  values: |
-    {{ .Values.configMap.values }}
-{{ end }}
+  name: "{{ .config.configMap.name }}"
+  namespace: "{{ .config.configMap.namespace }}"
+  values: {{ toYaml (index $.Values.configMaps .config.configMap.name) | indent 2 }}
```
tomahawk28: I checked both indent 2 and 4 are working, so chose lower number.

helm/apiextensions-app-e2e-chart/values.yaml
```diff
@@ -1,45 +1,58 @@
...
+      secret:
+        name: "app-secrets"
+        namespace: "giantswarm"
+    version: "1.0.0"
+  - name: "chart-operator"
+    namespace: "giantswarm"
+    catalog: "giantswarm-catalog"
+    kubeconfig:
+      inCluster: "true"
+    version: "0.9.0"
```
tomahawk28: This should be hard-coded anyway.

rossf7: But what is this 0.9.0 version? Looks like chartconfig CR but this is chart CR so we should use 1.0.0.

https://github.com/giantswarm/chart-operator/blob/e05c2af301b9e35a25fcf56f3b28b0e20dcfec91/service/controller/chart/v1/version_bundle.go#L18

helm/apiextensions-app-e2e-chart/values.yaml
```diff
@@ -1,45 +1,58 @@
...
-secret:
-  values:
+secrets:
+  app-secrets:
     Installation:
       V1:
         Secret:
           API:
-            token: "test"
+            token: "test"
```
rossf7: There should be a newline here.

rossf7 (approved):
LGTM but I don't get the version change.

helm/apiextensions-app-e2e-chart/values.yaml
tomahawk28: That means I need to update tag ` v1.0.0 ` on chart-operator. I would do that.

rossf7: Ack its because you're pulling the tarball from here. Sorry I get it now. 

https://github.com/giantswarm/giantswarm-catalog

Let's leave it as 0.9.0. Since the chart is being pushed to the catalog and to Quay.

PR #270, opened 2019-08-22
Fix chart-operator version as 0.9.0

tomahawk28:
Following https://github.com/giantswarm/apiextensions/pull/269#discussion_r316692952
> Ack its because you're pulling the tarball from here. Sorry I get it now.
   https://github.com/giantswarm/giantswarm-catalog
   Let's leave it as 0.9.0. Since the chart is being pushed to the catalog and to Quay.


rossf7 (approved):
LGTM

PR #271, opened 2019-08-22
Fix on app chart issues

tomahawk28:
1. We need to move out `.spec.verison` from loop statement. 
2. Version in label was referenced to .spec.version before. 🤦‍♂ 

3. Fix an error from configmap structure


rossf7 (approved):

tomahawk28:
@rossf7 Sorry, I added one more commit onto configmap changes, I found out during manual install.

rossf7 (commented):
Np :) these helm templates are tricky since so much is configurable.

PR #275, opened 2019-08-27
Fix hard-coding on secret name

tomahawk28:
Fixing hard-coding on the secret name, it should follow the name user suggested.


tuommaki (approved):
Sure :+1:

PR #276, opened 2019-08-29
Add status subresource to chart CRD

rossf7:
Adds the status subresource to the CRD YAML. As otherwise UpdateStatus fails in chart-operator.


rossf7:
When I vendor in chart-operator I'll extend the e2e to check the status. As that would have caught this.

tomahawk28 (approved):

PR #279, opened 2019-09-02
Add validation to app CRD

rossf7:

Tested in https://github.com/giantswarm/app-operator/pull/131.

### DIFF
--- a/pkg/apis/application/v1alpha1/app_catalog_types.go
+++ b/pkg/apis/application/v1alpha1/app_catalog_types.go
@@ -60,7 +60,6 @@ spec:
                   required: ["name", "namespace"]
             logoURL:
               type: string
-              format: uri
             storage:
               type: object 
               properties:
--- a/pkg/apis/application/v1alpha1/app_catalog_types.go
+++ b/pkg/apis/application/v1alpha1/app_catalog_types.go
@@ -104,7 +104,7 @@ func NewAppCatalogCRD() *apiextensionsv1beta1.CustomResourceDefinition {
 
 func NewAppCatalogTypeMeta() metav1.TypeMeta {
 	return metav1.TypeMeta{
-		APIVersion: version,
+		APIVersion: SchemeGroupVersion.String(),
 		Kind:       kindAppCatalog,
 	}
 }
--- a/pkg/apis/application/v1alpha1/app_catalog_types.go
+++ b/pkg/apis/application/v1alpha1/app_catalog_types.go
@@ -131,7 +131,7 @@ func NewAppCatalogTypeMeta() metav1.TypeMeta {
 //        secret:
 //          name: "app-catalog-secrets"
 //          namespace: "giantswarm"
-//      logoURL: "https://s.giantswarm.io/..."
+//      logoURL: "/images/repo_icons/incubator.png"
 //      storage:
 //        type: "helm"
 //        URL: "https://giantswarm.github.com/app-catalog/"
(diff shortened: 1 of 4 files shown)


pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -104,7 +103,7 @@ func NewAppCatalogCRD() *apiextensionsv1beta1.CustomResourceDefinition {
 
 func NewAppCatalogTypeMeta() metav1.TypeMeta {
 	return metav1.TypeMeta{
-		APIVersion: version,
+		APIVersion: SchemeGroupVersion.String(),
```
rossf7: Aligning other CRDs with https://github.com/giantswarm/apiextensions/pull/281.

@TheoBrigitte Thx for fixing.

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -60,7 +60,6 @@ spec:
                   required: ["name", "namespace"]
             logoURL:
               type: string
-              format: uri
```
rossf7: The logoURL is just for use in the UI. It caused errors in e2e so I think it's better to remove it. cc @tomahawk28 

```
{"caller":"github.com/giantswarm/app-operator/vendor/github.com/giantswarm/backoff/notifier.go:13","level":"warning","message":"retrying backoff in '5s' due to error","stack":"[{/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:355: } {rpc error: code = Unknown desc = release giantswarm-apiextensions-app-e2e-chart failed: AppCatalog.application.giantswarm.io \"test-app-catalog\" is invalid: []: Invalid value: map[string]interface {}{\"apiVersion\":\"application.giantswarm.io/v1alpha1\", \"kind\":\"AppCatalog\", \"metadata\":map[string]interface {}{\"creationTimestamp\":\"2019-09-03T12:16:31Z\", \"generation\":1, \"labels\":map[string]interface {}{\"app-operator.giantswarm.io/version\":\"1.0.0\"}, \"name\":\"test-app-catalog\", \"uid\":\"9277074e-c80b-4bb1-95a1-85b310091e13\"}, \"spec\":map[string]interface {}{\"description\":\"\", \"logoURL\":\"\", \"storage\":map[string]interface {}{\"URL\":\"https://giantswarm.github.com/sample-catalog\", \"type\":\"helm\"}, \"title\":\"test-app-catalog\"}}: validation failure list:\nspec.logoURL in body must be of type uri: \"\"}]","time":"2019-09-03T12:16:31.681925+00:00"}
--- FAIL: TestAppLifecycle (5.90s)
```

sslavic: Maybe example / test data could be adjusted to be valid - either remove logoURL or make it a dummy but valid URI. Seems to be empty string now or not set so zero value / empty string.

Seems also that `omitempty` tag is missing https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/application/v1alpha1/app_catalog_types.go#L155

tomahawk28 (approved):

pkg/apis/application/v1alpha1/app_catalog_types.go
rossf7: I updated the example to use what we use in installations which is actually a relative path.

On `omitempty` there are only a few instances where we use that. It does create empty fields in the CRs but so far this hasn't been a problem.

PR #280, opened 2019-09-02
Adding appCatalog CRD validation

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/5407

Adding AppCatalog validation.

### DIFF
--- a/pkg/apis/application/v1alpha1/app_catalog_types.go
+++ b/pkg/apis/application/v1alpha1/app_catalog_types.go
@@ -1,9 +1,86 @@
 package v1alpha1
 
 import (
+	"github.com/ghodss/yaml"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+const (
+	kindAppCatalog = "AppCatalog"
+)
+
+const appCatalogCRDYAML = `
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: appcatalogs.application.giantswarm.io
+spec:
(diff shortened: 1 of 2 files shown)


pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -1,10 +1,87 @@
...
+            storage:
+              type: object 
+              properties:
+                type:
+                  type: string
+                URL:
+                  type: string
+                  format: uri 
+              required: ["type", "URL"]
+          required: ["title", "description", "storage"]
```
tomahawk28: I think logoURL is not that important, right?

rossf7: Yes it's for display purposes and maybe not needed on internal catalogs.

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -1,10 +1,87 @@
...
+spec:
+  group: application.giantswarm.io
+  scope: Cluster
+  version: v1alpha1
+  names:
+    kind: AppCatalog
+    plural: appcatalogs
+    singular: appcatalog
+  subresources:
+    status: {}
```
rossf7: This isn't enabled in the current CRD. But I don't think it does any harm to add it.

rossf7 (approved):
LGTM when tested. 

Please make sure this plays nicely with `opsctl ensure appcatalogs`.

tomahawk28:
Tested on ginger. ```opsctl ensure appcatalogs --installation ginger -b appcatalog_validation --user jhbok
Ensuring app catalogs for `ginger`
E0902 14:27:40.523455   37405 portforward.go:376] error copying from local connection to remote stream: tls: use of closed connection
Ensured app catalogs for `ginger`.
Finished.
...

```

PR #284, opened 2019-09-12
New AppCatalogEntry

tomahawk28:
In on-site we agree that we need to have a list of version for each app we are managing, so we can validate the version that customers are using. Later we can expand it to support other ideas around it.


kopiczko:
I don't block this PR but I have a feeling it goes poor's man harbour direction.

rossf7:
This is so we can do validation with OPA for app versions. We discussed this in our team sessions at the onsite. But I think its best if we do a spec PR before merging this PR. I can look at that next week.

tomahawk28:
> But I think its best if we do a spec PR before merging this PR. I can look at that next week.

@rossf7 I think we do not add `deletion` annotation into a spec PR yet either. do you mind to include it too?

rossf7:
> I think we do not add deletion annotation into a spec PR yet either.

@tomahawk28 Yes, good point. I'll add that too.

puja108:
> Doesn't solution like chart museum or harbour provide those features?

Some, like listing versions, it does, others it doesn't. We sadly need this internally for a few use cases and I also still feel a bit bad about it, but I also do not see a way around it right now

puja108:
Let's write this down in a spec and what we need it for before continuing here. Not sure if we could have a way of getting our use cases solved with a chart registry which would be a nicer SoT. Maybe if we need the registry info inside a cluster centrally, instead of a CRD it needs to be a microservice that abstracts away the registry for us. Sth inside the cluster that asks and maybe caches registry SoT smh

corest:
What is the status of this? @tomahawk28

puja108:
This is on hold but we should tackl  it next cycle most probably. Cc @rossf7 @cokiengchiara @piontec

xh3b4sd:

tomahawk28:
@xh3b4sd this is outdated, I don't know what we talked in the latest on-site, we can start it over when we need this.

PR #283, opened 2019-09-04
app: fix yaml output

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/6847

This PR intends to fix the generated yaml output for App CR.

It does so by setting the `omitempty` field tag on optional fields and using pointer for those.

yaml output without this PR:
```yaml
apiVersion: application.giantswarm.io/v1alpha1
kind: App
metadata:
  creationTimestamp: null
  labels:
    app-operator.giantswarm.io/version: 1.0.0
  name: foo-bar
  namespace: giantswarm
spec:
  catalog: baz
  config:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: ""
    inCluster: true
    secret:
      name: ""
      namespace: ""
  name: foo
  namespace: giantswarm
  userConfig:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  version: bar
status:
  appVersion: ""
  release:
    lastDeployed: "0001-01-01T00:00:00Z"
    status: ""
  version: ""
```

yaml output with this PR:
```yaml
apiVersion: application.giantswarm.io/v1alpha1
kind: App
metadata:
  creationTimestamp: null
  labels:
    app-operator.giantswarm.io/version: 1.0.0
  name: foo-bar
  namespace: giantswarm
spec:
  catalog: baz
  kubeConfig:
    inCluster: true
  name: foo
  namespace: giantswarm
  version: bar
status: {}
```

PR #285, opened 2019-09-16
Revert "app: fix yaml output (#283)"

rossf7:
This reverts the changes made by @TheoBrigitte.

I really like that the app CR output is cleaner. The problem is the operator changes got big and it needs logic to handle nils which is easy to miss. See https://github.com/giantswarm/app-operator/pull/136. 

So overall I think it's better to live with the empty fields.


sslavic (approved):
LGTM

tomahawk28 (approved):
Make sense, LGTM.

PR #281, opened 2019-09-02
fix APIVersion in NewAppTypeMeta

TheoBrigitte:
This PR fixes the APIVersion returned by `NewAppTypeMeta`.

Currently `NewAppTypeMeta` returns `v1alpha1` which is wrong and should returns `application.giantswarm.io/v1alpha1`.

I know wonder how the entire thing is working without this. Can someone enlighten me please?

PR #286, opened 2019-09-17
Align release CRDs with app CRDs

sslavic:

### DIFF
--- a/pkg/apis/release/v1alpha1/release_cycle_types.go
+++ b/pkg/apis/release/v1alpha1/release_cycle_types.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	"github.com/ghodss/yaml"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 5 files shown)


pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -14,115 +15,65 @@ const (
...
-			Validation: releaseCycleValidation,
-		},
-	}
+	return releaseCycleCRD.DeepCopy()
 }
 
 func NewReleaseCycleTypeMeta() metav1.TypeMeta {
 	return metav1.TypeMeta{
-		APIVersion: version,
+		APIVersion: SchemeGroupVersion.String(),
```
sslavic: This is also aligned based on https://github.com/giantswarm/apiextensions/pull/281

rossf7 (approved):
LGTM but it would be good to test this vendored in release-operator.

For app-operator it crashed when deploying the change and I needed to add RBAC perms for the status subresource for CRDs.

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -17,213 +17,94 @@ const (
...
+              pattern: "^\\d+\\.\\d+\\.\\d+$"
+            version:
+              type: string
+              minLength: 5
+            required:
+            - changelog
+            - components
+            - parentVersion
+            - version
+`
```
kopiczko: It is missing validation for status

```
//	        status:
//	          type: object
//	          properties:
//	            cycle:
//	              type: object
//	              required:
//	                - phase
//	              properties:
//	                disabledDate:
//	                  type: string
//	                  format: date
//	                enabledDate:
//	                  type: string
//	                  format: date
//	                phase:
//	                  enum:
//	                  - upcoming
//	                  - enabled
//	                  - disabled
//	                  - eol
```

sslavic: Yes, omitted status details part for both CRDs. @rossf7 do you remember why we omitted specifying validation for app CRDs? I asked earlier in #sig-api-machinery to get their thoughts.

kopiczko: I don't understand why we change semantics here.

rossf7: I think reasoning was because the status is set by the operator rather than users.

But no strong opinion. If we want to validate the status I'm fine with that.

kopiczko: I think it's good to rely on some schema here. I think we're going in direction where other components are reading those statuses for communication and monitoring and then having this part versioned is useful.

sslavic: > I don't understand why we change semantics here.

If you're referring to `required` being under `properties` - that was wrong indentation. Fixed it.

kopiczko: I'm referring to `status` validation.

rossf7: Ack. CR status being consumed by other components is a good reason for the validation.

kopiczko (approved):

pkg/apis/release/v1alpha1/release_types.go
sslavic: Added the status details in release CRD schema, and a note in release cycle CRD.

kopiczko: Already approved <3

sslavic:
> LGTM but it would be good to test this vendored in release-operator.

Finally got release-operator [e2e tests to pass](https://circleci.com/gh/giantswarm/release-operator/tree/test-release-crd-as-yaml) - there was a har to spot bug, wrong yaml library was imported.

@rossf7 are you suggesting to test release-operator with updated apiextensions on a test environment/cluster? RBAC stuff I guess would be more reliably tested with e2e tests if release-operator was tested with e2ectl/kind instead of e2e-harness/minikube.

pkg/apis/release/v1alpha1/release_cycle_types.go
```diff
@@ -14,115 +15,67 @@ const (
...
+  version: v1alpha1
+  names:
+    kind: ReleaseCycle
+    plural: releasecycles
+    singular: releasecycle
+  subresources:
+    status: {}
+  validation:
+    # Note: When changing ReleaseCycle CRD schema please make sure to update
+    # ReleaseCycle details in Release CRD schema accordingly.
```
rossf7: <3

rossf7 (approved):
Nice <3 and yes I get now why we want validation on the status.

sslavic:
Just the format (from structs to yaml) has been changed, so IMO no need to test these changes in release-operator on a cluster, confident enough with e2e tests (initially failing with wrong yaml library imported and now) passing. Merging.

PR #287, opened 2019-09-24
Update to kubernetes-1.15.4

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/6302


Gopkg.toml
```diff
@@ -30,35 +30,31 @@ required = [
 
 [[constraint]]
   branch = "master"
-  name = "github.com/giantswarm/microerror"
-
-[[constraint]]
-  branch = "master"
-  name = "github.com/giantswarm/operatorkit"
```
sslavic: This seemed not to be used, it was just creating circular dependency, so I removed it. Let me know if I'm missing here something.

TheoBrigitte (approved):
LGTM when tested

PR #290, opened 2019-10-04
Use new labels format for encryption key in Azure

rossf7:
Labels were changed for AWS but not Azure. So updated randomkeys in deps upgrade broke e2e.


sslavic (approved):
LGTM

PR #291, opened 2019-10-05
Set alt name for internal-api in certconfig

rossf7:
I'm trying to get aws-operator e2e to pass and I noticed the internal API alt name was missing.


rossf7:
Going with a ping.

rossf7:
See https://github.com/giantswarm/e2esetup/pull/39. We should delete this chart if its no longer used.

PR #292, opened 2019-10-09
Added warning for supported golang version

whites11:
Don't know if you think this warning could be useful or not; it could have saved me some debugging time and one help request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,12 @@
 Package apiextensions provides generated Kubernetes clients for the Giant Swarm
 infrastructure.
 
-## Contributing
-
 ## Golang version
 
 The supported version of golang to be used with this project is ~1.11 (tested 1.11.13).
 Using other GO versions could cause the gen.sh script to fail.
+
+## Contributing
 
 ### Adding a New Group and/or Version
 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,11 @@ infrastructure.
 
 ## Contributing
 
+## Golang version
+
+The supported version of golang to be used with this project is ~1.11 (tested 1.11.13).
+Using other GO versions could cause the gen.sh script to fail.
+
 ### Adding a New Group and/or Version
 
 This is example skeleton for adding new group and/or version.


README.md
```diff
@@ -7,6 +7,11 @@ infrastructure.
 
 ## Contributing
 
+## Golang version
+
+The supported version of golang to be used with this project is ~1.11 (tested 1.11.13).
```
tomahawk28: May we could indicate that as ` < 1.11` ? and possibly outside of `Contribution`.

whites11: Not sure about the version, what I know is that 1.13 does not work and I've been suggested to use 1.11.13 (so I'm in your hands about the exact version).
Moved the section out of `Contribution`

sslavic: @whites11 what do you mean 1.13 doesn't work? Recent commits were using recent giantswarm/architect (CLI and runtime / Docker container) versions which use Go 1.13 and most recently 1.13.1 and at least CI builds were passing.

whites11: I mean that running the scripts/gen.sh lead to an error and no `zz_generated...` file was created for my new Resource.

The error was this:

```
F1008 16:29:06.098923  197278 main.go:82] Error: Failed making a parser: unable to add directory "github.com/giantswarm/apiextensions/pkg/apis/database/v1alpha1": unable to import "github.com/giantswarm/apiextensions/pkg/apis/database/v1alpha1": go/build: importGo github.com/giantswarm/apiextensions/pkg/apis/database/v1alpha1: exit status 1
go: finding github.com/giantswarm/apiextensions latest
can't load package: package github.com/giantswarm/apiextensions/pkg/apis/database/v1alpha1: module github.com/giantswarm/apiextensions@latest (v0.0.0-20191005094604-efa5c38da056) found, but does not contain package github.com/giantswarm/apiextensions/pkg/apis/database/v1alpha1
```

and it disappeared as soon as I downgraded to go 1.11.13

kopiczko: Then this got to be fixed. We shouldn't switch Go version for that.

whites11: Okay, giving up my PR then. Thanks for feedback

xh3b4sd: I think we should hang and talk and clarify how anything works now while we switch to go modules. I had the same problem as Cristian had in other projects as well which is why I went back to an older go version. Projects behave differently now and it is complicated to get them work in a consistent way. I am fine with `go1.11.13` now. For me `go1.13.x` does not work.

PR #293, opened 2019-11-06
Added RootVolumeSizeGB field for Azure nodes

whites11:
This edit is related to this issue: https://github.com/giantswarm/giantswarm/issues/7526
Sometimes the customer needs to adjust the size of root disk for master or workers.
I added a new field on the AzureConfig CR to allow for that customization.

### DIFF
--- a/pkg/apis/provider/v1alpha1/azure_types.go
+++ b/pkg/apis/provider/v1alpha1/azure_types.go
@@ -112,7 +112,7 @@ type AzureConfigSpecAzureVirtualNetwork struct {
 type AzureConfigSpecAzureNode struct {
 	// VMSize is the master vm size (e.g. Standard_A1)
 	VMSize string `json:"vmSize" yaml:"vmSize"`
-	// Size of a volume mounted to /var/lib/docker.
+	// DockerVolumeSizeGB is the size of a volume mounted to /var/lib/docker.
 	DockerVolumeSizeGB int `json:"dockerVolumeSizeGB" yaml:"dockerVolumeSizeGB"`
 	// RootVolumeSizeGB is the size of a volume mounted to /.
 	RootVolumeSizeGB int `json:"rootVolumeSizeGB" yaml:"rootVolumeSizeGB"`
--- a/pkg/apis/provider/v1alpha1/azure_types.go
+++ b/pkg/apis/provider/v1alpha1/azure_types.go
@@ -114,7 +114,7 @@ type AzureConfigSpecAzureNode struct {
 	VMSize string `json:"vmSize" yaml:"vmSize"`
 	// Size of a volume mounted to /var/lib/docker.
 	DockerVolumeSizeGB int `json:"dockerVolumeSizeGB" yaml:"dockerVolumeSizeGB"`
-	// Size of a volume mounted to /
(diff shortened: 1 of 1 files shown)


pkg/apis/provider/v1alpha1/azure_types.go
```diff
@@ -114,6 +114,8 @@ type AzureConfigSpecAzureNode struct {
 	VMSize string `json:"vmSize" yaml:"vmSize"`
 	// Size of a volume mounted to /var/lib/docker.
 	DockerVolumeSizeGB int `json:"dockerVolumeSizeGB" yaml:"dockerVolumeSizeGB"`
+	// Size of a volume mounted to /
```
kopiczko:
```suggestion
	// RootVolumeSizeGB is the size of a volume mounted to /.
```

pkg/apis/provider/v1alpha1/azure_types.go
```diff
@@ -114,6 +114,8 @@ type AzureConfigSpecAzureNode struct {
 	VMSize string `json:"vmSize" yaml:"vmSize"`
 	// Size of a volume mounted to /var/lib/docker.
```
kopiczko:
```suggestion
	// DockerVolumeSizeGB is the size of a volume mounted to /var/lib/docker.
```

kopiczko (approved):

pkg/apis/provider/v1alpha1/azure_types.go
```diff
@@ -112,8 +112,10 @@ type AzureConfigSpecAzureVirtualNetwork struct {
 type AzureConfigSpecAzureNode struct {
 	// VMSize is the master vm size (e.g. Standard_A1)
 	VMSize string `json:"vmSize" yaml:"vmSize"`
-	// Size of a volume mounted to /var/lib/docker.
+	// DockerVolumeSizeGB is the size of a volume mounted to /var/lib/docker.
 	DockerVolumeSizeGB int `json:"dockerVolumeSizeGB" yaml:"dockerVolumeSizeGB"`
+	// RootVolumeSizeGB is the size of a volume mounted to /.
+	RootVolumeSizeGB int `json:"rootVolumeSizeGB" yaml:"rootVolumeSizeGB"`
```
tuommaki: Wait! I think this is wrong. Instead of supporting configuration of root volume size, why don't we support `/var/lib/kubelet` where `EmptyDir` lands? That would be a lot more legit. AFAIK one can use root volume only by host mounts and from my understanding that would be really discourage when it's all about just using ephemeral storage for temporary files?

tuommaki: ...that's how we went for Node Pool's workers as well: https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/cluster/v1alpha1/aws_machine_deployment_spec_types.go#L51-L54

whites11: I was thinking about doing both.
I agree it is more legit to have a separate and size-configurable /var/lib/kubelet but having a configurable root disk size won't hurt I reckon.
Opinions more than welcome.

whites11:
We decided to rollback this PR and go on with https://github.com/giantswarm/apiextensions/pull/294

PR #296, opened 2019-11-11
Add PublicIPName to AzureConfig status

tuommaki:
With slight logic change we can eliminate unnecessary public IP allocation in
Azure by re-using existing one from ingress LB. To do this, it makes sense to
store the value in CR status field as it doesn't change after cluster creation.

Towards https://github.com/giantswarm/giantswarm/issues/6636


xh3b4sd (approved):

pkg/apis/provider/v1alpha1/azure_types.go
```diff
@@ -123,7 +123,16 @@ type AzureConfigSpecVersionBundle struct {
...
+	Cluster  StatusCluster             `json:"cluster" yaml:"cluster"`
+	Provider AzureConfigStatusProvider `json:"provider" yaml:"provider"`
+}
+
+type AzureConfigStatusProvider struct {
+	IngressLB AzureConfigStatusProviderIngressLB `json:"ingressLB" yaml:"ingressLB"`
+}
+
+type AzureConfigStatusProviderIngressLB struct {
+	PublicIPName string `json:"publicIPName" yaml:"publicIPName"`
```
kopiczko: Is it name of just the `IP`?

pkg/apis/provider/v1alpha1/azure_types.go
```diff
@@ -123,7 +123,16 @@ type AzureConfigSpecVersionBundle struct {
 }
 
 type AzureConfigStatus struct {
-	Cluster StatusCluster `json:"cluster" yaml:"cluster"`
+	Cluster  StatusCluster             `json:"cluster" yaml:"cluster"`
+	Provider AzureConfigStatusProvider `json:"provider" yaml:"provider"`
+}
+
+type AzureConfigStatusProvider struct {
+	IngressLB AzureConfigStatusProviderIngressLB `json:"ingressLB" yaml:"ingressLB"`
```
kopiczko: Does it make sense to make it `Ingress.LoadBalancer` instead?

kopiczko (approved):

pkg/apis/provider/v1alpha1/azure_types.go
tuommaki: It's name of the public IP (Azure resource as of itself, [this one](https://github.com/giantswarm/azure-operator/blob/master/service/controller/v12/resource/deployment/template/public_load_balancer_setup.json#L49-L64)). See the last row here: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/docs/services/README.md#loadbalancer-annotations

pkg/apis/provider/v1alpha1/azure_types.go
kopiczko: The question should be if this provides more value :D

tuommaki: I don't expect there to be much new fields as we should transition towards Cluster CRs on Azure in the future, but your proposal reads nicer anyway so that's why I'm totally fine converting to it. It's longer to write, but more explicit.

PR #298, opened 2019-11-22
add infrastructure/v1alpha2 for CAPI dependency update

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/7215. See also https://cluster-api.sigs.k8s.io/providers/v1alpha1-to-v1alpha2.html#the-cluster-and-machine-resources.


rossf7 (approved):
LGTM

As this is based on the previous v1alpha1 CR but my knowledge of CAPI is limited.

pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
```diff
@@ -0,0 +1,243 @@
...
+                      type:
+                        enum:
+                          - Creating
+                          - Created
+                          - Updating
+                          - Updated
+                          - Deleting
+                          - Deleted
+                    type: object
+                  type: array
```
MarcelMue: This is hard to read for me. Can we find a convention to always have `type` as the first entry? Right now I have to look all the way through whitespace during review to figure out the type.
It's hard to read for me but might be only me.

MarcelMue (approved):
Looks sane to me.

pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
xh3b4sd: I found that annoying as well. It was the yaml linter I used. Working with the text files is awful.

xh3b4sd: I found that annoying as well. It was the yaml linter I used. Working with the multi line strings is awful.

tuommaki:
Going with this. Current understanding is that this should be complete and functional. In case there are any issues, let's keep communication in close loop and iterate here.

PR #300, opened 2019-11-27
Add support for availability zones for Azure Clusters

axbarsan:
References giantswarm/api#948

Copied the AWS implementation


fiunchinho:
The PR is a good start but I think we are missing some things.

Apart from adding the AZ fields to the _status_ structs to show the current status of the infrastructure, we also need to allow the user to select which AZ's the user wants. That means we also need to change the `AzureConfigSpecAzure` struct with the AZs field.

A couple of questions that we need to answer:
- AZ's are set up differently than in AWS. IIRC, Availability Zones on Azure don't depend on subnets, so I don't think the `AzureConfigStatusAzureAvailabilityZone` object needs a subnet field. Correct?
- On AWS, IIUC, [we pass an integer value](https://github.com/giantswarm/apiextensions/blob/7788d0c28a5e28c04d1df463afeb1ef129abfa3b/pkg/apis/provider/v1alpha1/aws_types.go#L91) specifying how many Availability Zones we want to use, and then [we randomly select that many AZ's from the available availability zones in the region](https://github.com/giantswarm/aws-operator/blob/59edf6485db0992417b61ec27264d51dd5a1f9d3/service/controller/legacy/v30/resource/ipam/create.go#L64). Do we want to do the same for Azure? Or do we want the user to specify exactly which AZ's to use?

fiunchinho:
Btw, I believe this PR is towards https://github.com/giantswarm/giantswarm/issues/7737

pkg/apis/provider/v1alpha1/azure_types.go
```diff
@@ -65,12 +65,12 @@ type AzureConfigSpec struct {
 }
 
 type AzureConfigSpecAzure struct {
-	CredentialSecret CredentialSecret                   `json:"credentialSecret" yaml:"credentialSecret"`
-	DNSZones         AzureConfigSpecAzureDNSZones       `json:"dnsZones" yaml:"dnsZones"`
-	VirtualNetwork   AzureConfigSpecAzureVirtualNetwork `json:"virtualNetwork" yaml:"virtualNetwork"`
-
-	Masters []AzureConfigSpecAzureNode `json:"masters" yaml:"masters"`
-	Workers []AzureConfigSpecAzureNode `json:"workers" yaml:"workers"`
+	AvailabilityZones []int                              `json:"availabilityZones" yaml:"availabilityZones"`
```
fiunchinho: it looks weird but I just added `AvailabilityZones` and reorder the properties.

fiunchinho: On AWS, IIUC, [we pass an integer value](https://github.com/giantswarm/apiextensions/blob/7788d0c28a5e28c04d1df463afeb1ef129abfa3b/pkg/apis/provider/v1alpha1/aws_types.go#L91) specifying how many Availability Zones we want to use, and then [we randomly select that many AZ's from the available availability zones in the region](https://github.com/giantswarm/aws-operator/blob/59edf6485db0992417b61ec27264d51dd5a1f9d3/service/controller/legacy/v30/resource/ipam/create.go#L64). 

I'm not sure if this makes sense for Azure. [According to the docs](https://docs.microsoft.com/en-us/azure/availability-zones/az-overview)
> To ensure resiliency, there’s a minimum of three separate zones in all enabled regions

The [ARM template](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-use-availability-zones#zone-redundant-scale-set-2) receives an array of integers with the numbers of the regions you want to use. So we have two options:
- We could try to keep the same UI for both AWS and Azure and make users pass the number of AZ's and take a random number from the [1, 2, 3] array.
- We allow users to specify which AZ's to use. The UI would be different from AWS.

whites11: I think UX consistency is important. We might not have customers that are azure and AWS customers at the same time, but we might have in the future. Providing them different UIs for the same feature would confuse them I think.
Althought selecting the actual AZs to be used would be a better thing IMHO.

axbarsan:
Thank you so much, @fiunchinho for taking the lead on this 🙏

pkg/apis/provider/v1alpha1/azure_types.go
fiunchinho: I just found out that @marians talked about the same issue [on this comment](https://github.com/giantswarm/giantswarm/issues/5805#issuecomment-555081384) some days ago.

tuommaki: The UX should be consistent, but we allow user to select specific AZs on AWS when using node pool clusters. That way the consistency is slightly loose. It's inconsistent with legacy AWS releases, but more or less consistent with latest releases for both AWS and Azure.

tuommaki: I think `[]int` for AZs looks more weird, but I get why that is. All good here I think.

pkg/apis/provider/v1alpha1/azure_types.go
```diff
@@ -123,6 +123,7 @@ type AzureConfigSpecVersionBundle struct {
 }
 
 type AzureConfigStatus struct {
+	Azure    AzureConfigStatusAzure    `json:"azure" yaml:"azure"`
```
tuommaki: I think this is wrong. `Provider` below is for provider specific status and `Cluster` is provider independent structure. Therefore `AvailabilityZones` field should be in `AzureConfigStatusProvider` above `Ingress`.

tuommaki (changes requested):

pkg/apis/provider/v1alpha1/azure_types.go
fiunchinho: Oh! I see what you mean. Fixed.

tuommaki (approved):
LGTM :+1:

pkg/apis/provider/v1alpha1/azure_types.go
kopiczko: @fiunchinho 

> I'm not sure if this makes sense for Azure.

Why not?

fiunchinho: I thought it didn't make sense since all regions have the same AZ's, but thinking more about it, it doesn't really matter. Do you think we should keep the random selection instead of choosing the specific AZs?

kopiczko: For me it makes sense to do the same thing we do in AWS.

PR #303, opened 2019-12-10
Allow to pass AZ's as a parameter

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/7744

The parameter is a space-separated string containing the numbers of the AZ's to use. I'm using a string instead of an array because this is used for testing, and we will pass the AZ's as env vars when installing the chart, so it's easier to pass strings.

Another option would be to do this when reading the env var here https://github.com/giantswarm/azure-operator/blob/master/integration/env/azure.go
And using a function like
```
func AzureAvailabilityZones() []int {
	azs := strings.Split(strings.TrimSpace(azureAvailabilityZones), " ")
	zones := make([]int, len(azs))

	for i, s := range azs {
		zones[i], _ = strconv.Atoi(s)
	}
	return zones
}
```

WDYT?


whites11 (approved):
Looks ok to me

fiunchinho:
Closed in favor of #304

PR #306, opened 2019-12-16
Add interface to abstract CommonClusterStatus manipulation

tuommaki:
Provider specific CRs have provider independent common cluster status type in
their status structures. This change adds interface to provide type specific
way to manipulate that field in provider CRs.


pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
```diff
@@ -219,7 +219,7 @@ type AWSClusterSpecProviderMaster struct {
 }
 
 type AWSClusterStatus struct {
-	Cluster  CommonClusterStatus      `json:"cluster" yaml:"cluster"`
+	CommonClusterStatus
```
tuommaki: Embed the shared common type and remove `AWSCluster.Status.Cluster.Cluster` repetition.

xh3b4sd (approved):
I think it is good enough to embed the type. No real need for the interface AFAIK.

xh3b4sd (approved):
LGTM

PR #308, opened 2019-12-16
Add types for githubsearchindices.example.giantswarm.io

marians:
This should add types required for CRs/CRDs handled by [github-search-index-operator](https://github.com/giantswarm/github-search-index-operator).


rossf7:
@marians I added sig-operator. Anyone in the SIG that is onboarded can review this IMO.

xh3b4sd:
you also need to add your type to `register.go`. Then you will notice that this does not compile as your types do not implement `runtime.Object`. To fix that you need to run `./scripts/gen.sh`. Also note that there is a new way of defining the CRDs. Check how releases do that, or the new infrastructure types.

marians:
So I did the generation part (sorry for not reading README properly) and now am here:

```
pkg/apis/example/v1alpha1/zz_generated.deepcopy.go:183:19: in.LastCommitTime.DeepCopyInto undefined (type time.Time has no field or method DeepCopyInto)
```

One of the attributes in my CR is defined as time.Time. Should I make it a string instead?

xh3b4sd:
Check the infrastructure types. There is a `DeepCopyTime` type to work around that problem.

MarcelMue:
Also here:
https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/provider/v1alpha1/deep_copy.go

marians:
Thanks! I now duplicated this into the examples/v1alpha1 folder. Let me know if that needs de-duplication.

rossf7:
@marians We agreed to drop this operator and the CRD but the CRD got missed.

https://github.com/giantswarm/giantswarm/issues/6570#issuecomment-539909986

But it does no harm to update it IMO.

xh3b4sd:

PR #309, opened 2019-12-16
Create initial CRD for diagnostic operator

omegas27:
I only created empty CRD as I want to start working on running tcpdump in a TC because I add configurations to tcpdump


pkg/apis/diagnostic/v1alpha1/diagnostic_types.go
```diff
@@ -0,0 +1,86 @@
...
+	// TBD
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type TcpdumpList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	Items           []Tcpdump `json:"items"`
+}
```
MarcelMue: newline at the end please

PR #314, opened 2019-12-17
Add dns network policy

corest:
Towards https://github.com/giantswarm/giantswarm/issues/8150

CR is super simple for now. There are only two fields:
 - list of domains
 - target networkpolicy to update

No ports for now. No global policies, even if I have an idea of implementing GlobalDNSNetworkPolicy, which will use the label selector to update multiple target network policies instead of 1:1 mapping. But starting with simple MVP.

Status holds the last successful update, not sure why would I need that, but in the future, there might be a list of last successful updates for every domain so that we can e.g. implement TTL for domains.


pkg/apis/example/v1alpha1/dns_network_policy_types.go
```diff
@@ -0,0 +1,87 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NewDNSNetworkPolicyCRD returns a new custom resource definition for
+// DNSNetworkPolicy. This might look something like the following.
+//
+//	apiVersion: apiextensions.k8s.io/v1beta1
+//	kind: CustomResourceDefinition
+//	metadata:
+//	  name: dnsnetworkpolicies.example.giantswarm.io
```
pipo02mix: Why do you use `example` instead of `networking` or something more meaningful?

rossf7: We need to be careful after operator week that it isn't confusing which CRDs are in production.

Putting them in the example group makes sense to me. They can be moved if and when we start using the operator in production.

pkg/apis/example/v1alpha1/dns_network_policy_types.go
```diff
@@ -0,0 +1,87 @@
...
+	Spec              DNSNetworkPolicySpec   `json:"spec" yaml:"status"`
+	Status            DNSNetworkPolicyStatus `json:"status" yaml:"status"`
+}
+
+type DNSNetworkPolicySpec struct {
+	// Domains is the list of domain names, which should be resolved
+	// before updating reference network policy.
+	// e.g. ["kubernetes.local", "http://google.com", "https://twitter.com"]
+	Domains []string `json:"domains" yaml:"domains"`
+	// targetNetworkPolicy is an existing network policy in the object namespace,
```
rossf7:
```suggestion
	// TargetNetworkPolicy is an existing network policy in the object namespace,
```

pkg/apis/example/v1alpha1/dns_network_policy_types.go
pipo02mix: good to know, I will do the same for mine

xh3b4sd:

xh3b4sd:
Since there was no feedback and the branch is super stale I close this. When this is still necessary please recover and reopen. Thanks gang.

PR #316, opened 2019-12-18
Make example code more complete

yasn77:
Add missing bits to the example code so if copy/paste, it can actually work with
minimal changes.


xh3b4sd:

PR #318, opened 2019-12-18
improve openapi v3 schema for app CRD

puja108:
towards https://github.com/giantswarm/giantswarm/issues/8169

### DIFF
--- a/pkg/apis/application/v1alpha1/app_types.go
+++ b/pkg/apis/application/v1alpha1/app_types.go
@@ -49,6 +49,7 @@ spec:
                       type: string
                     namespace:
                       type: string
+                  required: ["name", "namespace"]
                 secret:
                   type: object
                   properties:
@@ -56,6 +57,7 @@ spec:
                       type: string
                     namespace:
                       type: string
+                  required: ["name", "namespace"]
             kubeConfig:
               type: object
               properties:
@@ -66,6 +68,7 @@ spec:
                   properties:
                     name:
                       type: string
+                  required: ["name"]
                 secret:
                   type: object
                   properties:
@@ -74,7 +77,7 @@ spec:
                     namespace:
                       type: string
                   required: ["name", "namespace"]
-				requried: ["inCluster"]
+              requried: ["inCluster"]
             userConfig:
               type: object
               properties:
@@ -85,6 +88,7 @@ spec:
                       type: string
                     namespace:
                       type: string
+                  required: ["name", "namespace"]
                 secret:
                   type: object
                   properties:
@@ -92,7 +96,9 @@ spec:
                       type: string
                     namespace:
                       type: string
+                  required: ["name", "namespace"]
           required: ["catalog", "name", "namespace", "version"]
+      required: ["spec"]
 `
 
 var appCRD *apiextensionsv1beta1.CustomResourceDefinition
--- a/pkg/apis/application/v1alpha1/app_types.go
+++ b/pkg/apis/application/v1alpha1/app_types.go
@@ -77,7 +77,6 @@ spec:
                     namespace:
                       type: string
                   required: ["name", "namespace"]
-              required: ["inCluster"]
             userConfig:
               type: object
               properties:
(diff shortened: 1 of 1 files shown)


pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -49,15 +49,13 @@ spec:
                       type: string
                     namespace:
                       type: string
-                  required: ["name", "namespace"]
```
puja108: like in appcatalogs config is not required

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -49,15 +49,13 @@ spec:
...
                 secret:
                   type: object
                   properties:
                     name:
                       type: string
                     namespace:
                       type: string
-                  required: ["name", "namespace"]
```
puja108: see above

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -76,6 +74,7 @@ spec:
                     namespace:
                       type: string
                   required: ["name", "namespace"]
+				requried: ["inCluster"]
```
puja108: IIRC this is required.

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -76,6 +74,7 @@ spec:
                     namespace:
                       type: string
                   required: ["name", "namespace"]
```
puja108: this should be only required if inCluster false, AFAIK openapi v3 schemas has some logic towards such things, but not sure if it is sufficient to express our needs, otherwise might need to move to admission controll validation

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -86,15 +85,13 @@ spec:
                       type: string
                     namespace:
                       type: string
-                  required: ["name", "namespace"]
```
puja108: userconfig should be even more optional than general config

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -76,6 +77,7 @@ spec:
                     namespace:
                       type: string
                   required: ["name", "namespace"]
+              requried: ["inCluster"]
```
puja108: is there conditional requirements, like when incluster false, require context and secret?

puja108: also incluster is required right? I see us setting it to false when we provide kubeconfings and I do rememebr we had to, right @rossf7 ?

rossf7:
```suggestion
              required: ["inCluster"]
```

I'm not sure. Either way the default will be false due to K8s / Go. I think in cluster-operator we set it just to be explicit.

Maybe if it's true some validation that there is no secret would be useful?

puja108: ok, so it is not required?

rossf7: Yes it is not required.

puja108: ok, removing then, I thought we had it required, but as you explained it is just for clarity not technical requirement

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -76,6 +77,7 @@ spec:
                     namespace:
                       type: string
                   required: ["name", "namespace"]
+              required: ["inCluster"]
```
puja108:
```suggestion
```

xh3b4sd:
Can we merge or close this PR?

puja108:
This was handed over to @giantswarm/team-batman-engineers (specifically 
@pipo02mix IIRC, but he now also moved teams).

rossf7:
There have been a lot of changes in structure to this project. So I think this PR will be stale.

@pipo02mix IIRC you worked on the openapi schema as well as docs. Do you know if this is already done?

pipo02mix:
I worked in the documentation part, not in the schema sorry. But I can check if it is our of date

xh3b4sd:
If anyone wants to pick this up feel free to reopen.

PR #319, opened 2020-01-16
migrate from dep to Go modules

omegas27:
Hello team, I created this as a draft as I wanted to make sure we could test it before merging.

Do you know how we could do that? The gen script works as I clone the code generator on every run.

WDYT?


go.mod
```diff
@@ -0,0 +1,20 @@
...
+	github.com/pkg/errors v0.9.1 // indirect
+	golang.org/x/crypto v0.0.0-20191206172530-e9b2fee46413 // indirect
+	golang.org/x/net v0.0.0-20191207000613-e7e4b65ae663 // indirect
+	golang.org/x/oauth2 v0.0.0-20191202225959-858c2ad4c8b6 // indirect
+	golang.org/x/sys v0.0.0-20191206220618-eeba5f6aabab // indirect
+	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
+	gopkg.in/yaml.v2 v2.2.7 // indirect
+	k8s.io/apiextensions-apiserver v0.17.1
+	k8s.io/apimachinery v0.17.1
+	k8s.io/client-go v0.17.1
```
MarcelMue: This will not work currently as the `go` dependencies we have in our operators are 1.16 IIRC.

@sslavic Has worked on this extensively before - maybe he can judge how much effort it would be?

omegas27: I reverted to 1.16.5. @sslavic could you give me you input?
Also @giantswarm/team-ludacris would it be nice to create a released version of api extensions according to which kubernetes version it targets? a 0.16.5 for now and a 0.17.0 after that?

sslavic: Marcel is right, challenge is non-trivial, IMO
- our whole stack (as listed in recent 1.16 update https://github.com/giantswarm/giantswarm/issues/7675) would not only have to be updated to 1.17 but
- our stack would also have to be switched to Go modules in same go, to ensure dependency information is correctly propagated, since dep is not aware of Go modules dependency info and fallbacks to deriving it itself and can get it wrong so one would have to override/duplicate transitive dependency info to keep it under control
- switch to Helm 3 compatible with 1.17 would be coupled to the stack update and switch to Go modules, to get rid of k8s.io/kubernetes dependency Helm 2 has which if not done would complicate Go modules (see "replace" mess in opsctl for example, you don't want to do it in all those projects comprising our stack) and dependency updates. Helm 3 update is complex in itself (//cc @rossf7); to make it easier there were ideas to first get rid of Helm dependencies in lots of projects which get it just for driving e2e tests by localizing that dependency in architect-orb e2e-integration-test only, and have e2e tests driven and assertions use k8s client only
- for k8s libs 1.16 update, update of cluster-api dependency was also required and it was non-trivial one (similar to helm 3) since cluster-api dependency which was 1.16 compatible was recent one with breaking changes which triggered lots of work in operators and libraries
- last but not least, @kopiczko shared in https://github.com/giantswarm/microerror/pull/21#issuecomment-533227457 what he'd like to get done first, before switching the stack to Go modules, and that refactoring doesn't seem to have been completed

For testing apiextensions branch with k8s lib 1.16 updates in say aws-operator which uses dep for dependency management, we'd create aws-operator branch and pin apiextensions dependency and all other aws-operator dependencies to their branch with k8s lib updates, so no merging of anything was required, multiple people could work in parallel on stack updates, and CI could use shared branches with updates. I don't know how is that process supposed to be effectively done with Go modules, some docs I read before were just evaluating very simple cases of single library used by single project, and publishing and testing changes in the two projects/modules all locally only.

So, maybe we should deal with Pawel's error lib refactoring issue and minimizing helm lib usage first, and then coordinate main part with Helm 3 + Go modules + k8s lib update.

sslavic: Switching apiextensions to Go modules, even with keeping 1.16 will also be a problem, for reasons stated above.

kopiczko: Yes. With the previous version it isn't so easy. See how it's done in opsctl https://github.com/giantswarm/opsctl/blob/090fffdcdf4ab2cba8394b42fabff4b873ccd76d/go.mod#L101.

Also we need to make sure `scripts/gen.sh` is still working. And it won't work out of the box w/o vendor directory.

rossf7: Thanks this write up is very helpful. I'm writing a spec for the Helm 3 upgrade today and I'll cover the Go modules part.

I'm also waiting for https://github.com/helm/helm/pull/7378 to be merged (which should be soon). It updates helm to use go modules with K8s `1.17.1` with the recent improvements from upstream. The Helm dependencies are still quite heavy but they no longer use `k8s.io/kubernetes`. :tada:

omegas27: Thank for your feeback :)

omegas27:
@tfussell That is awesome :) Thanks for finding that out :) I think we can change that to a true PR

go.mod
```diff
@@ -0,0 +1,14 @@
...
+
+go 1.13
+
+require (
+	github.com/google/go-cmp v0.3.0
+	k8s.io/apiextensions-apiserver v0.16.6
+	k8s.io/apimachinery v0.16.6
+	k8s.io/client-go v0.16.6
+	k8s.io/code-generator v0.16.6
+	sigs.k8s.io/yaml v1.1.0
```
rossf7: So this is a wrapper around `ghodss/yaml`? Do we want to update https://github.com/giantswarm/fmt/blob/master/go/yaml_marshalling.md ?

go.mod
```diff
@@ -0,0 +1,14 @@
+module github.com/giantswarm/apiextensions
+
+go 1.13
+
+require (
+	github.com/google/go-cmp v0.3.0
+	k8s.io/apiextensions-apiserver v0.16.6
```
rossf7: Should we target `v0.17.1`?

The move to modules is a big change and so is upgrading to K8s 1.17. So I think combining makes sense.

1.17.2 is out already but Helm is still on 1.17.1 but we can probably do an upstream PR.

go.mod
omegas27: IIRC, it is a fork maintained by the Kubernetes community which supports go modules and has more releases than ghodss/yaml

tfussell:
> @tfussell Nice. 🎉❤️

Thanks. I was going to see if I could come up with some flags for `generate-groups.sh` that would allow me to skip the ugly `cp` and `rm` steps at the end (the output is in the `scripts/` directory instead of in-place), but it's a start.

> Should we target `v0.17.1`?

On the one hand it would be nice to do everything in a big push like with 1.16 and Helm 2.16, but on the other it would be nice to minimize the number of changes to keep this self-contained. If we could switch everything to go modules separately from updating K8s and Helm, I sort of feel like we should try to do it. Maybe we should discuss in the WG.

MarcelMue:
I am also in favour of moving any projects we can to go modules already. Question for me would be if importing apiextensions like this already works for dep projects?

omegas27:
We can vendor using `go mod vendor`. 
So maybe we need to see if we can keep the vendor directory for dep to work while having go modules?

kopiczko:
`go mod vendor` won't help. `vendor` directory of dependencies isn't used.

We can keep Gopkg.{toml,lock} for a while.

.gitignore
```diff
@@ -0,0 +1 @@
+vendor/
```
kopiczko: Why? I can see myself running into issues because of having vendor directory I didn't notice because it's ignored.

rossf7:
Another 2 cents from me. The big bang approach sucks and for me is a last resort.

> We can keep Gopkg.{toml,lock} for a while.

If we can do this for dep compatibility and merge stuff sooner with K8s 1.16.6 that is fine with me. 👍

tfussell:
> We can keep Gopkg.{toml,lock} for a while.

This seems like the way to go as long as we don't forget to keep them in sync. Maybe try to switch to go modules iteratively in 1.16 and then get rid of `Gopkg.{toml,lock}` in 1.17?

go.mod
tfussell: `fmt` has been updated to recommend `sigs.k8s.io/yaml`.

go.mod
tfussell: Going to resolve this as the WG has agreed to target 1.16.6 for now.

tfussell:
Tried to resolve a few comments this evening. Unless someone has an idea for a better way to use `code-generator` without moving directories around in `scripts/gen.sh`, I think this is basically ready to go.

tfussell:
Changelog is created. Waiting for a few small changes to some CRDs before creating the 0.1.0 release.

xh3b4sd:
Hi gang, I wanna push this over the finish line. Thanks for the hard work so far already. 🚀

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+
+
```
xh3b4sd:
```suggestion
```

rossf7:
@xh3b4sd Hey, I needed to merge https://github.com/giantswarm/apiextensions/pull/364 because its blocking me in app-operator and chart-operator.

I'm about to release 0.1.2 just waiting on CI.

rossf7:
@xh3b4sd OK done. Thx!

CHANGELOG.md
```diff
@@ -7,18 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [0.1.2] - 2020-03-20
 
 ### Added
 
 - Add kube-proxy configuration to Cluster type in provider.giantswarm.io/v1alpha1.
 
-### Fixed
 
-- Fixed CRD OpenAPISchemas.
```
tfussell: Strange that this is being removed. It's in master https://github.com/giantswarm/apiextensions/blob/master/CHANGELOG.md#fixed

xh3b4sd: Good catch. I had to merge master a couple of times and this got apparently lost. Added back.

PR #321, opened 2020-01-20
adding apiextensions for HA-Masters

anvddriesch:
Here we can work on the CR we are going to use for HA-Masters. 
Towards: https://github.com/giantswarm/giantswarm/issues/8273


anvddriesch:
@xh3b4sd @lachlanh I have sketched a CR to work on :) Let's iterate from here!

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
```diff
@@ -0,0 +1,128 @@
...
+
+// NewG8sControlPlaneCRD returns a new custom resource definition for
+// G8sControlPlane. This might look something like the following.
+//
+//	apiVersion: apiextensions.k8s.io/v1beta1
+//	kind: CustomResourceDefinition
+//	metadata:
+//	  name: g8sControlPlanes.core.giantswarm.io
+//	spec:
+//	  group: core.giantswarm.io
```
xh3b4sd: We should maybe make this part of the infrastructure type. /cc @tuommaki WDYT?

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
```diff
@@ -0,0 +1,128 @@
...
+//	  scope: Namespaced
+//	  version: v1alpha1
+//	  names:
+//	    kind: G8sControlPlane
```
xh3b4sd: So the idea is we use this for all providers, right? Any argument against that? When we have some infrastructure reference this should be fine, which I do not see in the spec below. So we would need to reference our own e.g. AWS specific type for the control plane so that we can also propagate instance type information and the like.

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
```diff
@@ -0,0 +1,128 @@
...
+	// scale subresource and additional integrations for things like kubectl
+	// describe.. The string will be in the same format as the query-param syntax.
+	// More info about label selectors: http://kubernetes.io/docs/user-guide/labels#label-selectors
+	// +optional
+	Selector string `json:"selector,omitempty"`
+
+	// Total number of non-terminated machines targeted by this control plane
+	// (their labels match the selector).
+	// +optional
+	Replicas int32 `json:"replicas,omitempty"`
```
xh3b4sd: So this is the number of masters we would need to define. The `aws-operator` would not reconcile the `G8sControlPlane` type bit the provider specific one, which then would not contain information about the replicas. I think there is the "owner" reference thingy which we would need to make use of. So when the `aws-operator` sees the AWS specific CP CR it can fetch the `G8sControlPlane` and read the replicas.

xh3b4sd (changes requested):
Please also check the infrastructure types and the CRD Schema Validation. We need this here as well once we sorted the spec. Just FYI.

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
tuommaki: You mean the group should be infrastructure? Yes.
```suggestion
//	  group: infrastructure.giantswarm.io
```

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
tuommaki: The current [kubeadm proposal type](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md#new-api-types) has `InfrastructureTemplate` as `corev1.ObjectReference` in it so that might make sense here as well.

xh3b4sd: I think that is wrong and is likely to change. I would like to stick with the ref naming.

anvddriesch:
I changed the group to `infrastructure` here as suggested. I agree it fits. However, I'm not really sure what I should look at in terms of infrastructure types. Currently I do not see something with that name in the apiextensions.

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
anvddriesch: If there is no argument against having the `g8sControlPlane` for all providers I think it would make sense. Otherwise we could use a "awsControlPlane" directly which somehow makes the infrastructure reference redundant.

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
anvddriesch: Am I understanding it correctly that while `g8sControlPlane` CR (reconciled by `cluster-operator`) has the reference to the aws specific infrastructure CR "downwards", the aws specific infrastructure CR (reconciled by `aws-operator` ) can also have an "owner" reference "upwards"?

xh3b4sd: When we look at our architecture it is not really hierarchical. There is no up or down I think. We have a network of processes in which each node (process) of the network interacts with peers. The owner reference is as you said a way to figure out which data object is referencing the data object you are looking at.

xh3b4sd:
Please move the files to the infrastructure specific group directory. It is still in core.

anvddriesch:
As I said above, I did not see something with the name "infrastructure" so I wasn't entirely sure where to move it. In that case I'm going to make a new directory?

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
anvddriesch: Thank you for the explanation, you are right about the architecture. I also think it makes perfect sense to reference in both directions here.

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
tuommaki: >  I would like to stick with the ref naming.

Do you mean you prefer `InfrastructureReference` instead of `InfrastructureTemplate`? I'm fine with both. At current state I'm not 100% sure which one will end up being more appropriate, but since the type is also reference, I'm perfectly fine with `InfrastructureRef` / `InfrastructureReference`.

> If there is no argument against having the g8sControlPlane for all providers I think it would make sense.

I think `G8sControlPlane` would be the correct way to go since there's infrastructure specific object reference also.

tuommaki (commented):
Regarding the infrastructure apigroup for this, Tim meant this directory: https://github.com/giantswarm/apiextensions/tree/master/pkg/apis/infrastructure/v1alpha2

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
xh3b4sd: +1 `InfrastructureRef`.

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
```diff
@@ -0,0 +1,131 @@
...
+				Status: &apiextensionsv1beta1.CustomResourceSubresourceStatus{},
+			},
+		},
+	}
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type G8sControlPlane struct {
```
xh3b4sd: Let's please have godoc for this and a YAML example like in the other types. Then we also want to show the usage of the `giantswarm.io/docs` annotation. See https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md#common-annotations.

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
```diff
@@ -0,0 +1,131 @@
...
+	Spec              G8sControlPlaneSpec   `json:"spec"`
+	Status            G8sControlPlaneStatus `json:"status"`
+}
+
+type G8sControlPlaneSpec struct {
+	// Replicas is the number replicas of the master node.
+	Replicas int `json:"replicas" yaml:"replicas"`
+	// InfrastructureReference is a required reference to provider-specific
+	// Infrastructure.
+	InfrastructureReference string `json:"infrastructureReference"`
```
xh3b4sd:
```suggestion
	InfrastructureReference string `json:"infrastructureRef"`
```

pkg/apis/core/v1alpha1/g8s_control_plane_types.go
```diff
@@ -0,0 +1,131 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              G8sControlPlaneSpec   `json:"spec"`
+	Status            G8sControlPlaneStatus `json:"status"`
+}
+
+type G8sControlPlaneSpec struct {
+	// Replicas is the number replicas of the master node.
+	Replicas int `json:"replicas" yaml:"replicas"`
+	// InfrastructureReference is a required reference to provider-specific
```
xh3b4sd:
```suggestion
	// InfrastructureRef is a required reference to provider-specific
```

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,133 @@
...
+	Spec              G8sControlPlaneSpec   `json:"spec"`
+	Status            G8sControlPlaneStatus `json:"status"`
+}
+
+type G8sControlPlaneSpec struct {
+	// Replicas is the number replicas of the master node.
+	Replicas int `json:"replicas" yaml:"replicas"`
+	// InfrastructureRef is a required reference to provider-specific
+	// Infrastructure.
+	InfrastructureReference string `json:"infrastructureReference"`
```
tuommaki:
```suggestion
	InfrastructureReference corev1.ObjectReference `json:"infrastructureReference"`
```
See https://godoc.org/k8s.io/kubernetes/pkg/apis/core#ObjectReference

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// KubeadmControlPlaneStatus defines the observed state of G8sControlPlane.
```
tuommaki:
```suggestion
// G8sControlPlaneStatus defines the observed state of G8sControlPlane.
```

xh3b4sd:
It looks like the version of the code-generator changed and with it some behaviour so some expected code is not generated. We need a better understanding of this and see which version of the code-generator we need to use in order to make that thing here fly.

anvddriesch:
I looked a little bit into the code generator problem but don't quite understand it yet. : (

anvddriesch:
@xh3b4sd Lach and I worked on the prototypes of `g8sControlPlane` and `awsControlPlane`
Take a look when you can. :)

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -0,0 +1,99 @@
...
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            availabilityZone:
+              type: string
```
xh3b4sd: We want to run masters in different AZs.

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -0,0 +1,99 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// AWSControlPlane is the infrastructure provider referenced in ControlPlane CRs.
+//
+//     apiVersion: infrastructure.giantswarm.io/v1alpha2
+//     kind: AWSControlPlane
+//	   metadata:
+//       labels:
+//         label1: ????
+//	     name: ????
```
xh3b4sd: That needs more love. Check the other godocs. We want version labels, docs annotations (probably not in the other godocs), and the name is not a label.

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -0,0 +1,99 @@
...
+  group: infrastructure.giantswarm.io
+  names:
+    kind: AWSControlPlane
+    plural: awscontrolplanes
+    singular: awscontrolplane
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
```
xh3b4sd: We need to start to version this. Check https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,102 @@
...
+const g8sControlPlaneCRDYAML = `
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: g8sControlPlanes.core.giantswarm.io
+spec:
+  group: infrastructure.giantswarm.io
+  scope: Namespaced
+  version: v1alpha1
+  names:
```
xh3b4sd: This needs versioned schema validations as well.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,102 @@
...
+//	metadata:
+//    labels:
+//      label1: ????
+//	  name: ????
+//	spec:
+//    replicas: 3
+//    //TODO LH infrastructureRef ?
+//  status:
+//    replicas: 3
+//    readyReplicas: 3
```
xh3b4sd: What is the status difference between replicas and ready replicas?

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,102 @@
...
+//
+//	apiVersion: infrastructure.giantswarm.io/v1alpha2
+//	kind: G8sControlPlane
+//	metadata:
+//    labels:
+//      label1: ????
+//	  name: ????
+//	spec:
+//    replicas: 3
+//    //TODO LH infrastructureRef ?
```
xh3b4sd: Yes.

xh3b4sd (changes requested):
Please put a bit more love into the godoc. The YAMLs are a bit sloppy there.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
xh3b4sd: Ack, so far this distinction was never important and both of these where set equal.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
lachlanh: have added the schema validation, please check.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,134 @@
...
+
+const g8sControlPlaneCRDYAML = `
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: g8sControlPlanes.core.giantswarm.io
+spec:
+  group: infrastructure.giantswarm.io
+  scope: Namespaced
+  version: v1alpha1
```
lachlanh: is this the right version ? awscontrolplane looks like it is at v1alpha2 ? @xh3b4sd @anvddriesch

xh3b4sd: My thinking was to keep it aligned with the package path version. So it would be `v1alpha2`. But then when we introduce `v1alpha3` we would create a new package. I would assume we would not want to create a new CRD then, but use the one which is versioned, so we can make the conversions. @giantswarm/sig-operator any ideas?

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,134 @@
...
+            infrastructureRef:
+              properties:
+                kind:
+                  type: string
+                namespace:
+                  type: string
+                name:
+                  type: string
+                uid:
+                  type: string
```
xh3b4sd: AFAIK `uid` is set automatically. We should drop it here.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,134 @@
...
+                apiVersion:
+                  type: string
+                resourceVersion:
+                  type: string
+                fieldPath:
```
xh3b4sd: `fieldpath` is not used in our case AFAIK. Let's drop.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,134 @@
...
+                  type: string
+                namespace:
+                  type: string
+                name:
+                  type: string
+                uid:
+                  type: string
+                apiVersion:
+                  type: string
+                resourceVersion:
```
xh3b4sd: We may not want to specify this. Shouldn't be validated IMO.

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
anvddriesch: good point! I changed that to an array- hopefully the syntax is correct.

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
anvddriesch: I updated it. I am not sure how the docs example should look like, though. What would it be?

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
anvddriesch: I tried to add the versioning as I understood it from the doc. I dropped the "validation" part since it seemed redundant. Is that correct?

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -0,0 +1,111 @@
...
+	storage: true
+	schema :
+      openAPIV3Schema:
+        properties:
+          spec:
+            properties:
+              availabilityZones:
+			    type: array
+				  items:
+				    type: string
```
xh3b4sd: Check the multi line strings. I always have the same problem when working with these so I am not sure why we decided to do it that way at all. @giantswarm/sig-operator why though?

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -0,0 +1,111 @@
...
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// AWSControlPlane is the infrastructure provider referenced in ControlPlane CRs.
+//
+//     apiVersion: infrastructure.giantswarm.io/v1alpha2
+//     kind: AWSControlPlane
+//	   metadata:
+//       annotations:
+//		   giantswarm.io/docs:
```
xh3b4sd: Indentation.

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -0,0 +1,111 @@
...
+//       labels:
+//         aws-operator.giantswarm.io/version: 6.2.0
+//         giantswarm.io/cluster: "8y5kc"
+//         giantswarm.io/organization: "giantswarm"
+//         release.giantswarm.io/version: 7.3.1
+//       name: 8y5kc
+//     spec:
+//       availabilityZones: eu-central-1a
```
xh3b4sd: That should be a list. Or do we want to do some comma separated string magic?

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,134 @@
...
+//      giantswarm.io/organization: "giantswarm"
+//      release.giantswarm.io/version: 7.3.1
+//    name: 8y5kc
+//	spec:
+//    replicas: 3
+//    infrastructureRef:
+//      kind: AWSControlPlane
+//      namespace: infrastructure.giantswarm.io
+//      name: awscontrolplanes.infrastructure.giantswarm.io
+//      apiVersion: v1apha2
```
lachlanh: @xh3b4sd does this look like what we would expect here ?

xh3b4sd: You can check the Node Pool clusters on the installations to see if this makes sense.

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
xh3b4sd: This is still not 100%.

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -0,0 +1,117 @@
...
+	Spec              AWSControlPlaneSpec   `json:"spec" yaml:"spec"`
+	Status            AWSControlPlaneStatus `json:"status,omitempty" yaml:"status,omitempty"`
+}
+
+type AWSControlPlaneSpec struct {
+	AvailabilityZones []string `json:"availabilityZones" yaml:"availabilityZones"`
+	InstanceType      string   `json:"instanceType" yaml:"instanceType"`
+}
+
+// TODO
```
xh3b4sd: Just for the record we iterate here as we are not sure yet what we will need. In worst case we just clean up later.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,134 @@
...
+
+const (
+	kindG8sControlPlane = "G8sControlPlane"
+)
+
+const g8sControlPlaneCRDYAML = `
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: g8sControlPlanes.core.giantswarm.io
```
xh3b4sd:
```suggestion
  name: g8scontrolplanes.infrastructure.giantswarm.io
```

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,134 @@
...
+spec:
+  group: infrastructure.giantswarm.io
+  scope: Namespaced
+  names:
+    kind: G8sControlPlane
+    plural: g8sControlPlanes
```
xh3b4sd:
```suggestion
    plural: g8scontrolplanes
```

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,134 @@
...
+    singular: g8sControlPlane
```
xh3b4sd:
```suggestion
    singular: g8scontrolplane
```

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -0,0 +1,134 @@
...
+
+// G8sControlPlane defines the ControlPlane (Master nodes) of a
+// Giant Swarm Tenant Cluster
+//
+//	apiVersion: infrastructure.giantswarm.io/v1alpha2
+//	kind: G8sControlPlane
+//	metadata:
+//    labels:
+//      aws-operator.giantswarm.io/version: 6.2.0
+//      cluster-operator.giantswarm.io/version:0.17.0
```
xh3b4sd:
```suggestion
//      cluster-operator.giantswarm.io/version: 0.17.0
```

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// AWSControlPlane is the infrastructure provider referenced in ControlPlane CRs.
+//
+//     apiVersion: infrastructure.giantswarm.io/v1alpha2
+//     kind: AWSControlPlane
+//	   metadata:
```
xh3b4sd:
```suggestion
//	 metadata:
```

xh3b4sd: We also miss the owner reference.

PR #323, opened 2020-01-30
SSH users is a list, instead of hardcoding a single user

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/7982


helm/apiextensions-azure-config-e2e-chart/values.yaml
```diff
@@ -1,8 +1,9 @@
 clusterName: "test-cluster"
 commonDomain: "test-host-cluster.westeurope.azure.gigantic.io"
 commonDomainResourceGroup: "test-host-cluster"
-sshUser: "test-user"
-sshPublicKey: "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAAYQCurvzg5Ia54kb3NZapA6yP00//+Jt6XJNeC7Seq3TeCqMR9x7Snalj19r0lWok1PkRgDo1PXj+3y53zo/wqBrPqN4cQqp00R06kNfnhAgesaRMvYhuyVRQQbfXV5gQg8M= dummy-key"
+sshUserList:
+- name: jose
```
fiunchinho: this is just a default value, but in reality, the list of our keys from the installations repository will be passed

PR #324, opened 2020-01-31
Change IgnitionStatusConfigMap to IgnitionStatusSecret

stone-z:
Toward https://github.com/giantswarm/giantswarm/issues/8643 and https://github.com/giantswarm/giantswarm/issues/8644


pkg/apis/core/v1alpha1/ignition_types.go
```diff
@@ -201,10 +201,10 @@ type IgnitionSpecSSO struct {
 
 // IgnitionStatus holds the rendering result.
 type IgnitionStatus struct {
-	ConfigMap IgnitionStatusConfigMap `json:"template" yaml:"template"`
+	Secret IgnitionStatusSecret `json:"template" yaml:"template"`
```
MarcelMue: Can you change this to match the status in the spec:
https://github.com/giantswarm/giantswarm/blob/5a871c1971b75043d8545b6f3fcc7456dc14ba67/specs/ignition_operator.md#ignition-cr-design

stone-z: This should match now

PR #325, opened 2020-02-08
Use k8s.io/api instead of k8s.io/kubernetes

tfussell:
In a previous PR, the core Kubernetes core API group was imported via k8s.io/kubernetes. I think it's preferable to use k8s.io/api which comes from k8s.io/kubernetes.


pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -2,9 +2,9 @@ package v1alpha2
 
 import (
 	"github.com/ghodss/yaml"
+	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	corev1 "k8s.io/kubernetes/pkg/apis/core"
```
MarcelMue: The change looks very straight forward, could this have some unforeseen consequences?
I am a bit weary because I don't know how kubernetes reacts if the ref type essentially changes (even if it's just a different api group?) during upgrades.

tfussell: It was only added a few days ago in https://github.com/giantswarm/apiextensions/pull/321. Still, it might be good to make sure it's not being used by production operators yet. As far as I can tell, it's exactly the same code so I don't believe it will be treated any differently by Kubernetes.

MarcelMue: Ack - I agree. Might still be good to wait for one approval from @giantswarm/team-firecracker

xh3b4sd: We do not use it yet. Still fiddling with the PoC. There is no CRD established on any CP.

xh3b4sd (approved):

rossf7 (approved):
LGTM

PR #328, opened 2020-02-12
Implement Release CR from spec

tfussell:
Implements https://github.com/giantswarm/giantswarm/pull/8866


pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -1,125 +1,98 @@
...
-	kindRelease = "Release"
-
-	ChangelogKindAdded      ReleaseChangelogKind = "added"
-	ChangelogKindChanged    ReleaseChangelogKind = "changed"
-	ChangelogKindDeprecated ReleaseChangelogKind = "deprecated"
-	ChangelogKindFixed      ReleaseChangelogKind = "fixed"
-	ChangelogKindRemoved    ReleaseChangelogKind = "removed"
-	ChangelogKindSecurity   ReleaseChangelogKind = "security"
+	kindRelease   = "Release"
+	semverPattern = `^(=|>=|<=|=>|=<|>|<|!=|~|~>|\^)?(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
```
tfussell: The cool thing about the approach of defining CRDs using go instead of YAML is that we can easily define rules for common properties and use them in several places without repeating.

pkg/apis/release/v1alpha1/release_types_test.go
```diff
@@ -1,13 +1,130 @@
...
 		t.Error("Release CRD was nil.")
+		return
 	}
 	if crd.Name == "" {
 		t.Error("Release CRD name was empty.")
 	}
 }
+
+func sortErrors(errors []*errors.Validation) {
+	sort.Slice(errors, func(i, j int) bool { return errors[i].Name < errors[j].Name })
```
tfussell: Maps in go are defined to have random order so we need to sort errors by name before comparing. This is an in-place sort.

pkg/apis/release/v1alpha1/release_types_test.go
```diff
@@ -1,13 +1,130 @@
...
+							Name:    "kubernetes",
+							Version: "1.18.0",
+						},
+					},
+					Status:  StatusActive,
+					Version: "13.1.2",
+				},
+			},
+			errors: nil,
+		},
```
tfussell: Some other cases might be: non-semver formatted versions in various places, a status that isn't "active", "deprecated", or "wip", a release with 0 components, and missing various properties.

pkg/apis/release/v1alpha1/release_types_test.go
```diff
@@ -1,13 +1,130 @@
...
+	}
+	crd := NewReleaseCRD()
+
+	var v apiextensions.CustomResourceValidation
+	err := v1beta1.Convert_v1beta1_CustomResourceValidation_To_apiextensions_CustomResourceValidation(crd.Spec.Versions[0].Schema, &v, nil)
```
tfussell: Taken from `apiextensions-apieserver`. It's ugly, but it's public so I think it's okay to use.

pkg/apis/release/v1alpha1/release_types_test.go
```diff
@@ -1,13 +1,130 @@
...
+		t.Fatal(err)
+	}
+
+	validator, _, err := validation.NewSchemaValidator(&v)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	opts := []cmp.Option{
+		cmpopts.IgnoreUnexported(errors.Validation{}),
```
tfussell: `cmp` tries to compare unexported fields so we need to specifically skip them here.

pkg/key/key.go
```diff
@@ -0,0 +1,64 @@
...
+func NewCRD(kind, group, version string, scope v1beta1.ResourceScope, schema v1beta1.JSONSchemaProps) *v1beta1.CustomResourceDefinition {
+	preserve := false
+	return &v1beta1.CustomResourceDefinition{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: v1beta1.SchemeGroupVersion.String(),
+			Kind:       "CustomResourceDefinition",
+		},
+		ObjectMeta: CRDObjectMeta(kind, group),
+		Spec: v1beta1.CustomResourceDefinitionSpec{
+			Group:                 group,
```
tfussell: This uses the new style of CRD spec which can contain multiple versions.

pkg/key/key.go
```diff
@@ -0,0 +1,64 @@
...
+
+// CRDObjectMeta returns a Kubernetes kind to a CRD plural form.
+// For example, AWSCluster would become awsclusters.
+func CRDObjectMeta(kind, group string) metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Name: fmt.Sprintf("%s.%s", kind, group),
+	}
+}
+
+func NewCRD(kind, group, version string, scope v1beta1.ResourceScope, schema v1beta1.JSONSchemaProps) *v1beta1.CustomResourceDefinition {
```
tfussell: Still using `v1beta1.CustomResourceDefinition`. Is it time to switch to `v1.CustomResourceDefinition`?

pkg/apis/release/v1alpha1/release_types.go
stone-z: "easily" :stuck_out_tongue_winking_eye:

stone-z: I think you may also need to anticipate a preceding `v` in the version pattern?

tfussell: Hey, beats copying and pasting this into multiple places in YAML. Good point about the `v`. We'll need to decide if it should always be present or never be present on versions. Also I'll remove these comparisons at the beginning since it should match exact versions only.

pkg/apis/release/v1alpha1/release_types_test.go
```diff
@@ -1,13 +1,130 @@
...
-import "testing"
+import (
+	"github.com/go-openapi/errors"
+	_ "github.com/go-openapi/validate"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apiextensions-apiserver/pkg/apiserver/validation"
+	"sort"
```
tomahawk28: imports need to be sorted in here.

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -138,79 +111,49 @@ func NewReleaseTypeMeta() metav1.TypeMeta {
...
+//	  name: "v6.1.0"
 //	spec:
-//	  changelog:
-//	    - component: "cloudconfig"
-//	      description: "Replace cloudinit with ignition."
-//	      kind: "changed"
+//    apps:
+//	    - name: "net-exporter"
+//	      version: "1.0.0"
+//        componentVersion: "0.2.0"
```
tomahawk28:
```suggestion
//          componentVersion: "0.2.0"
```
onre more indent.

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -138,79 +111,49 @@ func NewReleaseTypeMeta() metav1.TypeMeta {
...
-//	  parentVersion: "6.0.1"
-//	  version: "6.1.0"
-//	status:
-//	  cycle:
-//	    phase: "eol"
-//	    enabledDate: 2019-01-08
-//	    disabledDate: 2019-01-12
+//	    - name: "kubernetes"
+//	      version: "1.18.0-alpha.3"
+//    status: active
```
tomahawk28: `status` felt it like its reporting status from objects, just like app CRs, chart CRs. 
Can we have `spec.status` as a boolean type? I think it makes sense.

tfussell: This isn't about the status of the resource, but the release itself like when we change a release from `active: true` to `active: false` to disable old releases. This is too similar to the normal concept of CR status so I've renamed it to `state` now.

pkg/apis/release/v1alpha1/release_types.go
tfussell: Decided to enforce no `v` in versions since we control this and want to be consistent. We can change this later if desired.

pkg/apis/release/v1alpha1/release_types_test.go
tfussell: I added tests to cover all of these cases.

pkg/apis/release/v1alpha1/release_types_test.go
tfussell: Thanks. Fixed.

pkg/apis/release/v1alpha1/release_types.go
tfussell: Fixed.

pkg/apis/release/v1alpha1/validation.go
```diff
@@ -0,0 +1,81 @@
+package v1alpha1
+
+import (
+	"fmt"
+	"github.com/giantswarm/to"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"strings"
```
tomahawk28: Sorry, again imports :)

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -4,84 +4,23 @@ import (
...
 )
 
 const (
-	kindRelease   = "Release"
-	semverPattern = `^(=|>=|<=|=>|=<|>|<|!=|~|~>|\^)?(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
+	kindRelease = "Release"
 )
 
-type ReleaseStatus string
+type ReleaseState string
```
tomahawk28: I'm fine with `State` but you might need to get agreement from your team.

tomahawk28 (approved):

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -135,82 +47,52 @@ func NewReleaseTypeMeta() metav1.TypeMeta {
...
+//  kind: Release
+//  metadata:
+//    name: 13.0.0
+//  spec:
+//    version: 13.0.0
+//    state: active
+//    apps:
+//      - name: net-exporter
+//        version: 1.0.0
+//        componentVersion: 0.2.0
```
kopiczko: Do we need this componentVersion?

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -135,82 +47,52 @@ func NewReleaseTypeMeta() metav1.TypeMeta {
...
+//    components:
+//      - name: kubernetes
+//        version: 1.18.0-alpha.3
```
kopiczko: I don't see authorities here.

pkg/apis/release/v1alpha1/release_types.go
tfussell: Releases are going to become the single source of truth for app versions including upstream component versions. Since the authorities will no longer expose owned component versions, this will allow us to still support showing `kube-state-metrics: 1.9.0` in `gsctl show release <verson>` for example. As for whether that's important, I'm not sure. I can discuss with the team.

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -1,125 +1,37 @@
...
-                  format: date
-                phase:
-                  enum:
-                  - upcoming
-                  - enabled
-                  - disabled
-                  - eol
-              required:
-              - phase
-`
```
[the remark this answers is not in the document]
tfussell: I had two main reasons. The first is that it allows us to use functions to create CRDs programatically. I wanted to use the long semver regex and it's much nicer to define it once and reuse it for each property that needs that validation. The second reason is that I didn't know there had been agreement about the definition. I looked through a few CRDs and saw some in each format.

pkg/apis/release/v1alpha1/release_types.go
tfussell: I thought it would be nicer to combine authorities into components so things like `aws-operator` will be in this section too. Since they are only authorities for themselves now, the fact that they expose version information is less important.

kopiczko: That makes sense 👍 Is that included in the spec?

pkg/apis/release/v1alpha1/release_types.go
kopiczko: I see 👍

pkg/apis/release/v1alpha1/release_types.go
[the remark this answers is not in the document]
kopiczko: It's possible we had this agreement before you joined the company. We didn't move all the CRDs actively though. I see what you mean with this semver regexp. Is that big of an issue to copy it in multiple placed though?

pkg/apis/release/v1alpha1/validation.go
tfussell: Good catch.

pkg/apis/release/v1alpha1/release_types.go
tfussell: No. I'll look for some consensus and make sure this PR matches the spec PR before merging both.

pkg/apis/release/v1alpha1/release_types.go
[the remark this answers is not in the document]
tfussell: It's okay to copy and paste it here I guess but not that pretty. I thought this could be a launching point for adding some validation to all of the CRDs in which case we'd want to reuse this programmatic CRD generation method. Something to discuss with SIG Operator I guess.

pkg/key/key.go
```diff
@@ -0,0 +1,64 @@
+package key
+
+import (
+	"fmt"
+	v1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strings"
```
xh3b4sd: This should not pass CI. Can we check to get the latest architect orb into `apiextensions`?

tfussell: I fixed this manually. I will create a separate PR for using architect-orb.

pkg/apis/release/v1alpha1/release_types.go
MarcelMue: This is included in the spec now.

pkg/apis/release/v1alpha1/release_types.go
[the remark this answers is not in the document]
MarcelMue: My impression was that we only had the text representation because the code one seemed hard to implement and it was easier to do. 

I like the unit test coverage and so on - so a favour the code approach.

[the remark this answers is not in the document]
kopiczko: What is not / cannot be tested with the text-based approach?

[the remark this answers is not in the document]
tfussell: Aside from requiring copying and pasting of validation, test-based CRDs are potentially more error prone when it comes to field names. For example, if we have a typo in `openAPIV3Schema` such as `openAPIV3Scema`, `yaml.Unmarshal` would silently return a `CustomResourceDefinition` with an empty schema. Without validation tests, we wouldn't easily catch this.

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -135,82 +47,52 @@ func NewReleaseTypeMeta() metav1.TypeMeta {
...
-//	  version: "6.1.0"
-//	status:
-//	  cycle:
-//	    phase: "eol"
-//	    enabledDate: 2019-01-08
-//	    disabledDate: 2019-01-12
+//  apiVersion: release.giantswarm.io/v1alpha1
+//  kind: Release
+//  metadata:
+//    name: 13.0.0
```
marians: Can we make it a convention to prefix the name with a `v`? So make it `v13.0.0` ?

I would like to use that convention in most cases, except where we have an attribute that should only contain a semver version number, like in the `version` attribute unter `apps`.

pkg/apis/release/v1alpha1/release_types.go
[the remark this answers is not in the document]
marians: +1 for YAML CRDs. That's one step closer to making CR(D)s documentable.

@tfussell Have you considered using `yaml.UnmarshalStrict`?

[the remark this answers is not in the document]
kopiczko: I'm still not convinced. That can be easily tested.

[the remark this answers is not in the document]
tfussell: > I'm still not convinced. That can be easily tested.

It is testable but not currently being tested. I see what you mean though. I thought the functions for generating a CRD would let us reduce errors and inconsistencies across the codebase, but it's not that important to me. I'll revert to a string-based definition, and we can bring these functions back in the future if desired.

> @tfussell Have you considered using `yaml.UnmarshalStrict`?

I just found out about `UnmarshalStrict` myself--that may address some of the concerns.

pkg/apis/release/v1alpha1/release_types.go
tfussell: I was on the fence about whether it should have a `v` or not here. Thinking back, we had a little discussion in #sig-ux about the `v` and I guess the vague consensus was that it's a little clearer to add it where it's shown publicly. We also recently discussed whether release names should have any meaning or are only for display purposes and we decided on the latter. That being the case, I'll add it to the CR name but not to the versions elsewhere in the CR.

pkg/key/key.go
```diff
@@ -0,0 +1,65 @@
...
+			Names:                 KindNames(kind),
+			Versions: []v1beta1.CustomResourceDefinitionVersion{
+				{
+					Name:    version,
+					Served:  true,
+					Storage: true,
+					Schema: &v1beta1.CustomResourceValidation{
+						OpenAPIV3Schema: &schema,
+					},
+				},
```
xh3b4sd: Last week I just learned that defining a single version is invalid. See https://github.com/giantswarm/cluster-operator/pull/928. The validation of the schema versions within Kubernetes is as such that versions must not be equal while this very validation does not care if there is only one version given. I was hesitant to suggest a fix to upstream. In any case we need to define a second dummy version to work around the validation issues.

xh3b4sd:
Can I help somehow getting this sorted? There are a couple of interesting advances that I also would like to make use of for the HA Masters story. It looks like nothing happened here for some time now.

tfussell:
@xh3b4sd I'm working on this now.

pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -1,123 +1,111 @@
...
-	ChangelogKindSecurity   ReleaseChangelogKind = "security"
-)
-
-const releaseCRDYAML = `
-apiVersion: apiextensions.k8s.io/v1beta1
+	releaseCRDYAML = `apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: releases.release.giantswarm.io
+  creationTimestamp: null
```
MarcelMue: Why is this needed? Sorry if I miss something obvious.

pkg/apis/release/v1alpha1/release_types.go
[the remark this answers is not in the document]
tfussell: Similarly to the ignition PR, I've reverted to a text-based CRD definition and added "golden"-type tests which generate actual YAML from the CRD and CR for documentation. See https://github.com/giantswarm/apiextensions/blob/8bd62cacf86eb27ebd0fe4eea76e1d2c6f933f52/docs/cr/release.giantswarm.io_v1alpha1_release.yaml for an example. I've also switched to `UnmarshalStrict` so I think all of these issues should be resolved.

pkg/apis/release/v1alpha1/release_types.go
tfussell: This is done. There's now a regex which ensures that every CR has a name that looks like `v1.2.3` (`something-non-semver`, `1.2.3` and `v1.2.3-dev` are not allowed).

pkg/key/key.go
tfussell: Thanks for letting me know about this. I don't really like the idea of having a dummy version. I'm going to investigate other options including using the old style definition with a single top-level version or using updated Kubernetes with a fix for this.

MarcelMue (approved):

tfussell:
Here's what the list of releases looks like with the new printer columns: 
```
kubectl get releases
NAME      KUBERNETES VERSION   STATE        AGE
v12.0.0   1.18.0               deprecated   115m
```

tfussell:
And the type is mostly documented now:
```
kubectl explain releases.spec
KIND:     Release
VERSION:  release.giantswarm.io/v1alpha1

RESOURCE: spec <Object>

DESCRIPTION:
     Spec holds the data defining the desired state of a release.

FIELDS:
   apps	<[]Object> -required-
     Apps is a list of Giant Swarm-managed apps which will be installed by
     default on clusters created with this release version.

   components	<[]Object> -required-
     Components is a list of internal and upstream components making up the core
     of the cluster.

   state	<string> -required-
     State indicates how this release should be used. "wip" means the release is
     a work in progress. "deprecated" means old clusters using this version will
     continue to function but new clusters should use a more recent release.
     "active" means this is a current supported release.

   version	<string> -required-
     Semantic version of the release.
```

tfussell:
And here's the example CR from the docs directory:
```
apiVersion: release.giantswarm.io/v1alpha1
kind: Release
metadata:
  annotations:
    giantswarm.io/docs: https://docs.giantswarm.io/reference/releases.release.giantswarm.io/v1alpha1/
spec:
  apps: []
  components:
  - name: kubernetes
    version: 1.18.0
  state: wip
  version: 12.0.0
```

docs/cr/release.giantswarm.io_v1alpha1_release.yaml
```diff
@@ -0,0 +1,14 @@
...
+    giantswarm.io/docs: https://docs.giantswarm.io/reference/releases.release.giantswarm.io/v1alpha1/
+  creationTimestamp: null
+  name: v12.0.0
+spec:
+  apps: []
+  components:
+  - name: kubernetes
+    version: 1.18.0
+  state: wip
+  version: 12.0.0
```
xh3b4sd: Is this the example now? The godoc on the `Release` struct is gone which I do not like. I want to see the godoc and have an example. When this here is the example then can we have a link to this on the `Release` struct godoc?

xh3b4sd (approved):

docs/crd/release.giantswarm.io_release.yaml
```diff
@@ -0,0 +1,128 @@
...
+    openAPIV3Schema:
+      description: |
+        A Release holds information about a particular version of the Giant Swarm platform which
+        can be used as a target for creation or upgrade of a cluster. It is a tested package
+        comprising a particular Kubernetes version along with compatible Giant Swarm operators,
+        monitoring, and default apps.
+      properties:
+        metadata:
+          properties:
+            name:
```
kopiczko: Can you restrict the name with schema validation? TIL.

kopiczko (approved):

docs/crd/release.giantswarm.io_release.yaml
tfussell: I just found out as well. Only `name` and `generateName` can be validated in `metadata`: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

docs/cr/release.giantswarm.io_v1alpha1_release.yaml
tfussell: What's your use case for wanting an example in the godoc? I guess I'm wondering if the godoc should link to https://docs.giantswarm.io/reference/releases.release.giantswarm.io/v1alpha1/ (currently doesn't exist, but I think this is being worked on in https://github.com/giantswarm/giantswarm/issues/8804) or https://github.com/giantswarm/apiextensions/blob/35b9818f2dbe2396b1b86e079bb1fad004e53241/docs/cr/release.giantswarm.io_v1alpha1_release.yaml (https://github.com/giantswarm/apiextensions/blob/master/docs/cr/release.giantswarm.io_v1alpha1_release.yaml once merged).

pkg/apis/release/v1alpha1/release_types.go
tfussell: It has `omitempty`, but it's not a pointer so it defaults to `null` during marshalling. I just missed it during some copying and pasting rounds. Good catch.

docs/cr/release.giantswarm.io_v1alpha1_release.yaml
xh3b4sd: At first it was a UX requirement to see on first side how the whole thing looks like. Then it helps with development. During Node Pools for instance we used the examples to create CRs manually and that was useful.

pkg/key/key.go
tfussell: Discussed a bit more here https://gigantic.slack.com/archives/C2MS2VB37/p1582832328036400 and decided to follow Jordan's suggestion of sticking to top level definitions as long as there's only one version and we're on CRD v1beta1.

docs/cr/release.giantswarm.io_v1alpha1_release.yaml
tfussell: I added a link to the example CR in the docs directory.

PR #329, opened 2020-02-15
Improve Ignition CRD

tfussell:
This adds godoc documentation to all Ignition CRD fields (for https://github.com/giantswarm/giantswarm/issues/8837), adds an object in the status for ignition verification (for https://github.com/giantswarm/ignition-operator/pull/32), adds OIDC fields (for https://github.com/giantswarm/giantswarm/issues/8640), adds a `CloudProvider` field which should be passed to the API server, and updates the CRD to the non-deprecated form.

I did it in one go because these are multi-part stories and the required CRD refactoring all felt related.


pkg/apis/core/v1alpha1/ignition_types.go
```diff
@@ -11,24 +11,8 @@ const (
...
-//       names:
-//         kind: Ignition
-//         plural: ignitions
-//         singular: ignition
-//       subresources:
-//         status: {}
-//
+// NewIgnitionCRD returns a new custom resource definition for an ignition.
+// Ignitions contain a rendered ignition template specific to nodes in a particular cluster.
 func NewIgnitionCRD() *apiextensionsv1beta1.CustomResourceDefinition {
```
omegas27: do we want ignition on old clusters before 1.16? Since CRD are GA now, should we not bump api versions?

pkg/apis/core/v1alpha1/ignition_types.go
```diff
@@ -11,24 +11,8 @@ const (
...
-//       group: core.giantswarm.io
-//       scope: Namespaced
-//       version: v1alpha1
-//       names:
-//         kind: Ignition
-//         plural: ignitions
-//         singular: ignition
-//       subresources:
-//         status: {}
-//
```
xh3b4sd: IMO we want this example in the godoc. For me it is super useful. I would also like to have it on the CR. And let's please always get @giantswarm/sig-ux in the CR specs as these define the interface for our users soon™.

pkg/apis/core/v1alpha1/ignition_types.go
```diff
@@ -74,145 +73,250 @@ type Ignition struct {
...
+	SSO IgnitionSpecSSO `json:"sso" yaml:"sso"`
 }
 
 type IgnitionSpecCalico struct {
-	CIDR    string `json:"cidr" yaml:"cidr"`
-	Disable bool   `json:"disable" yaml:"disable"`
-	MTU     string `json:"mtu" yaml:"mtu"`
-	Subnet  string `json:"subnet" yaml:"subnet"`
+	// CIDR is the CIDR-component of the overlay network. Combined with Subnet below.
+	CIDR string `json:"cidr" yaml:"cidr"`
```
omegas27: Maybe it is too soon but should we not specify this is for IPv4?

pkg/apis/core/v1alpha1/ignition_types.go
```diff
@@ -74,145 +73,250 @@ type Ignition struct {
...
 }
 
 type IgnitionSpecKubernetes struct {
-	API     IgnitionSpecKubernetesAPI     `json:"api" yaml:"api"`
-	DNS     IgnitionSpecKubernetesDNS     `json:"dns" yaml:"dns"`
-	Domain  string                        `json:"domain" yaml:"domain"`
+	// API holds information about the desired TC Kubernetes API.
+	API IgnitionSpecKubernetesAPI `json:"api" yaml:"api"`
+	// CloudProvider is the provider upon which the cluster is running. It is passed to API server as a flag.
+	CloudProvider string `json:"cloudProvider" yaml:"cloudProvider"`
```
omegas27: Why is this one in camel case when all the others are in lowercase?

pkg/apis/core/v1alpha1/ignition_types.go
```diff
@@ -74,145 +73,250 @@ type Ignition struct {
...
 type IgnitionSpecSSO struct {
+	// PublicKey is the public key of the SSO service.
 	PublicKey string `json:"publicKey" yaml:"publicKey"`
 }
 
+type IgnitionSpecOIDC struct {
+	// Enabled indicates that the OIDC settings should be applied when true.
+	Enabled bool `json:"enabled" yaml:"enabled"`
+	// The client ID for the OpenID Connect client, must be set if IssuerURL is set.
+	ClientID string `json:"clientId" yaml:"clientId"`
```
omegas27: Same questions regarding camelcase

pkg/apis/core/v1alpha1/ignition_types.go
tfussell: I discussed this with @MarcelMue last week and we decided that this should be a larger discussion. Once everyone agrees to update the CRD type version from `v1beta1` to `v1`, it should be applied across all CRDs at once. I'll try to push this forward soon. Good point though--we don't want to back ourselves into a corner like with the 1.16 deprecations.

pkg/apis/core/v1alpha1/ignition_types.go
tfussell: I felt that this YAML version was redundant since it can be read directly in the code following. I will add it back though since you find it useful. And I will be sure to involve SIG UX for CR related work in the future, thanks.

pkg/apis/core/v1alpha1/ignition_types.go
tfussell: IPv6 will have far-reaching implications across our code. It couldn't hurt to get started by being more specific when referring to IPs. I'll add this.

pkg/apis/core/v1alpha1/ignition_types.go
tfussell: Looks like we have a mix of camel and lower case throughout. I'll see if I can find a pattern and ensure that this is consistent with other CRDs.

pkg/apis/core/v1alpha1/ignition_types.go
xh3b4sd: It was super practical when working on Node Pools because you can copy/paste the structure and use it to apply via `kubectl` in your configuration. Also SIG UX may have an opinion on how informative and useful this is.

kopiczko: I think it would be cool to replace that with golden file tests.

tfussell: Golden files would be an option. In fact, to create the CR and CRD YAMLs I'm just writing a simple test that creates the object and prints it using the YAML encoder so it's the same idea.
I think the long-term solution is to generate sample CR and CRD files with `./scripts/generate.sh` like `kubebuilder` does and have a CI check that there is not a difference (so we don't make a change and forget to generate the corresponding code).

pkg/apis/core/v1alpha1/ignition_types.go
tfussell: Done.

pkg/apis/core/v1alpha1/ignition_types.go
tfussell: Every other CRD uses camel case only. I've updated this one to match.

pkg/apis/core/v1alpha1/ignition_types.go
tfussell: I'm still interested to get your opinions on how to do this in a more general way, but I have added comments with the YAML form of the CR and CRD so I'm going to call this particular comment done.

pkg/apis/core/v1alpha1/ignition_types.go
```diff
@@ -74,145 +174,250 @@ type Ignition struct {
...
-	IsMaster                bool                   `json:"ismaster" yaml:"ismaster"`
-	Kubernetes              IgnitionSpecKubernetes `json:"kubernetes" yaml:"kubernetes"`
+	// APIServerEncryptionKey is used in EncryptionConfiguration to encrypt Kubernetes secrets at rest.
+	APIServerEncryptionKey string `json:"apiServerEncryptionKey" yaml:"apiServerEncryptionKey"`
+	// BaseDomain is the base domain for all cluster services.
+	// For test installations, this may be in the form
+	// <clusterId>.k8s.<installation>.<region>.<provider>.gigantic.io.
+	BaseDomain string `json:"baseDomain" yaml:"baseDomain"`
+	// Calico provides configuration for all calico-related services.
+	Calico IgnitionSpecCalico `json:"calico" yaml:"calico"`
```
omegas27: Maybe we can talk about this some other time but why do we not name this CNI as we might switch to another one later (i.e. cillium)? Maybe we can have a field in the CNI spec for provider=calico?

pkg/apis/core/v1alpha1/ignition_types.go
marians: Having a YAML representation of both the CRD and an example CR was something we went with as an outcome of an internal discussion on how to make CRs more readable in the company. For me that was a great benefit. So thanks for adding it again!

marians (approved):

pkg/apis/core/v1alpha1/ignition_types.go
kopiczko: Yeah having a CI check would be nice. However having a golden file test is still beneficial and it's a fraction of the effort. I'd be fine (of others are) with dropping the comment when we have golden files.

pkg/apis/core/v1alpha1/ignition_types.go
tfussell: I think we'll have to change a lot of small things like this once/if we switch to Cilium. We'll come back to it later if needed. Good to be forward thinking though!

pkg/apis/core/v1alpha1/ignition_types.go
tfussell: This is done now. I'm generating the CR and CRD using the golden pattern, but putting it in `docs/` to be a little closer to `kubebuilder`. Should be nice for future documentation work. This does feel a little circular now as the CRD is in a string, the unmarshalled into a CRD struct, then marshalled to a YAML string and written to a file. Does this seem acceptable?

pkg/apis/core/v1alpha1/ignition_types.go
tuommaki: Alternatively: Would it be too much overhead/over engineered if this was and `ObjectReference` to CNI specific CR?

tfussell: An object reference makes sense to me. For the time being, it doesn't seem like it's providing much value if we don't have any concrete plans for allowing alternative CNIs though.

MarcelMue (approved):

pkg/apis/core/v1alpha1/ignition_types.go
kopiczko: I don't mind.

PR #331, opened 2020-02-19
fix godoc examples for control plane CR

xh3b4sd:


tomahawk28 (approved):
super

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -70,31 +70,31 @@ func NewAWSControlPlaneTypeMeta() metav1.TypeMeta {
...
-// AWSControlPlane is the infrastructure provider referenced in ControlPlane CRs.
+// AWSControlPlane is the infrastructure provider referenced in ControlPlane
+// CRs.
 //
 //     apiVersion: infrastructure.giantswarm.io/v1alpha2
 //     kind: AWSControlPlane
 //     metadata:
 //       annotations:
-//         giantswarm.io/docs:
+//         giantswarm.io/docs: https://docs.giantswarm.io/reference/awscontrolplanes.infrastructure.giantswarm.io/v1alpha2/
```
marians: I'd love this URL to exist already. However, it doesn't, yet. And I'm not sure it will look exactly like that. So I'd prefer to not have a URL for now. Or, alternatively, use a https://pkg.go.dev/github.com/giantswarm/apiextensions/ URL for now.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -75,29 +75,31 @@ func NewG8sControlPlaneTypeMeta() metav1.TypeMeta {
...
-//      name: 5f3kb
-//      apiVersion: infrastructure.giantswarm.io/v1alpha2
-//  status:
-//    replicas: 3
-//    readyReplicas: 3
+//     apiVersion: infrastructure.giantswarm.io/v1alpha2
+//     kind: G8sControlPlane
+//     metadata:
+//       annotations:
+//         giantswarm.io/docs: https://docs.giantswarm.io/reference/g8scontrolplanes.infrastructure.giantswarm.io/v1alpha2/
```
marians: See above regarding the URL

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
xh3b4sd: This is just an example and IMO it is ok that the URL does not work because the cluster ID and all the other information in there do also not work. Before there was not even a URL. So this is just an example structure, not a specific representation of something real.

marians: OK, fine!

Interestingly, it's pretty close to what I have written my wirewframes for CP API docs. :)

xh3b4sd: Nice we have a similar idea about the docs/wiki thing. 🙈

PR #334, opened 2020-02-25
Add version field to chart CRD

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/9079

Adds a version field to chart CRs. This means that chart-operator doesn't need to pull the chart tarball to find this version.

The version is already present in app CRs and is used to generate the tarball URL.


pkg/apis/application/v1alpha1/chart_types.go
```diff
@@ -61,7 +61,9 @@ spec:
             tarballURL:
               type: string
               format: uri
-          required: ["name", "namespace", "tarballURL"]
+            version:
```
yasn77: Would it be worth changing this to `chartVersion`?

rossf7: The field is on the chart CR. So I'd prefer to just call it version.

PR #335, opened 2020-02-25
Extend cluster guest type with internal domain

corest:
Towards https://github.com/giantswarm/giantswarm/issues/8979

I need this to be able to pass internal domain value into `certconfig` of legacy `cluster-operator`


xh3b4sd:
Why do we want this for legacy clusters thought? I thought this is full metal node pools.

corest:
I thought they need this for legacy.
Abandoning 
Discussion https://gigantic.slack.com/archives/C7865GLSY/p1582644809008100

PR #336, opened 2020-02-25
Remove chart for creating chart CRs

rossf7:
Towards giantswarm/roadmap#30

In https://github.com/giantswarm/chart-operator/pull/354 I've removed the usage of this chart. Once it's merged this chart can be removed.


rossf7:
Usage removed in giantswarm/chart-operator#354 going with a ping.

PR #338, opened 2020-02-27
Adding LastReconciled into chart, app CRD

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/8931

Spec on https://github.com/giantswarm/giantswarm/blob/master/sigs/monitoring/app_observability.md

Creating `LastReconciled` to get information about the last reconciled time on each CRs.


xh3b4sd (commented):
Why don't we extend [the `operatorkit` metrics resource](https://github.com/giantswarm/operatorkit/tree/master/resource/wrapper/metricsresource) instead and get timestamps for the CRs for ALL operators? I can see this being useful in a lot of projects. So we may do not need to come up with something unique for a single operator. The difference should then only be in the Prometheus query at the end of the observability process. @giantswarm/sig-monitoring does that make sense?

JosephSalisbury:
Or if not in the metrics resource specifically, at least more generically than just on the app resource

tomahawk28:
To make it clear, We discuss this issue earlier in a spec PR, https://github.com/giantswarm/giantswarm/blob/master/sigs/monitoring/app_observability.md
(check the PR following the link)
I didn't think about how operatorkit resource could extend to have observability features. 

Should we keep discuss how to improve the design here?

tomahawk28:
I checked metrics resource in https://github.com/giantswarm/operatorkit/tree/master/resource/wrapper/metricsresource, 
but still not sure how to calculate the time gap from reconciliation w/o any information from `status` resource. perhaps @JosephSalisbury have a better idea?

xh3b4sd:
The general idea would be to emit a metric for a CR when it got reconciled last. So the value is a timestamp. The metrics has labels for CR names or whatever. Then you need some other metric that has a timestamp as value. In your case for the Helm action or whatever. Then we can create Dashboards using these two metrics or do some math and create alerts for when there is too much gap between the first and the second event. We would need to talk this through a bit to see if this makes sense because of the reconciliation and so on.

xh3b4sd:
Cross posting. https://github.com/giantswarm/giantswarm/issues/6755 wants what you want.

tuommaki:
> The general idea would be to emit a metric for a CR when it got reconciled last. So the value is a timestamp.

Actually I think you don't even need to push the timestamp because prometheus does that already for every metric (*time*series, you know :wink: ) so it can be just `1` or something.

So then we could have one metric for present CR objects and one metric for scraping a CR and when combining those you could filter out present CRs that haven't been reconciled.

WDYT?

tomahawk28:
It's not about pushing timestamps into metrics, please read a spec. 
What I'm planning to do is calculating the time gap 
```
    current-time - lastReconciled = timegap 
```
I don't understand what problems lie ahead if I put a timestamp on this, moreover don't get the point why it didn't start on a spec PR but by now.

rossf7:
It would be nice to do something generic here. But I don't think it works with what we need to do

Like Ji says we want to measure the time gap between the CR event and the helm release. This could be over several reconciliation loops (e.g if the tenant cluster is unavailable).

Maybe comparing 2 metrics would work but how do we link them? With the current design we use the CR status which is what is being added here.

kopiczko:
I see the point of having that in the CR. This is the thing you want to check when debugging. Going to prometheus to read a value seems wrong and it's inaccurate (1 min delay). We can create a common annotation set by the operatorkit for that.

tomahawk28:
We initially start this idea with `annotation` but came back to `status` since we thought that last reconciliation timestamp belongs to status of CRs. 
But using the operatorkit and having new timestamps on status is not possible since we can't assume the type of `status` subresource. 

I'm keen to move this reconciliation timestamp into the annotation. WDYT?

rossf7:
I think an annotation makes sense and we document these in fmt. https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md

If we're not extending the CR status we don't need this PR. I'd like to see the existing spec updated based on these discussions.

tomahawk28:
Closing now.

PR #339, opened 2020-02-27
More ignition improvements

tfussell:
In an effort to establish best practices around our CRDs and CRs, I'm submitting this PR to further refine the ignition type. It includes the following changes:
- Strict CRD unmarshalling--if we have an extraneous field in the CRD multi-line string, this will now panic during `init`.
- Replace `ghodss/yaml` with `sigs.k8s.io/yaml`--after some recent discussion it was decided that we should prefer `sigs.k8s.io/yaml`. Interestingly, `ghodss/yaml` strict unmarshalling didn't seem to actually fail on extraneous fields so this is another reason to switch.
- Remove empty status from CRD string.
- Add shortName `ign` for convenience--allows `kubectl get ign`.
- Use `runtime.Object` type for test case objects instead of `interface{}`--ensures that code generation for `DeepCopy` is working.
- Improve test case name--now looks like `Test_GenerateYAML/case_0:_generates_core.giantswarm.io_ignition.yaml_successfully`.

If this PR is accepted, I will try to document these practices and bring these changes over to the [release type](https://github.com/giantswarm/apiextensions/pull/328).


kopiczko:
Can you also update https://github.com/giantswarm/fmt/blob/master/go/yaml_marshalling.md ?

pkg/apis/core/v1alpha1/ignition_types_test.go
```diff
@@ -55,7 +56,7 @@ func Test_GenerateYAML(t *testing.T) {
 	}
 
 	for i, tc := range testCases {
-		t.Run(fmt.Sprintf("case %d: %s", i, tc.name), func(t *testing.T) {
+		t.Run(fmt.Sprintf("case %d: generates %s successfully", i, tc.name), func(t *testing.T) {
```
kopiczko: I'd prefer to revert that change and improve `name: ` in each case instead. I can see myself confused when I'm looking for the case when a test failed.

kopiczko (approved):

pkg/apis/core/v1alpha1/ignition_types_test.go
tfussell: This should be fixed if you want to take a look.

kopiczko: Cool. Shouldn't we also have things like `case 0: ` in those test names?

tfussell:
> Can you also update https://github.com/giantswarm/fmt/blob/master/go/yaml_marshalling.md ?

Done in https://github.com/giantswarm/fmt/pull/98.

pkg/apis/core/v1alpha1/ignition_types_test.go
tfussell: This feels like yet another thing that can get out of sync and slow down PRs which add/remove/reorder test cases however if this is the accepted style for now, I will adjust the PR to match.

kopiczko: Described here https://github.com/giantswarm/fmt/blob/master/go/unit_tests.md#table-driven-tests
Feel free to propose a change here if you don't like it.

tfussell: Maybe I'll try to get it changed later. Updated to match `fmt` for now.

PR #344, opened 2020-03-03
Add changelog

tfussell:
To prepare for the (probably) final `dep`-based release, we need a changelog to keep track of versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,15 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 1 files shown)


kopiczko:
We still don't have `v0.1.0` tag. Can I also ask to make releases a separate PR with a name suggesting this will create a new release? Then it's easier for people to realise their changes are included or not.

kopiczko:
I see this https://github.com/giantswarm/apiextensions/pull/319#issuecomment-594162348. Then it's confusing we have 0.1.0 in changelog. It should be `Unreleased` IMO. Also `Unreleased` section is missing.

PR #345, opened 2020-03-03
Various release changes

tfussell:
Towards https://github.com/giantswarm/releases/pull/142

We can't set `creationTimestamp` to reflect the release date so we need a separate date in the `spec`. I added this field with tests and improved the example `Release` for our docs since it wasn't showing much before.

I also took this opportunity to remove `version` after a [team discussion](https://gigantic.slack.com/archives/CHAHY6VMF/p1582851678004100?thread_ts=1582851678.004100&cid=CHAHY6VMF). There are good reasons to keep it and good reasons to get rid of it. In general, Kubernetes resource names should not have meaning. The name is metadata and the important information should be in the spec and status. At the same time, we are able to validate names with OpenAPI and stateful sets do use meaningful pod names. Moreover, getting the version from the name ensures that names are unique. Lastly, the name and .spec.version will always be equal for the foreseeable future so this is just repeated information. I feel this is sufficient reason to go this route but am willing to consider other arguments to the contrary.

### DIFF
--- a/docs/crd/release.giantswarm.io_release.yaml
+++ b/docs/crd/release.giantswarm.io_release.yaml
@@ -13,7 +13,7 @@ spec:
     description: State of the release
     name: State
     type: string
-  - JSONPath: .metadata.creationTimestamp
+  - JSONPath: .spec.date
     description: Time since release creation
     name: Age
     type: date
@@ -94,6 +94,9 @@ spec:
                 type: object
               minItems: 1
               type: array
+            date:
+              format: date-time
+              type: string
             state:
               description: |
                 State indicates how this release should be used. "wip" means the release is a work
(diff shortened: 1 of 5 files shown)


docs/crd/release.giantswarm.io_release.yaml
```diff
@@ -102,16 +105,11 @@ spec:
                 supported release.
               pattern: ^(active|deprecated|wip)$
               type: string
-            version:
```
MarcelMue: Why was this field removed here? AFAIK we did not decide to remove it an the PR description does not give me a reason why we should remove it.

tfussell: Sorry about that. Not intending to be sneaky. I added my motivation to the PR description.

PR #346, opened 2020-03-04
deploy CRDs via Circle CI

xh3b4sd:
Coming from https://gigantic.slack.com/archives/C2MS2VB37/p1583333375093400.


helm/apiextensions/templates/g8scontrolplanes.infrastructure.giantswarm.io.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
kopiczko: I assume this is copy/paste.

.circleci/config.yml
```diff
@@ -21,3 +21,15 @@ workflows:
...
+    jobs:
+      - architect/push-to-app-catalog:
+          name: "push-apiextensions-to-control-plane-app-catalog"
+          app_catalog: "control-plane-catalog"
+          app_catalog_test: "control-plane-test-catalog"
+          chart: "apiextensions"
+          # Trigger job on git tag.
+          filters:
+            tags:
+              only: /^v.*/
```
kopiczko: I think you also want to push it to all app collections?

kopiczko: Since we're doing things on release I think it would be good a proper changelog here.

helm/apiextensions/templates/g8scontrolplanes.infrastructure.giantswarm.io.yaml
xh3b4sd: It is. Anything wrong with it?

.circleci/config.yml
xh3b4sd: You mean the changelog is missing in general or what has to be done here?

helm/apiextensions/Chart.yaml
```diff
@@ -0,0 +1,3 @@
+name: apiextensions
+version: 0.1.0
```
kopiczko:
```suggestion
version: [[ .Version ]]
```

helm/apiextensions/templates/g8scontrolplanes.infrastructure.giantswarm.io.yaml
kopiczko: No that's good.

.circleci/config.yml
kopiczko: You do.

kopiczko: Yeah I mean the changelog is missing. For me it's important to have it before creating the first relese. Example https://github.com/giantswarm/microerror/blob/master/CHANGELOG.md

xh3b4sd: Will do once it works.

xh3b4sd: Why is that though? What is the difference/purpose of the catalog/collection? I miss context here.

kopiczko: The catalog is equivalent of quay.io, i.e. some registry that stores charts. The collection is what is actually installed, a place that holds App CRs that install those charts from catalog.

xh3b4sd: Let's talk about it tomorrow. I don't follow.

xh3b4sd: Added changelogs.

xh3b4sd:
This here does not work. Will write up a summary later.

PR #348, opened 2020-03-05
Release v0.1.0

tfussell:
Towards https://github.com/giantswarm/apiextensions/pull/319

In accordance with https://github.com/giantswarm/giantswarm/issues/9251, I am creating a dedicated PR for releasing `apiextensions` v0.1.0 before merging in the go modules migration. I will tag it as soon as it is merged to master. Further work will go into an eventual v0.1.1 (or v0.2.0 if the go modules PR is merged) release. Consumers of this library should use a `dep` constraint/override of `~0.1.0` or `=0.1.0`.


MarcelMue (approved):
This seems fine to me from my understanding. Maybe Pawel can confirm :P

mig4 (approved):

kopiczko (approved):

kopiczko:
I think we should use `~0.1.0` only. Otherwise we may need to do `[override]`.

PR #350, opened 2020-03-06
Enhance AWSCluster to provide YAML source for documentation

marians:

The purpose of this PR:

- Enable the generation of docs for the `AWSCluster` CRD (`awsclusters.infrastructure.giantswarm.io`).
- Provide an example for other CRDs.


pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
```diff
@@ -35,27 +35,49 @@ spec:
...
                           type: object
                         region:
+                          description: |
+                            AWS region the cluster is to be running in.
                           type: string
                       type: object
                   type: object
+                oidc:
+                  description: |
+                    Configuration for OpenID Connect (OIDC) authentication.
```
marians: Heads-up: The `oidc` property was missing completely, so I added it.

marians:
FYI: I decided to not add any schema for the .status part of the CRD. In order to provide an example that includes CRD and CR file generation, that isn't needed. Like it will also create a completely different type of discussion. Let's tackle that separately.

pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
```diff
@@ -7,7 +7,9 @@ import (
 )
 
 const (
-	kindAWSCluster = "AWSCluster"
+	crDocsAnnotation            = "giantswarm.io/docs"
+	kindAWSCluster              = "AWSCluster"
+	awsClusterDocumentationLink = "https://pkg.go.dev/github.com/giantswarm/apiextensions/pkg/apis/infrastructure/v1alpha2?tab=doc#AWSCluster"
```
tfussell: Eventually this will point to our own docs, but it seems like the best choice until then.

pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
marians: Extra info-ping to @giantswarm/team-firecracker for adding the `oidc` key here.

PR #352, opened 2020-03-09
Enhance App CRD for documentation

marians:

- Adds description to schema
- Adds creation of CRD and CR YAML files to unit tests

### DIFF
--- a/docs/cr/application.giantswarm.io_v1alpha1_app.yaml
+++ b/docs/cr/application.giantswarm.io_v1alpha1_app.yaml
@@ -1,0 +1,33 @@
+apiVersion: application.giantswarm.io/v1alpha1
+kind: App
+metadata:
+  annotations:
+    giantswarm.io/docs: https://pkg.go.dev/github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1?tab=doc#App
+  creationTimestamp: null
+  name: prometheus
+spec:
+  catalog: my-catalog
+  config:
+    configMap:
+      name: my-configmap
+      namespace: monitoring
+    secret:
+      name: my-secret
+      namespace: monitoring
+  kubeConfig:
+    context:
+      name: my-context-name
+    inCluster: false
+    secret:
+      name: my-kubeconfig-secret
+      namespace: monitoring
+  name: prometheus
+  namespace: monitoring
+  userConfig:
+    configMap:
+      name: my-userconfig-configmap
+      namespace: monitoring
+    secret:
+      name: my-userconfig-secret
+      namespace: monitoring
+  version: 1.0.0
--- a/docs/crd/application.giantswarm.io_app.yaml
+++ b/docs/crd/application.giantswarm.io_app.yaml
@@ -100,7 +100,7 @@ spec:
               type: object
             name:
               description: |
-                Name of this App.
+                Name of this App in the AppCatalog.
               type: string
             namespace:
               description: |
--- a/docs/crd/application.giantswarm.io_app.yaml
+++ b/docs/crd/application.giantswarm.io_app.yaml
@@ -100,7 +100,7 @@ spec:
               type: object
             name:
               description: |
-                Name of this App in the AppCatalog.
+                Name of this App.
               type: string
             namespace:
               description: |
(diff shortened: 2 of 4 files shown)


pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -158,17 +158,17 @@ func NewAppTypeMeta() metav1.TypeMeta {
...
+//         inCluster: false
+//         context:
+//           name: "giantswarm-12345"
+//         secret:
+//           name: "giantswarm-12345"
+//           namespace: "giantswarm"
+//         userConfig:
+//           configMap:
+//             name: "prometheus-user-values"
+//             namespace: "monitoring"
```
marians: Indentation here was broken

docs/cr/application.giantswarm.io_v1alpha1_app.yaml
```diff
@@ -0,0 +1,33 @@
...
+    inCluster: false
+    secret:
+      name: my-kubeconfig-secret
+      namespace: monitoring
+  name: prometheus
+  namespace: monitoring
+  userConfig:
+    configMap:
+      name: my-userconfig-configmap
+      namespace: monitoring
```
oponder: Is this example supposed to be as realistic as possible?

The namespace for `userConfig.configMap.namespace` would most likely be a tenant cluster ID in most situations.

I think the other namespaces would also be tenant cluster id's, and possibly have other names defined by a convention. Let me know if this should be more 'real' or is just used in some automation.

docs/crd/application.giantswarm.io_app.yaml
```diff
@@ -0,0 +1,155 @@
...
+                        Namespace holding the Secret resource.
+                      type: string
+                  required:
+                  - name
+                  - namespace
+                  type: object
+              type: object
+            name:
+              description: |
+                Name of this App.
```
oponder:
```suggestion
                Name of this App in the AppCatalog.
```

docs/cr/application.giantswarm.io_v1alpha1_app.yaml
marians: If I just make the example namespace something like `f74io`, which could be a tenant cluster ID in practice, it would require some context to explain that for apps deployed in the tenant cluster this should be the tenant cluster ID.

I don't think it is realistic to provide all that context within the CRD documentation.

And then, is it not true that an App CR can also represent a control plane app?

docs/crd/application.giantswarm.io_app.yaml
```diff
@@ -0,0 +1,155 @@
...
+    plural: apps
+    singular: app
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: |
+        Defines an App resource, which represents an application to be running in a Kubernetes cluster.
+        Reconciled by app-operator.
```
marians: @oponder Please also check whether this `Reconciled by app-operator` is true.

docs/cr/application.giantswarm.io_v1alpha1_app.yaml
oponder: > And then, is it not true that an App CR can also represent a control plane app?

Yes it's true. Hmmm....

I was looking at these docs from the point of view of a customer that might want to install an app on their tenant cluster though.

Install apps on the control plane is usually a giant swarm admin concern. 

But I leave it to your discretion on exactly what context / intention these docs will most likely be viewed in.

docs/crd/application.giantswarm.io_app.yaml
oponder: Yup that's right :+1:

`app-operator` then creates some Chart CR's in the tenant cluster (or control plane). Whatever the kubeConfig points to actually.

docs/cr/application.giantswarm.io_v1alpha1_app.yaml
marians: I think having the schema docs will not save us from providing more examples in guides for certain use cases.

marians:
Still looking for a general approval here 👀

PR #353, opened 2020-03-09
Enhance AppCatalog CRD for documentation

marians:

- Adds description to schema
- Adds creation of CRD and CR YAML files to unit tests


docs/crd/application.giantswarm.io_appcatalog.yaml
```diff
@@ -0,0 +1,95 @@
...
+    kind: AppCatalog
+    plural: appcatalogs
+    singular: appcatalog
+  scope: Cluster
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: |
+        An AppCatalog defines a collection of apps to install in one or several clusters.
```
rossf7:
```suggestion
        An AppCatalog defines a collection of apps that can be installed in one or more clusters.
```

This reads to me as if all apps need to be installed when actually you don't need to install any.

We've talked before about having an "app stack" to have a higher level abstraction for installing multiple apps. So I'd like it to be clearer that the catalog is just a grouping of apps that are stored together.

PR #359, opened 2020-03-16
Fix go vet error with chart DeepCopyTime

rossf7:
I'm trying to use `go-build` from architect-orb in https://github.com/giantswarm/chart-operator/pull/374 

But hitting a go vet error `composite literal uses unkeyed fields` with DeepCopyTime. Should I also do the other functions? As we'll hit this in more places eventually.

### DIFF
--- a/pkg/apis/application/v1alpha1/deep_copy.go
+++ b/pkg/apis/application/v1alpha1/deep_copy.go
@@ -7,7 +7,7 @@ import (
 // DeepCopyTime implements the deep copy logic for time.Time which the k8s
 // codegen is not able to generate out of the box.
 type DeepCopyTime struct {
-	time.Time
+	DeepCopyTime time.Time
 }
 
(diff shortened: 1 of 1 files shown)


pkg/apis/application/v1alpha1/deep_copy.go
```diff
@@ -7,7 +7,7 @@ import (
 // DeepCopyTime implements the deep copy logic for time.Time which the k8s
 // codegen is not able to generate out of the box.
 type DeepCopyTime struct {
-	time.Time
+	DeepCopyTime time.Time
```
kopiczko: I somehow dislike that because it's more difficult to realise this is an embedded struct. But if `go vet` says so...

kopiczko:
> Should I also do the other functions?

If we agree on this approach then I think yes.

tomahawk28:
Agreed with you.

rossf7:
Inbox cleanup. This is already done.

PR #361, opened 2020-03-18
Add kube proxy configuration to cluster type

tfussell:
Enables https://github.com/giantswarm/k8scloudconfig/pull/659 and ultimately towards https://github.com/giantswarm/giantswarm/issues/9154

Options come from https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-proxy/app/server.go#L182-L199 I don't think we have plans to use all of them immediately, but it's a marginal amount of effort to add them all, and it should improve flexibility in the future.

~~`conntrack-max` is deprecated so I'm not sure if we should support it or not.~~
`conntrack-max` is gone in newer `kube-proxy` so I'll remove now.

I will backport this to `legacy-1-15` if accepted for use in 1.15 operator branches.


tfussell:
> I don't love it but it gets the job done I guess 👍

Any ideas for improving? I didn't see anywhere else to put this, but I'm not attached to this approach.

tfussell:
I guess this also needs to be in `AWSCluster` and `Ignition`. I'll do those in separate PRs once merged.

PR #362, opened 2020-03-19
Fix legacy codegen

MarcelMue:


MarcelMue:
Would be cool to get some :eyes: on this @giantswarm/team-ludacris

.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  architect: giantswarm/architect@0.8.3
+  architect: giantswarm/architect@0.7.0
```
omegas27: Why did you revert to this one?

stone-z: I bet he didn't want to lint XD 
https://circleci.com/gh/giantswarm/apiextensions/1610?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

MarcelMue: I bumped this in a previous PR but it can't succeed because the generated code does not match the code guidelines :man_facepalming: 

We also use the older version in master:
https://github.com/giantswarm/apiextensions/blob/master/.circleci/config.yml#L4

tfussell: I added `//nolint` to a few of the generated files so we could use 0.8.3, but this isn't ideal. We could add some options to the linter or get rid of the generated clients which have these problems. Probably not worth it for now.

PR #364, opened 2020-03-20
Fix application crd errors

rossf7:
This fixes errors on boot in app-operator and chart-operator. The openapi validation root element needs to be of type object.

```
	/Users/rossf7/.gvm/pkgsets/go1.13.1/global/src/github.com/giantswarm/chart-operator/vendor/github.com/giantswarm/k8sclient/k8scrdclient/k8scrdclient.go:178: spec.validation.openAPIV3Schema.type: Required value: must not be empty at the root
	not established error
```

See upstream issue. https://github.com/kubernetes/kubernetes/issues/81445


rossf7:
Going with a ping as these errors on boot are blocking me in app-operator and chart-operator.

@xh3b4sd Can you hold off merging the go modules change till I've released `0.1.2`?

angelbarrera92:
I know is not fully related, but if you try to deploy: 
- https://github.com/jetstack/cert-manager/releases/download/v0.14.1/cert-manager-legacy.crds.yaml
In a Kubernetes > 1.14 Cluster....... the only way to make it working is adding the `type: object`.

Thanks google + github for this awesome "indexing engine"!

PR #369, opened 2020-03-23
legacy: Remove unused kube-proxy flags from cluster

tfussell:
Since we decided we only wanted to keep `--conntrack-max-per-core`, I am getting rid of the other kube-proxy flags before we merge https://github.com/giantswarm/aws-operator/pull/2218


MarcelMue:
Merging this so I can test 9.0.1

PR #371, opened 2020-03-25
Add spot instances configuration

MarcelMue:


teemow:
LGTM. AWS has chosen a different path in terms of configuring this. In the MixedInstancePolicy you can set on-demand and spot-instance parameters. But I think it is fine to add a struct for the spot instances configuration only. This can then be mapped in the provider specific operators accordingly.

MarcelMue:
We can change this later on easily. Let's see later if this is sufficient.

teemow:
We also need to update the status:

 * Which alike instance types are allowed?
 * How many spot instances are used in the cluster?
 * How many of which instance type is used? (not sure if we really need this now - we should instead have a representation of the instances in a node pool in the future)

docs/crd/infrastructure.giantswarm.io_awsmachinedeployment.yaml
```diff
@@ -83,6 +83,26 @@ spec:
...
                       type: string
                     type: array
+                  instanceDistribution:
+                    description: |
+                      Attributes defining the instance distribution in a node pool.
+                    properties:
+                      onDemandBaseCapacity:
+                        default: 0
+                        description: |
+                          Base capacity of on demand machines.
```
marians:
```suggestion
                          Base capacity of on-demand machines.
```

marians:
```suggestion
                          Base capacity of on-demand instances to use for worker nodes.
```

docs/crd/infrastructure.giantswarm.io_awsmachinedeployment.yaml
```diff
@@ -83,6 +83,26 @@ spec:
...
+                        format: int32
+                        minimum: 0
+                        type: integer
+                      onDemandPercentageAboveBaseCapacity:
+                        default: 100
+                        description: |
+                          Percentage of on demand instances above the base capacity.
```
marians:
```suggestion
                          Percentage of on-demand instances to use for worker nodes, for instances exceeding onDemandBaseCapacity.
```

docs/crd/infrastructure.giantswarm.io_awsmachinedeployment.yaml
```diff
@@ -91,6 +111,33 @@ spec:
                         description: |
                           AWS EC2 instance type name to use for the worker nodes in this node pool.
                         type: string
+                      useAlikeInstanceTypes:
+                        description: |
+                          If true, instances alike the instance_type will be used.
```
marians:
```suggestion
                          If true, certain instance types with specs similar to .spec.provider.worker.instanceType will be used.
```

marians:
Kudos for also adding the `.status` section to the validation schema and thus publishing docs for this!

docs/crd/infrastructure.giantswarm.io_awsmachinedeployment.yaml
marians: Sorry, my change suggestion is in the wrong place. Will move.

docs/crd/infrastructure.giantswarm.io_awsmachinedeployment.yaml
marians: Sorry, my change suggestion is in the wrong place. Will move.

docs/crd/infrastructure.giantswarm.io_awsmachinedeployment.yaml
marians: Sorry, my change suggestion is in the wrong place. Will move.

pkg/apis/infrastructure/v1alpha2/aws_machine_deployment_types.go
```diff
@@ -99,6 +99,26 @@ spec:
...
                       type: string
                     type: array
+                  instanceDistribution:
+                    description: |
+                      Attributes defining the instance distribution in a node pool.
+                    properties:
+                      onDemandBaseCapacity:
+                        default: 0
+                        description: |
+                          Base capacity of on demand machines.
```
marians:
```suggestion
                          Base capacity of on-demand instances to use for worker nodes.
```

pkg/apis/infrastructure/v1alpha2/aws_machine_deployment_types.go
```diff
@@ -99,6 +99,26 @@ spec:
...
+                        format: int32
+                        minimum: 0
+                        type: integer
+                      onDemandPercentageAboveBaseCapacity:
+                        default: 100
+                        description: |
+                          Percentage of on demand instances above the base capacity.
```
marians:
```suggestion
                          Percentage of on-demand instances to use for worker nodes, for instances exceeding onDemandBaseCapacity.
```

pkg/apis/infrastructure/v1alpha2/aws_machine_deployment_types.go
```diff
@@ -108,6 +128,33 @@ spec:
                         description: |
                           AWS EC2 instance type name to use for the worker nodes in this node pool.
                         type: string
+                      useAlikeInstanceTypes:
+                        description: |
+                          If true, instances alike the instance_type will be used.
```
marians:
```suggestion
                          If true, certain instance types with specs similar to instanceType will be used.
```

pkg/apis/infrastructure/v1alpha2/aws_machine_deployment_types_test.go
```diff
@@ -92,8 +92,24 @@ func newAWSMachineDeploymentExampleCR() *AWSMachineDeployment {
...
+		Provider: AWSMachineDeploymentStatusProvider{
+			Worker: AWSMachineDeploymentStatusProviderWorker{
+				InstanceTypes: []string{
+					"m4.xlarge",
+					"m5.xlarge",
+				},
+				SpotInstances: 39,
 			},
 		},
 	}
```
marians: I am thinking that a realistic example should have even more Status attributes defined. Can we add that as part of this PR, or should we make that separate?

MarcelMue: This is a complete example of all Status attributes of this CR - what others did you have in mind? 

This matches pretty much what we see in clusters currently.

marians: I was expecting the desired and actual number of instances, coming from the cluster-autoscaler. Is that in a different CR then? Anyway, comment solved.

MarcelMue: Yes that is different CR.

PR #375, opened 2020-03-26
Document Cluster CRD

marians:

Changes:

- Lets the CRD be populated from YAML string
- Create CRD and CR YAML files via unit tests

As a prerequisite this imports new dependencies, which had the effect of updating existing dependency versions.

*Note:* This PR does not add a validation schema nor property descriptions.


docs/crd/infrastructure.giantswarm.io_cluster.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: clusters.cluster.x-k8s.io
```
xh3b4sd: I think the file name should align with this one.

pkg/apis/infrastructure/v1alpha2/cluster_api_crds.go
```diff
@@ -1,10 +1,56 @@
 package v1alpha2
 
 import (
+	"fmt"
+
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterapiv1alpha2 "sigs.k8s.io/cluster-api/api/v1alpha2"
```
xh3b4sd: The alias should be `clusterapiapiv1alpha2` or `apiv1alpha2` to stay consistent with the actual import path.

pkg/apis/infrastructure/v1alpha2/cluster_api_crds.go
```diff
@@ -1,10 +1,56 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	kindCluster = "Cluster"
+	// TODO: Change to this CRD's docs URL once published.
```
xh3b4sd: Do we track this? We usually have TODO issues for this so people or teams can take care and do not forget about it.

marians: I am optimistic that I will re-visit those as part of my work on https://github.com/giantswarm/giantswarm/issues/8804

pkg/apis/infrastructure/v1alpha2/cluster_api_crds.go
marians: Changed

docs/crd/infrastructure.giantswarm.io_cluster.yaml
marians: Changing that system here now would derive from other CRDs and break crd-docs-generator. If it is important to you I suggest to create an issue and revisit later. OK?

xh3b4sd: Doesn't make sense at all. The file name reflects the apiversion group. `infrastructure.giantswarm.io` is ours, Giant Swarm. The CRD here is upstream, CAPI, the community. I think this needs fixing.

pkg/apis/infrastructure/v1alpha2/cluster_api_crds.go
xh3b4sd: Your call, I just want to say that I heard that often from all kinds of people and then after two years we found a TODO comment somewhere, which is why we started to adhere to some guidelines and created the TODO issues which we linked in the code as well. See https://github.com/giantswarm/fmt/blob/master/go/todo_comments.md.

docs/crd/infrastructure.giantswarm.io_cluster.yaml
marians: I'd like to hear @tfussell 's, as he introduced this scheme.

Would `docs/crd/clusters.cluster.x-k8s.io.yaml` make sense as a file name for the  CRD?

The CR would then be `docs/cr/clusters.cluster.x-k8s.io_v1alpha2.yaml`.

This would have to change for all YAML files consistently.

pkg/apis/infrastructure/v1alpha2/cluster_api_crds.go
marians: Changed

docs/crd/infrastructure.giantswarm.io_cluster.yaml
tfussell: I chose this naming roughly based on what `kubebuilder` does. I agree that it should be the actual API group rather than where it's located in our repo. I didn't even realize that we had upstream CRDs in our apiextensions so I guess this was an oversight on my part in the initial implementation. On the other hand, should we even be documenting upstream types though?

marians: I'd very much like to document the Cluster API ones, since customers have to deal with them. Reasons:

- The right place for documentation of the ones we actually use is otherwise hard to find.
- The documentation is sparse and since it is Go-type based, it is hard to read. See https://pkg.go.dev/sigs.k8s.io/cluster-api@v0.3.3/api/v1alpha2?tab=doc#Cluster

tfussell: I see what you mean. There could even be an opportunity to contribute to upstream CAPI documentation with this in that case.

As far as next steps, I can try to improve the CR/CRD YAML generation in the following three ways:
- Use the actual API group for the generated file path (e.g. `clusters.cluster.x-k8s.io`).
- Centralize the CR/CRD YAML generation code into some common helper functions so we're not copying and pasting code for every type.
- Invert the process so the YAML is the source of truth and we just deserialize/unmarshal them into structs as needed.

What do y'all think?

marians: 1 and 2 sounds very good.

3 may need more discussion. Having the CRD YAML, the actual source of truth, located in a folder called `docs` doesn't sound right to me. Just a detail.

marians: If you @tfussell will take care of this globally, can we merge the PR as it is for now? If yes, it needs some approval.

tfussell: I made a note in my team board to take care of 1 and 2. I have ideas for how 3 might work, but I will need to build a prototype before I try to push it here.

go.mod
```diff
@@ -4,13 +4,14 @@ go 1.13
...
 	github.com/go-openapi/errors v0.19.4
-	github.com/google/go-cmp v0.3.0
-	k8s.io/api v0.16.6
-	k8s.io/apiextensions-apiserver v0.16.6
-	k8s.io/apimachinery v0.16.6
-	k8s.io/client-go v0.16.6
-	k8s.io/code-generator v0.16.6
-	sigs.k8s.io/yaml v1.1.0
+	github.com/google/go-cmp v0.4.0
+	k8s.io/api v0.17.2
```
tfussell, line 8: I'm not sure about these changes. 1.17 should be backwards compatible, but I'd prefer if we use the same version in all of our repos.

tfussell:
I think this is ready to go after `go.mod` versions are fixed and the branch is updated from master.

go.mod
marians: Actually I wasn't able to prevent this as soon as I added `sigs.k8s.io/cluster-api` as a dependency. Maybe I could try with an older different versions of that.

tfussell: I didn't even notice that addition of `sigs.k8s-.io/cluster-api`. There was a big headache moving from CAPI v1alpha1 to v1alpha2 so we need to be careful with tying our apiextensions to a specific CAPI version and the effects it might have on operators. Maybe a PE more involved with CAPI like @xh3b4sd or @MarcelMue could take a look--is this dependency alright? Should it be v0.2.10?

xh3b4sd: This is due to the fact that we have to create the `Cluster` CRD for instance. So importing the right version makes sense and should be straight forward, especially now that we tag the library. With a new version comes a new tag and the operator that wants to make use of it uses the desired tag. I think this is all fine, or, I do not see a better way of doing it right now.

MarcelMue: I think we are currently pretty tied to `v1alpha2` of CAPI in here. So I do not see a blocker here. Moving from `v1alpha2` to `v1alpha3` should not be as big as a headache ( it was mainly the breaking changes from v1alpha1 to 2).

tfussell: Thanks. If I'm understanding correctly, then I would recommend to @marians that these be reverted to `v0.16.6` and cluster-api to `v0.2.10` to keep things consistent.

marians: That was super helpful, indeed it seems to work with these versions.

tfussell (approved):
LGTM. I'll try to get the other improvements discussed in this PR done on Monday.

PR #385, opened 2020-03-31
Add third AZ to AWSControlPlane example

marians:
I figured that the example I crafted would be invalid according to our own description.


marians:
In our UIs we will not support 2 AZs, so I just wanted to be consistent. Might be that it actually works from a technical side.

PR #386, opened 2020-04-02
Add validation for AzureConfig

fiunchinho:


tfussell:
Is this still relevant? If so it should probably be translated into kubebuilder comments.

fiunchinho:
Not anymore, thanks.

PR #393, opened 2020-04-06
add VPC ID to AWSCluster CR

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/9274. 

## Checklist

### DIFF
--- a/pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
+++ b/pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
@@ -178,6 +178,7 @@ func NewAWSClusterCR() *AWSCluster {
 //       provider:
 //         network:
 //           cidr: 10.1.6.0/24
+//           vpcID: vpc-1234567890abcdef0
 //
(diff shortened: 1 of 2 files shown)


pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
```diff
@@ -271,6 +272,8 @@ type AWSClusterStatusProvider struct {
 type AWSClusterStatusProviderNetwork struct {
 	// IPv4 address block used by the tenant cluster, in CIDR notation.
 	CIDR string `json:"cidr" yaml:"cidr"`
+	// VPCID contains the ID of the tenant cluster, e.g. vpc-1234567890abcdef0.
+	VPCID string `json:"vpcID" yaml:"vpcID"`
```
xh3b4sd, line 276: So as discussed with @marians we thought to go with the most simple change which is exposing the VPC ID in the CR status. In `aws-operator` we would then only need to fill it accordingly. Then customers can read the information from there.

PR #397, opened 2020-04-07
Add Azure Tools types

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/10112

Add CRDs for https://github.com/giantswarm/azure-tools-operator


pkg/apis/tools/v1alpha1/azure_tool_types.go
```diff
@@ -0,0 +1,80 @@
...
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+)
+
+const azureToolCRDYAML = `
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: azuretool.tools.giantswarm.io
```
omegas27, line 13: it can work for the hackaton but I would go with azure.tooling.giantswarm.io if this works :D

pkg/apis/tools/v1alpha1/azure_tool_types.go
```diff
@@ -0,0 +1,80 @@
...
+spec:
+  group: tools.giantswarm.io
```
omegas27, line 15: same, I would go with tooling.giantswarm.io

omegas27:
You rocking

axbarsan:
I have to merge master and add v1 version

tuommaki:
Something's off here. Generator produced `v1` and `v1beta1` YAMLs for the CRD, but the Go type is defined for `v1alpha1`. Possibly some local leftovers and need for one more run of `./scripts/gen.sh` or does the generator somehow misbehave here?

axbarsan:
> Something's off here. Generator produced `v1` and `v1beta1` YAMLs for the CRD, but the Go type is defined for `v1alpha1`. Possibly some local leftovers and need for one more run of `./scripts/gen.sh` or does the generator somehow misbehave here?

That is a bit unclear to me, I pretty much followed the implementation of other CRDs. The CRD is defined as `v1` here.
https://github.com/giantswarm/apiextensions/blob/2f6d03719e8c667f110e84fff3a402e4cbeee578/pkg/apis/tooling/v1alpha1/azure_tool_types.go#L14

But the folder name is `v1alpha1`, which makes it be generated as `v1alpha1` inside the clients. This is strange, but it is also the way all the other CRDs are used right now.

tuommaki:
> This is strange, but it is also the way all the other CRDs are used right now.

Ah, right. I mixed the version of used `apiextensions` API and the version of your CRD. All good. :+1:

PR #401, opened 2020-04-09
Add schema and property descriptions for CertConfig

marians:

This adds a schema to the CRD version v1alpha1 with the goal to document the properties.


pkg/apis/core/v1alpha1/cert_types.go
```diff
@@ -32,6 +32,73 @@ spec:
...
+                  clusterID:
+                    description: |
+                      Unique identifier of the tenant cluster this certificate is for.
+                    type: string
+                  commonName:
+                    description: |
+                      The value of the Common Name (CN) attribute of the certificate.
+                  disableRegeneration:
+                    description: |
+                      Toggles the automatic recreation before expiry.
```
marians, line 77: I wasn't sure about this. Could someone please verify?

omegas27: This boolean just tells cert operator if it should disable certificate renewal. It updates the secret on the control plane, not the nodes themselves. I do not understand why we added that but the message should be correct. Maybe it can be improved though using "Disable automatic certificate rotation before expiry".

marians: Done, thanks

xh3b4sd: We added this "feature" for service account certificates so that they are only generated once and then never change. These would then have had a super long TTL. I am not sure if this is still used. @calvix may know more about the changes we made in the Control Plane to make this obsolete.

PR #403, opened 2020-04-09
Fix group name in generated YAML filenames of Cluster and MachineDeployment

marians:


marians:
Still need to fix something here...

pkg/apis/infrastructure/v1alpha2/cluster_api_crds_test.go
```diff
@@ -107,19 +113,24 @@ func Test_NewMachineDeploymentCRD(t *testing.T) {
 }
 
 func Test_GenerateMachineDeploymentYAML(t *testing.T) {
+	crd := NewMachineDeploymentCRD()
+
+	crdGroup := crd.Spec.Group
+	crdKindLower := strings.ToLower(crd.Spec.Names.Kind)
```
tfussell, line 119: I've been playing around with `kubebuilder` recently and discovered that the suffix should actually be `crd.Spec.Names.Plural` (so `cluster.x-k8s.io_machinedeployments.yaml`) I plan to make this change in my upcoming PR so it's not super important here--just wanted to mention.

PR #405, opened 2020-04-09
Document G8sControlPlane CRD

marians:

- Adds YAML CRD and CR files generated by unit tests
- Adds property descriptions to the schema

### DIFF
--- a/docs/crd/infrastructure.giantswarm.io_g8scontrolplane.yaml
+++ b/docs/crd/infrastructure.giantswarm.io_g8scontrolplane.yaml
@@ -1,0 +1,46 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: g8scontrolplanes.infrastructure.giantswarm.io
+spec:
+  conversion:
+    strategy: None
+  group: infrastructure.giantswarm.io
+  names:
+    kind: G8sControlPlane
+    plural: g8scontrolplanes
+    singular: g8scontrolplane
+  scope: Namespaced
+  subresources:
+    status: {}
+  versions:
+  - name: v1alpha1
+    served: false
+    storage: false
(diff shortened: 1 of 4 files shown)


docs/crd/infrastructure.giantswarm.io_g8scontrolplane.yaml
```diff
@@ -0,0 +1,55 @@
...
+          This resource represents an abstract control plane (Kubernetes master node and Etcd server)
+          of a tenant cluster in a Giant Swarm installation. It is reconciled by cluster-operator.
+          For implementation details it references a provider specific resource.
+        properties:
+          spec:
+            properties:
+              infrastructureRef:
+                description: |
+                  Reference to an [AWSControlPlane](https://docs.giantswarm.io/reference/cp-k8s-api/awscontrolplanes.infrastructure.giantswarm.io/)
+                  resource defining provider specific details for the c
```
xh3b4sd, line 34: For the what? There got probably something lost.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -79,6 +89,18 @@ func NewG8sControlPlaneTypeMeta() metav1.TypeMeta {
 	}
 }
 
+// NewG8sControlPlaneCR returns a Cluster Custom Resource.
```
xh3b4sd, line 92: I am not sure "Cluster Custom Resource" is correct in this case. It is also that the value of the type returned is mostly empty. Do we really need that function?

marians, line 92:
```suggestion
// NewG8sControlPlaneCR returns a G8sControlPlane Custom Resource.
```

marians: Fixed

PR #409, opened 2020-04-13
Generate CRDs using controller-tools

tfussell:
Unhappy with our current setup around CRDs, I thought it was time to try upstream Kubernetes `controller-tools` to generate CRDs from CRs (and extra metadata from comments, see https://book.kubebuilder.io/reference/generating-crd.html). For now, I would just like a review on the resulting CRDs and whether we could migrate to these CRDs in the near future without breaking anything.

This PR does not include the additional step of converting CRD YAMLs to functions that return `apiextensionsv1beta1.CustomResourceDefinition` such as `NewG8sControlPlaneCRD()`. This can be done in a separate PR since this PR doesn't actually affect any code yet, just the `docs/` directory. Until then there would be duplicate CRD definitions, those in the multiline strings in source code and the generated CRDs in `docs/`. This would be a temporary state.

An extra benefit of this approach is that we can have full schemata for all types which is required for moving to `apiextensionsv1.CustomResourceDefinition` due to enforced pruning. See https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#controlling-pruning for more info. This is currently a blocker for spot instances as I understand it.

## Checklist


docs/crd/application.giantswarm.io_app.yaml
```diff
@@ -1,167 +0,0 @@
-apiVersion: apiextensions.k8s.io/v1beta1
```
tfussell, line 1: `controller-gen` names CRD files like `<group>_<plural name>.yaml` so this file has been renamed `docs/crd/application.giantswarm.io_apps.yaml`.

docs/crd/application.giantswarm.io_appcatalogs.yaml
```diff
@@ -0,0 +1,122 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: (devel)
```
tfussell, line 7: It would be preferable to use a released binary of `controller-gen` somehow to avoid this label.

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -115,21 +115,6 @@ func init() {
 }
 
 // NewAppCatalogCRD returns a new custom resource definition for AppCatalog.
-// This might look something like the following.
```
tfussell, line 118: We don't need these comments now that we have full YAMLs for all CRDs.

scripts/gen.go
```diff
@@ -6,4 +6,5 @@ package scripts
 import (
 	_ "k8s.io/code-generator/cmd/client-gen"
 	_ "k8s.io/code-generator/cmd/deepcopy-gen"
+	_ "sigs.k8s.io/controller-tools/cmd/controller-gen"
```
tfussell, line 9: Ensuring the correct version of `controller-gen` is tracked in `go.mod` using this fake import.

scripts/gen.sh
```diff
@@ -25,3 +25,6 @@ if [ ! -x "$(command -v goimports)" ]; then
 fi
 echo "Fixing imports in-place with goimports"
 goimports -local github.com/giantswarm/apiextensions -w ./pkg
+
+go run "$(go list -m -f '{{.Dir}}' sigs.k8s.io/controller-tools)"/cmd/controller-gen/main.go  \
```
tfussell, line 29: This is the actual generating step.

docs/crd/release.giantswarm.io_releases.yaml
```diff
@@ -0,0 +1,98 @@
...
+                - name
+                - version
+                type: object
+              type: array
+            date:
+              description: Date that the release became active.
+              format: date-time
+              type: string
+            state:
+              description: 'State indicates the availability of the release: deprecated,
```
tfussell, line 75: Validation was lost here. Needs to be restored using a `kubebuilder` comment.

docs/crd/application.giantswarm.io_appcatalogs.yaml
kopiczko: Can we maybe dockerize release version in some (separate or this one) repository? Maybe docker image already exists? then we can generate stuff with `docker run` and volume bindings.

kopiczko (commented):
Awesome work man! It would be cool if we somehow could use those yaml directly to install them in CPs (I saw what you wrote in the description). Would be good to try to align with https://github.com/giantswarm/opsctl/pull/596.

docs/crd/infrastructure.giantswarm.io_awsmachinedeployments.yaml
```diff
@@ -0,0 +1,137 @@
...
+                instanceDistribution:
+                  properties:
+                    onDemandBaseCapacity:
+                      type: integer
+                    onDemandPercentageAboveBaseCapacity:
+                      type: integer
+                  required:
+                  - onDemandBaseCapacity
+                  - onDemandPercentageAboveBaseCapacity
+                  type: object
```
xh3b4sd, line 81: Here is the defaulting missing now. How can this be generated?

scripts/gen.go
xh3b4sd: When this is how we want to keep it we need some useful comment in here to explain how this works and why this is necessary to be like that.

xh3b4sd (commented):
I think this is a step into the right direction. Really cool. But we need to figure out how to extend the CRDs based on the CR related go types/comments. With this approach it would then maybe make sense to simply download the YAML files in `opsctl` in order to ensure them.

docs/crd/infrastructure.giantswarm.io_awsmachinedeployments.yaml
tfussell: I added this defaulting back for this type. Check it out https://github.com/giantswarm/apiextensions/blob/13a6fe276127e87caceb26cbe44a431e286aa51d/docs/crd/infrastructure.giantswarm.io_awsmachinedeployments.yaml#L74-L82 There's a learning curve, but it feels pretty intuitive once you get going. I can cross-check generated CRDs against existing validation tomorrow to make sure we don't change any semantics accidentally.

scripts/gen.go
tfussell: I agree. I got this pattern from somewhere. I will document it unless we end up changing the method using docker or something.

pkg/apis/provider/v1alpha1/kvm_types.go
```diff
@@ -93,10 +76,10 @@ type KVMConfigSpecKVMK8sKVMDocker struct {
 }
 
 type KVMConfigSpecKVMNode struct {
-	CPUs               int     `json:"cpus" yaml:"cpus"`
-	Disk               float64 `json:"disk" yaml:"disk"`
```
tfussell, line 97: According to https://github.com/kubernetes-sigs/controller-tools/issues/245#issuecomment-518465214, `float64` is not a good type to use in Kubernetes due to round-tripping issues. I'm not sure what the implications of changing this type would be though.

docs/crd/core.giantswarm.io_certconfigs.yaml
```diff
@@ -0,0 +1,97 @@
...
+      required:
+      - metadata
+      - spec
+      type: object
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+status:
```
marians, line 92: I am missing all the property descriptions here from https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/core/v1alpha1/cert_types.go#L36-L100 :-(

marians:
Property descriptions apparently are now coming from Go comments. This has several ramifications:

1. All the descriptions from our YAML schema are missing. For an example, look at CertConfig.

2. There is a mix between Go and JSON/YAML casing, as the description always repeats the Go name of the field as a first word. Example: 

```yaml
  properties:
    configMap:
      description: ConfigMap references a config map containing catalog
        values that should be applied to apps in this catalog.
```

From a UX perspective, that's less than ideal.

3. Also not ideal in terms of readability is that the descriptions are repeating the field name.

I see the benefit of having descriptions only in one place. However, using the Golang one is a trade-off for where UX loses.

marians (changes requested):
As it is this PR means we are loosing descriptions.

xh3b4sd:
Let's not mix docs with internal tooling. In this PR the docs are actually overwritten which we could change. It may not be reasonable to try to solve our documentation needs in one solution together with our internal CRD management. So separating this would probably be better.

kopiczko:
> Let's not mix docs with internal tooling. In this PR the docs are actually overwritten which we could change. It may not be reasonable to try to solve our documentation needs in one solution together with our internal CRD management. So separating this would probably be better.

I don't get that. Can you elaborate here? Or maybe tell what you are referring to?

kopiczko:
> I see the benefit of having descriptions only in one place. However, using the Golang one is a trade-off for where UX loses.

Is the lost bigger than the benefit in your opinion? I'd say people who read those are kind of used to Go style comments. For me not copying stuff over and having simpler workflows by far outweighs a problem of description starting with a repeated name.

xh3b4sd:
Marian and Thomas started to generate CRDs for documentation reasons and put them in the `docs/` folder. This is what is used to display in https://docs.giantswarm.io/reference/cp-k8s-api. Thomas now wanted to generate the CRDs we use with our tooling instead of having these multi line string definitions in go variables. His result as per this pull request is written to the `docs/` folder, which now mixes two different purposes. I was questioning if it is good to mix these two use cases which both have likely totally different requirements. So maybe we want to thrive for one approach to rule them all but maybe this is more complicated than necessary. Just wanted to nudge on that a bit.

xh3b4sd:
> For me not copying stuff over and having simpler workflows by far outweighs a problem of description starting with a repeated name.

Which is you, the developer speaking. Marian thinks about the users. See my comment about mixing use cases above.

marians:
@xh3b4sd Everything in the `docs/` folder is generated automatically. The CRD content with all the schema property `description` texts is contained in YAML strings in Go code. [Example](https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/core/v1alpha1/cert_types.go#L36-L100)

The problem is not that we overwrite `docs/`. The problem is that we

- have some descriptions in there coming from Go code. This means the outcome is a little strange for end users (but probably still usable).
- lose descriptions which are currently present only in YAML. This means we have to at least make sure we transfer the descriptions from the YAML strings to Go comments before re-generating the `docs` content.

marians:
@kopiczko 
> Is the lost bigger than the benefit in your opinion?

That's hard to judge and to compare, mainly as the gains and losses happen on different ends (internal vs. external).

I think most users will be able to cope with the strangeness and still be able to make the right conclusions. If this helps us getting more complete, up-to-date and accurate property docs, fine. I'm not in the way.

tfussell:
Thanks for the review and comments. I'll try to address each concern.

docs/crd/core.giantswarm.io_certconfigs.yaml
tfussell: Sorry, didn't want to erase your hard work. I will copy and paste them into CR field comments so they make it back into the CRD before I merge. I didn't want to invest the time initially if it seemed like a bad approach in general.

tfussell:
> > For me not copying stuff over and having simpler workflows by far outweighs a problem of description starting with a repeated name.
> 
> Which is you, the developer speaking. Marian thinks about the users. See my comment about mixing use cases above.

I think we have the opportunity to address both use cases with this change. Here's how:
- As you know, with the old method, we had multiline strings in source code containing manually defined CRDs. These CRDs were unmarshalled into `runtime.Object`s and then marshalled back into standalone YAML files in the `docs/crd` directory using bespoke tooling. Aligning with upstream Kubernetes means our libraries are more easily understood by internal developers and external technical users.
- The old method required us to write CRDs by hand without syntax highlighting or validation because they were YAML documents in Go source which provided many opportunities for mistakes or outdated schemata/documentation. By generating CRDs from CR structs, we know that our CRDs are valid Kubernetes objects and that they are in sync with CRs.
- We now have a CRD with a schema [for every type](https://github.com/giantswarm/apiextensions/tree/generate-crds/docs/crd) "for free" which means more documentation and validation for computers and humans even if it's computer generated. It also gives us the ability to move to `v1` `CustomResourceDefinition` which does not allow disabling pruning.
- When we add GoDoc comments to our CR fields, we get them in the CRD field descriptions automatically. This means improving DX (developer experience) also improves UX and vice versa.

docs/crd/infrastructure.giantswarm.io_awsmachinedeployments.yaml
```diff
@@ -0,0 +1,142 @@
...
+                  properties:
+                    onDemandBaseCapacity:
+                      default: 0
+                      minimum: 0
+                      type: integer
+                    onDemandPercentageAboveBaseCapacity:
+                      default: 100
+                      maximum: 100
+                      minimum: 0
+                      type: integer
```
xh3b4sd, line 82: In `master` there is also `format: int32`. I am not sure how relevant this is. Just found the difference.

docs/crd/infrastructure.giantswarm.io_awsmachinedeployments.yaml
```diff
@@ -0,0 +1,142 @@
...
+          type: object
+      required:
+      - spec
+      - status
+      type: object
+  version: v1alpha2
+  versions:
+  - name: v1alpha2
+    served: true
+    storage: true
```
xh3b4sd, line 136: With this now how does it work to add a new version? I thought the whole schema should be in that list but instead I see that the schema is defined at the top level `spec.validation`. I would actually like to put everything into `spec.versions` since I understood this is how to define multiple ones. It is further unclear to me how the CR has to be defined to make that change effective in the generated CRD.

xh3b4sd (commented):
This is cool. We discussed a little. Here are some final thoughts. 

* When the go types changed, but the CRDs are not re-generated, CI should fail in an idea world. Not a big deal in the first step, but let's consider this for the upcoming PRs. 
* I already made the experience that the files are hard to determine in the Github file tree UI because the generated file names are super long. This makes it super hard for humans to identify where the file you are looking for is. Maybe we could work more with a directory structure. Just an idea. 
* It is amazing that now all CR status structures are also defined and that the whole schemas are actually structural and valid. This makes so much sense. 
* We should test this somehow on a Control Plane to see if the CRDs are still fine. We can maybe talk about it and try something out together. It is a bit dangerous to deal with so we should make sure this does not break production environments.

docs/crd/infrastructure.giantswarm.io_awsmachinedeployments.yaml
tfussell: I've done some research on this, and this is an important difference between `v1beta1.CustomResourceDefinition` and `v1.CustomResourceDefinition`. In `v1beta1`, you must either:
- Define validation at the top level `.spec.validation`.
- OR Have multiple versions with different `.schema` fields.
When we ran into this a few months ago, we got around it by defining a dummy unused version which had a null schema thereby satisfying the second requirement. This generator takes the first approach and defines a top level validation.

In `v1`, the top level fields are removed so it's naturally allowed to have a single version with a schema.

These issues might not become apparent until we actually try to create the CRDs on a running Kubernetes API so it's important that we actually create them during testing I think.

xh3b4sd: Then let's already graduate the `AWSMachineDeployment` CRD to `apiextensions/v1` and test that.

docs/crd/application.giantswarm.io_appcatalogs.yaml
tfussell: This would be nice, but I tend to worry about the added complexity of keeping another image updated. What do you think about the `cluster-api` approach of installing the tool before build? https://github.com/kubernetes-sigs/cluster-api/blob/master/Makefile#L164-L165 In this case it's using a `Makefile`, but I guess we can just add it to `scripts/gen.sh`.

docs/crd/infrastructure.giantswarm.io_awsmachinedeployments.yaml
tfussell: I've made some significant improvements to the scripts directory and upgraded all CRDs to `v1`. Feel free to test them.

xh3b4sd: We only need to graduate `AWSMachineDeployment` for now. At least that was what we agreed on with Firecracker, Marcel and Timo when we talked earlier. Reason was to keep the friction down since we did not know what breaks when we roll out all at once the new way. So unless you are super confident this is the right move we should only do it with `AWSMachineDeployment` because this is necessary due to our wish to use defaulting. All others are not so super important and we can sort them in further iterations.

kopiczko:
Another thought which I don't know why I didn't propose in the first place. I'd be fine with breaking godoc conventions for there files to have CRDs properly generated. We can add `// nolint:godoc` (don't know if this is godoc or something else) for those files and treat them exceptionally.

docs/crd/infrastructure.giantswarm.io_awsmachinedeployments.yaml
tfussell: Would it be alright to do all of `infrastructure.giantswarm.io`? There's not an easy way to select individual types, but I can select a package with the upstream tools.

tfussell: All of the other groups have been reverted to `v1beta1`.

xh3b4sd: Works for me when we automate the migration and test that. Should be fine since the proper schemas are generated.

docs/crd/core.giantswarm.io_certconfigs.yaml
marians: Sure, I understand.

tfussell:
> Another thought which I don't know why I didn't propose in the first place. I'd be fine with breaking godoc conventions for there files to have CRDs properly generated. We can add `// nolint:godoc` (don't know if this is godoc or something else) for those files and treat them exceptionally.

I checked cluster-api and it uses normal GoDoc style comments for CRs (field name in CamelCase at the beginning of each one). Both ways seem reasonable to me.

marians:
Anyway, I'd leave the comment changes out of this PR. We can tackle that later.

docs/crd/application.giantswarm.io_appcatalogs.yaml
tfussell: This is improved in the following PR (#410) so I'm not too worried about this being perfect.

scripts/gen.go
tfussell: Resolved in #410

tfussell:
This PR isn't perfect, but most of the issues are resolved in the next PR, #410. If it's okay with everyone, I would merge this to master after some approvals and tie up any loose ends in the next PR.

tfussell:
The only remaining task for this PR is figuring out what to do about the `float64` fields. I will work on that tomorrow. Open to ideas.

xh3b4sd:
I did not fully understand the problem and we can talk about it in more detail if you like. I just found `kubebuilder:validation:Type` which may help to enforce the right type?

pkg/apis/core/v1alpha1/kvm_cluster_types.go
```diff
@@ -73,11 +59,14 @@ type KVMClusterConfigSpecGuestWorker struct {
...
 
+// TODO: change MemorySizeGB and StorageSizeGB to resource.Quantity
 type KVMClusterConfigSpecGuestNode struct {
-	ID            string  `json:"id" yaml:"id"`
-	CPUCores      int     `json:"cpuCores,omitempty" yaml:"cpuCores,omitempty"`
-	MemorySizeGB  float64 `json:"memorySizeGB,omitempty" yaml:"memorySizeGB,omitempty"`
-	StorageSizeGB float64 `json:"storageSizeGB,omitempty" yaml:"storageSizeGB,omitempty"`
+	ID       string `json:"id" yaml:"id"`
+	CPUCores int    `json:"cpuCores,omitempty" yaml:"cpuCores,omitempty"`
+	// +kubebuilder:validation:Type=number
```
tfussell, line 66: Specifying the type here for validation. This will still be a breaking change for the `kvm-operator` but should be able to work around it with `strconv.ParseFloat(MemorySizeGB, 64)`. I'll test in `kvm-operator` now.

tfussell:
Now that `float64` has been fixed, this PR has been effectively subsumed by https://github.com/giantswarm/apiextensions/pull/410 so I'm going to merge it and continue work in the next PR with an announcement that `apiextensions@master` is a WIP.

xh3b4sd:
Closing in favour of https://github.com/giantswarm/apiextensions/pull/410.

PR #410, opened 2020-04-14
Improve CRD generation

tfussell:
As a follow-up from https://github.com/giantswarm/apiextensions/pull/409, this PR allows CRDs to be loaded in go code as structs from the generated YAMLs. Most of the interesting changes are in the scripts directory.

## Checklist


pkg/crd/crd.go
```diff
@@ -0,0 +1,104 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/giantswarm/microerror"
+	"github.com/markbates/pkger"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"sigs.k8s.io/yaml"
+
+	_ "github.com/giantswarm/apiextensions"
```
xh3b4sd, line 15: For my interest, because I cannot figure it out. Why is this imported here?

.gitignore
```diff
@@ -1 +1,2 @@
 scripts/bin
+.DS_Store
```
kopiczko, line 2: Are you aware of global gitignore? (core.excludesfile)

xh3b4sd (commented):
This looks good, I am only a bit confused about the generation results I get now locally when I run `./scripts/gen.sh`. The virtual file system changed and I am not sure why. Uncertain which is now the desired state. Running tests after the generation doesn't work because the files are put into the wrong place as it seems. 

```
$ git status
On branch improve-generation
Your branch is up-to-date with 'origin/improve-generation'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   pkged.go

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	github.com/

no changes added to commit (use "git add" and/or "git commit -a")
```

.gitignore
tfussell: Yes, just an old habit. Fixed.

pkg/crd/crd.go
```diff
@@ -0,0 +1,142 @@
...
+
+	_ "github.com/giantswarm/apiextensions" // Load pkger generated filesystem for side-effects
+)
+
+const (
+	crdDirectory = "/config/crd/bases"
+)
+
+var (
+	// GVK of CustomResourceDefinition in apiextensions.k8s.io/v1beta1
```
MarcelMue, line 26: Nit: Can we avoid `GVK` as an abbreviation? It's fairly easy to get in this context but we don't abbreviate CRD here either :D

pkg/crd/crd.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, microerror.Mask(err)
+	}
+	if found == nil {
+		return nil, microerror.Mask(notFoundError)
+	}
+
+	return found, nil
+}
+
+// LoadV1Beta1 loads a v1beta1 CRD from the filesystem
```
MarcelMue, line 118:
```suggestion
// LoadV1Beta1 loads a v1beta1 CRD from the filesystem.
```
Nit to end with `.` - just did it here.

MarcelMue (approved):
Fine with me if tested and other are happy.

pkg/crd/crd.go
```diff
@@ -0,0 +1,142 @@
...
+		panic(microerror.JSON(microerror.Mask(conversionFailedError)))
+	}
+	return crd
+}
+
+// LoadV1Beta1 loads a v1 CRD from the filesystem
+func LoadV1(group, kind string) (out *v1.CustomResourceDefinition) {
+	found, err := Find(group, kind)
+	if err != nil {
+		panic(microerror.JSON(microerror.Mask(conversionFailedError)))
```
kopiczko, line 135: `panic` prints the line so no need to Mask it here.

scripts/gen.sh
```diff
@@ -9,12 +9,17 @@ export GOROOT=$(go env GOROOT)
...
 mkdir -p "$dir/bin"
 cd "$dir"
 for tool in $tools; do
-  go build -o "$dir/bin/$(basename "$tool")" "$tool"
+  base=$(echo "$tool" | sed "s/\/cmd\/.*//")
+  version=$(go list -m -f '{{.Version}}' "$base")
+  echo "Rebuilding $tool@$version"
+  go build -o "$dir/bin" "$tool"
 done
 cd ..
```
kopiczko, line 24: I'd propose having a function instead of using iteration over string.

```
install_tool() {
  local tool=$1
  go build -o "$dir/bin/$(basename "$tool")" "$tool"
  base=$(echo "$tool" | sed "s/\/cmd\/.*//")
  version=$(go list -m -f '{{.Version}}' "$base")
  echo "Rebuilding $tool@$version"
  go build -o "$dir/bin" "$tool"
}

install_tool "sigs.k8s.io/controller-tools/cmd/controller-gen"
install_tool ...
```

scripts/gen.sh
```diff
@@ -49,14 +55,26 @@ cd "$dir/.." || exit
...
+  output:dir=config/crd/bases \
   crd:crdVersions=v1
+
+echo "Kustomizing CRDs"
+"$dir/bin/kustomize" build \
+  config/crd \
+  -o config/crd/bases/release.giantswarm.io_releases.yaml
+
+echo "Using pkger to package CRDs into go source virtual file system"
+"$dir/bin/pkger"
```
kopiczko, line 80: Now it would be really good to call `gen.sh` from the CI and assert that `git diff` doesn't print anything.

scripts/gen.sh
```diff
@@ -9,12 +9,17 @@ export GOROOT=$(go env GOROOT)
...
-golang.org/x/tools/cmd/goimports"
+golang.org/x/tools/cmd/goimports
+github.com/markbates/pkger/cmd/pkger
+sigs.k8s.io/kustomize/kustomize/v3"
 
 mkdir -p "$dir/bin"
 cd "$dir"
 for tool in $tools; do
-  go build -o "$dir/bin/$(basename "$tool")" "$tool"
+  base=$(echo "$tool" | sed "s/\/cmd\/.*//")
```
kopiczko, line 19:
```
$ basename "sigs.k8s.io/controller-tools/cmd/controller-gen"
controller-gen
```

pkg/crd/crd.go
tfussell: For better or worse, `pkger` uses global state to load the virtual filesystem. This was imported for the side effect of loading the data into global state. I found a better way to do it so this is no longer necessary.

config/crd/patches/releases.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: releases.release.giantswarm.io
+spec:
+  validation:
+    openAPIV3Schema:
+      properties:
+        metadata:
+          properties:
+            name:
+              type: string
+              pattern: ^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$
```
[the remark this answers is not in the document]
tfussell: I wish. Kubernetes allows validation on `.metadata.name` and `.metadata.namespace`, but this isn't supported with `kubebuilder` so I came up with this hack. I think `kustomize` to modify CRDs will be useful in the future, but it's also a learning curve. The alternative would be to drop validation of Release CR names and enforce it other ways.

pkg/crd/crd.go
tfussell: Thanks, fairly new to panics.

scripts/gen.sh
tfussell: Looks nicer--I'll do it.

scripts/gen.sh
tfussell: I've done this now. I modified a comment on a CR field temporarily and ensured that it failed CI https://circleci.com/gh/giantswarm/apiextensions/2020

scripts/gen.sh
tfussell: Not sure if I understand.

kopiczko: Sorry I should have looked better. I thought this will leave the segment after `/cmd/` and not everything before.

scripts/gen.sh
```diff
@@ -49,14 +63,33 @@ cd "$dir/.." || exit
...
+# Ensure that we have fresh CRDs in case any have been deleted
+rm -rf "$dir/../config/crd/bases"
+
+# Using kubebuilder comments, create new CRDs from CR definitions in source files
+echo "Generating all CRDs as v1beta1"
 "$dir/bin/controller-gen" \
   crd \
   paths=./pkg/apis/... \
-  output:dir=docs/crd \
+  output:dir=config/crd/bases \
```
kopiczko, line 74: Here I come with a nitpick. Is `bases` a convention? Otherwise I'm biased towards using singular forms (`base` in this case).

apiextensions.go
```diff
@@ -0,0 +1,4 @@
+package apiextensions
+
+// This root package is currently required for pkger to read child directories.
+// Please do not remove.
```
kopiczko, line 4: Is it outdated?

config/crd/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+patchesStrategicMerge:
+- patches/releases.yaml
+resources:
+- bases/release.giantswarm.io_releases.yaml
```
kopiczko, line 6: How does it work when we have more than one Kustomization? Do we need separate configuration for each one? If so then this file name is probably too generic.

xh3b4sd:
I pushed a couple of commits to fix a couple of things. I got the `AWSMachineDeployment` CRD graduated on `ginger` now but there are a couple of things left. 

* The code generation still keeps some code behind. There is a `github.com/` directory that should not be there. That needs fixing. Uncertain where it is coming from. 
* We need to make all statuses optional by adding `// +kubebuilder:validation:Optional` to all of them. I did that for the `AWSMachineDeployment` already. Without it the CR cannot be created and you cannot create CRs with the status. I can push a commit to fix that for all of them. 
* As a general note. When new fields are added to a CR they have to be defaulted or made optional in the first place. Otherwise the migration does not work at all.

kopiczko:
> * As a general note. When new fields are added to a CR they have to be defaulted or made optional in the first place. Otherwise the migration does not work at all.

Maybe we should put also that in the PR template.

pkg/crd/crd_test.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func Test_LoadAll(t *testing.T) {
+	groupKinds := map[string][]string{
+		"application.giantswarm.io": {
```
xh3b4sd, line 69: According to the implemented file system walk in `pkg/crd/crd.go` the functionality implemented in the `Find` function could be adapted to have an `All` function which then provides a list of generated asstes/CRDs. Then we do not need to remind humans to add their type to the test once added. But this is not mission critical now and can be improved in a separate step later.

docs/pull_request_template.md
```diff
@@ -2,3 +2,4 @@
 
 - [ ] Consider SIG UX feedback.
 - [ ] Update changelog in CHANGELOG.md.
+- [ ] If adding a new type, ensure that it is added to the [CRD unit tests](https://github.com/giantswarm/apiextensions/blob/d78aba91d578d56ef49f58e336035d35edc4e1b0/pkg/crd/crd_test.go#L9).
```
xh3b4sd, line 5: Not a fan of that but left a comment.

xh3b4sd (approved):
I think we are good to go. A bit quick and dirty but the result and level of improvement is amazing. We can further iterate here.

xh3b4sd:
> Maybe we should put also that in the PR template.

I would not like that. We should only add something to the template that provides real value to most of the PRs. I feel like adding a new field happens once a quarter. Within 3 months there happen dozens of PRs which do not benefit and people start to ignore the advices of PR templates. I think we should be careful with that. I recently added the note about considering UX feedback and I am not so sure anymore this was a good move. I am thinking to remove it again due to the same reasons.

pkg/apis/core/v1alpha1/cert_types.go
```diff
@@ -12,106 +13,8 @@ const (
...
-                  ttl:
-                    description: |
-                      Expiry duration after creation. The value must consist of a number
-                      combined with a unit, without blanks, to be parsed by the Go
-                      [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration) function.
-                    type: string
-              versionBundle:
-                description: |
-                  No longer used
-                type: object
```
marians, line 100: I still don't see where these descriptions are going. The according go fields don't have any comments, so the only way to recover this information will be to dig in the git history of the repo if we remove this now. (This CRD is only one example out of several affected.)

pkg/crd/crd_test.go
kopiczko: Assuming you trust `pkger`.

kopiczko:
> I recently added the note about considering UX feedback and I am not so sure anymore this was a good move.

I like that. But surely we shouldn't write books in those templates.

xh3b4sd:
We took this over. After testing and gaining confidence we merge this to master now to get a workable tag in other projects. Thanks a lot to everyone especially @tfussell who did some invaluable contributions for the sake of the whole team. It was a bit bumpy but I hope we are all kind of cool with it. We rollin'.

marians:
I'm wondering why there are still two CRD YAML files in `docs/crd`:

image

Why have those not moved to `config/bases/crd`?

marians:
When we work on bringing back descriptions, and we use Go comments for them, can we agree that we don't have to repeat the name of the property in the comment? We already have this here with the `LastTransitionTime` property:

image

It doesn't seem to trigger any lint/fmt errors in my tooling (Go 1.14).

config/crd/bases/release.giantswarm.io_releases.yaml
```diff
@@ -16,56 +18,50 @@ spec:
...
   names:
     kind: Release
+    listKind: ReleaseList
     plural: releases
-    shortNames:
-    - rel
     singular: release
-  preserveUnknownFields: false
-  scope: Cluster
+  scope: Namespaced
```
ubergesundheit, line 28: Why did this change from cluster to namespace scoped? Just curious..

xh3b4sd: Uncertain, but I think @tfussell knows more about it.

tfussell: We need to add `// +kubebuilder:resource:scope=Cluster` to CRs to get the back to parity with our old manually defined CRDs. I can do some if it, but the work will hopefully be divided into teams responsible for each CRD.

PR #412, opened 2020-04-16
Add CNI attributes

paurosello:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
-
 ## [Unreleased]
+
+## [0.2.6] - 2020-04-15
+
+### Added
+
+- Document G8sControlPlane CRD [#405](https://github.com/giantswarm/apiextensions/pull/405)
+- Document Chart CRD [#406](https://github.com/giantswarm/apiextensions/pull/406)
+
+## [0.2.5] - 2020-04-09
+
+### Changed
+
+- Fix a problem in the MachineDeployment CRD YAML file [#404](https://github.com/giantswarm/apiextensions/pull/404)
+
+## [0.2.4] - 2020-04-08
+
+### Changed
+
+- Fix path of CR and CRD yaml files for Cluster and MachineDeployment [#403](https://github.com/giantswarm/apiextensions/pull/403)
+- Add schema documentation for CertConfig [#401](https://github.com/giantswarm/apiextensions/pull/401)
 
 ## [0.2.3] - 2020-04-08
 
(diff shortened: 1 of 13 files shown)


pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
```diff
@@ -78,6 +78,22 @@ spec:
                   AWS-specific configuration details.
                 type: object
                 properties:
+                  cni:
```
teemow, line 81: Just as a heads-up. With Cluster API this value gets defined in the provider independent ClusterSpec. Since we aren't able to handle this structure in the aws-operator at the moment we decided to keep it here for now.

See: https://github.com/kubernetes-sigs/cluster-api/blob/master/api/v1alpha3/cluster_types.go#L78

pkg/apis/application/v1alpha1/chart_types_test.go
```diff
@@ -1,10 +1,96 @@
...
+)
 
 func Test_NewChartCRD(t *testing.T) {
 	crd := NewChartCRD()
 	if crd == nil {
 		t.Error("Chart CRD was nil.")
 	}
 }
+
+func Test_GenerateChartYAML(t *testing.T) {
```
kopiczko, line 23: @tfussell doesn't that collide with your work on generating CRDs?

pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
```diff
@@ -78,6 +78,22 @@ spec:
...
+                    description: |
+                      Pod networking configuration
```
kopiczko, line 83:
```suggestion
                      Pod networking configuration.
```

pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
kopiczko: Why we don't use the same structure then?

pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
paurosello: thanks!

pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
paurosello: I changed the field to match upstream, but not as an array, I think that would cause more confusion...

paurosello:
Closing in favor of https://github.com/giantswarm/apiextensions/pull/413

CHANGELOG.md
```diff
@@ -6,16 +6,36 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
-
 ## [Unreleased]
 
-## [0.2.3] - 2020-04-08
+- Add pod networking parameters
```
kopiczko, line 11: I'd like those changelogs to be more meaningful. The biggest value for me (as a dev) is to see the migration path from one version to another. What those changes are doing can be figured from the docs.

```suggestion
### Added

- Add add `.spec.pods` field to `awsconfigs.infrastructure.giantswarm.io` CR.
```

PR #413, opened 2020-04-16
Add pods fields to AWSCluster

paurosello:
## Checklist


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add pod field to AWSCluster.
```
marians, line 13:
```suggestion
- Add `.spec.cluster.pods` field to AWSCluster.
```

config/crd/bases/infrastructure.giantswarm.io_awsclusters.yaml
```diff
@@ -107,6 +107,16 @@ spec:
                     - availabilityZone
                     - instanceType
                     type: object
+                  pods:
+                    description: Pod network configuration.
+                    properties:
+                      cidrBlock:
+                        description: Subnet size, expresses as the count of leading
+                          1 bits in the subnet mask of this subnet.
```
marians, line 115: If the property is expected to carry a string like `"1.2.3.4/8"` then I think this description does not explain it well. I would expect something like

"IPv4 address block used for pods, in CIDR notation."

Also we might want to provide a validation regex.

marians:
@teemow That's a way to make me feel "heard" ;-)

PR #415, opened 2020-04-16
Make code generation idempotent

tfussell:
We should only regenerate the virtual filesystem when the CRD YAMLs have changed. To allow this in a way compatible with Linux and macOS, I calculate a hash of each file in `config/crd/bases`, hash that list of files and hashes, and save the result to a file (`scripts/.hash`). This value will change if any file in the directory has changed indicating that the virtual filesystem should be regenerated. This reduces clutter in the working directory and the cognitive load of deciding when to commit the modified `pkged.go` file.


scripts/gen.sh
```diff
@@ -94,8 +94,14 @@ echo "Kustomizing CRDs"
   -o config/crd/bases/release.giantswarm.io_releases.yaml
 
 # Package CRD YAMLs into a virtual filesystem so they can be accessed by New*CRD() functions
-echo "Using pkger to package CRDs into go source virtual file system"
-"$toolpath/pkger" -include /config/crd/bases -o pkg/crd
+hash=$(find config/crd/bases -type f -print0 | xargs -0 sha1sum | sort -df | sha1sum)
+prevhash=$(cat "$dir/.hash" 2> /dev/null || echo "")
+if [ "$hash" != "$prevhash" ]; then
+  echo "Detected changes in CRD YAMLs"
+  echo "Using pkger to package CRDs into go source virtual file system"
```
MarcelMue, line 101: Should probably add something in the readme about how this works etc

tfussell: Done!

PR #417, opened 2020-04-17
Add AWSMachineDeployment docs again

marians:

## Checklist


CHANGELOG.md
```diff
@@ -19,17 +20,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-
 ## [0.2.6] - 2020-04-15
 
 ### Added
 
 - Document G8sControlPlane CRD [#405](https://github.com/giantswarm/apiextensions/pull/405)
 - Document Chart CRD [#406](https://github.com/giantswarm/apiextensions/pull/406)
 
-
-
```
xh3b4sd, line 32: When you change the padding here you should change it in the whole file but I would rather ask you to keep it as it is. For me as a human it is easier to read with more padding so I can more quickly digest where changelogs for a specific version start and end.

pkg/apis/infrastructure/v1alpha2/aws_machine_deployment_types.go
```diff
@@ -38,71 +38,103 @@ func NewAWSMachineDeploymentCR() *AWSMachineDeployment {
...
+// It contains provider-specific specification and status for a node pool.
+// In use on AWS since Giant Swarm release v10.x.x and reconciled by aws-operator.
 type AWSMachineDeployment struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec              AWSMachineDeploymentSpec `json:"spec" yaml:"spec"`
+	// Contains the specification.
+	Spec AWSMachineDeploymentSpec `json:"spec" yaml:"spec"`
+	// Holds status information.
 	// +kubebuilder:validation:Optional
```
xh3b4sd, line 50: I would like to do it the other way around. The comment close to the type. Above the comment the generator tags.

CHANGELOG.md
marians: Hmm. markdownlint complains in my case, because several empty lines are no different in terms of output from just one empty line. But I would value your feelings higher than those of markdownlint ;-)

pkg/apis/infrastructure/v1alpha2/aws_machine_deployment_types.go
marians: Changed for all fields

PR #419, opened 2020-04-17
drop docs crds

xh3b4sd:
This is based on https://github.com/giantswarm/apiextensions/pull/418. We generate the CRDs in `config/crd/bases/` now. The golden file tests for the CRs do not have to contain the CRD generation anymore AFAIK.


marians:
Why base this on g8scontrolplane-docs instead of master?

xh3b4sd:
I saw your PR and wanted to suggest the changes here because it is not necessary to generate the CRDs in the `docs/` folder anymore.

PR #420, opened 2020-04-17
Add cluster-api types

tfussell:
## Checklist


scripts/go.sum
```diff
@@ -1,649 +0,0 @@
-cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
tfussell, line 1: I have realized that there's no reason to keep a `go.sum` in the script directory. It will change as each tool is installed and when files are generated. We only use `go.mod` to keep track of direct dependencies so I think this is okay.

scripts/go.mod
```diff
@@ -5,6 +5,9 @@ go 1.14
 require (
 	github.com/markbates/pkger v0.15.1
 	k8s.io/code-generator v0.16.6
+	sigs.k8s.io/cluster-api v0.2.10
```
tfussell, line 8: Same version as was previously in the root `go.mod`.

pkg/apis/infrastructure/v1alpha2/g8s_control_plane_types.go
```diff
@@ -32,7 +32,7 @@ func NewG8sControlPlaneCR() *G8sControlPlane {
 				crDocsAnnotation: g8sControlPlaneDocumentationLink,
 			},
 		},
-		TypeMeta: NewClusterTypeMeta(),
+		TypeMeta: NewG8sControlPlaneTypeMeta(),
```
tfussell, line 35: I think this was a bug. Noticed while fixing other things.

scripts/gen.sh
```diff
@@ -86,6 +90,14 @@ echo "Generating infrastructure.giantswarm.io CRDs as v1"
   output:dir=config/crd/bases \
   crd:crdVersions=v1
 
+pushd "$dir" > /dev/null
+"$toolpath/controller-gen" \
+  crd \
+  paths=sigs.k8s.io/cluster-api/api/v1alpha2 \
+  output:dir=../config/crd/bases \
+  crd:crdVersions=v1beta1
```
tfussell, line 98: I'm not sure if this should be `v1` or `v1beta1`. What do you think @xh3b4sd?

tfussell: These are the CAPI CRDs.

scripts/gen.sh
```diff
@@ -14,34 +21,38 @@ install_tool() {
...
   local bin=$3
-  pushd "$dir" > /dev/null
+  rm go.sum 2> /dev/null || true
   version=$(go list -m -f '{{.Version}}' "$module")
   echo "Rebuilding $bin@$version"
   mkdir -p "$toolpath"
-  go build -o "$toolpath/$bin" "$module""$package"
-  popd > /dev/null
+  go build -o "$toolpath/$bin" "$module""$package" 2> /dev/null
+  git checkout go.mod 2> /dev/null
```
tfussell, line 29: There's no way to prevent `go` from changing `go.mod`, so I'm just reseting it with `git` after each `go build`.

scripts/gen.sh
xh3b4sd: For now this should stay like that. We just need to know how to make it `v1` once we want to graduate them as well. The only CRDs we graduated so far where AWS related.

xh3b4sd: So for my understanding how does this work now? How does the generator know where to find the source of truth for CAPI?

tfussell: `controller-gen` uses standard go modules logic to resolve external dependencies. I defined `cluster-api@0.2.10` in `scripts/go.mod` so as long as this command is executed in the `scripts/` directory, it generates CAPI 0.2.10 v1alpha2 CRDs. I tested this by clearing my module cache, changing the version in `go.mod` to 0.2.9 and regenerating the CRDs. Then I found `~/go/pkg/mod/sigs.k8s.io/controller-tools@v0.2.9/` had been downloaded. When we're ready for CAPI v1alpha3, it should be sufficient to change this version to `v0.3.3` and `paths` in the above command to `sigs.k8s.io/cluster-api/api/v1alpha3`.

xh3b4sd: Can you put this comment as a code comment in line 99 ? I think this would be useful.

xh3b4sd (approved):

scripts/gen.sh
kopiczko: I wonder if we need this go.mod vile at all. What is wrong with keeping binary versions in this file?

tfussell: `cluster-api` is for `controller-gen`, but if you have a better way to install these other binaries in a platform-independent way without `go.mod` that would be cool. I couldn't find compiled releases of any of these. I even tried with `go install` setting `GO111MODULE=off` and changing `GOBIN`, but none of these worked well.

scripts/gen.sh
tfussell: Sure thing.

marians:
We are bringing in CRDs `Machine` and `MachineSet` here, too. Are these relevant to us? As it is, they would be published in our CRD docs.

pkg/crd/crd_test.go
```diff
@@ -106,6 +106,12 @@ func Test_LoadAll(t *testing.T) {
 			"Release",
 			"ReleaseCycle",
 		},
+		"cluster.x-k8s.io": {
+			"Cluster",
+			"MachineDeployment",
+			"Machine",
+			"MachineSet",
```
marians, line 113: Commented this elsewhere. Do we use `Machine` and `MachineSet`?

xh3b4sd: We do not use them yet.

marians: Then I'd like to see them removed here.

ubergesundheit (approved):

pkg/crd/crd_test.go
tfussell: Good call. I've removed them.

PR #398, opened 2020-04-08
Add more instructions on publishing and regarding generated files

marians:
Improvements to the PR template

PR #421, opened 2020-04-18
changelog: small improvements

kopiczko:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,19 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+### Added
 
-- Add descriptions to AWSMachineDeployment CRD properties.
+- Add properties descriptions to AWSMachineDeployment in infrastructure.giantswarm.io/v1alpha2.
 
 
 
 ## [0.3.0] - 2020-04-16
 
+### Added
+
+- Add `.spec.provider.pods` field to AWSCluster in core.giantswarm.io/v1alpha1.
+
 ### Changed
 
-- Add pod field to AWSCluster.
 - Replace custom `time.Time` wrapper `DeepCopyTime` with Kubernetes built-in `metav1.Time`.
-- Update `architect-orb` to `v0.8.8`.
 - Generate CRDs via `kubebuilder` tools based on CRs.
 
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -7,19 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added
+
+- Add `.spec.provider.pods` field to AWSCluster in core.giantswarm.io/v1alpha1.
+
 ### Changed
 
-- Add pod field to AWSCluster.
 - Replace custom `time.Time` wrapper `DeepCopyTime` with Kubernetes built-in `metav1.Time`.
-- Update `architect-orb` to `v0.8.8`.
```
kopiczko, line 22: I think lines like this should not appear in the changelog. They are not relevant from the user perspective. For me the most important part of the changelog is knowing the migration path.

xh3b4sd: I often have a hard time to decide what to put in the changelog and what not. What is the rule of thumb? Everything relevant to the user? Who is the user? What means relevant?

kopiczko: As a user of the library I should know:

- Why should I upgrade the library, e.g.: bug fix, security fix or a new feature (Updating architect-orb is not really relevant.)
- What I need to do to migrate to a newer breaking version.

I'm not clear if we should put docs entries here yet. First I'd need to understand how those generated CRDs in `/docs` are used. I'd appreciate some quick pointer here.

I'd like to to have some common understanding / common sense why we have those changelogs  because otherwise they're just some extra chore you have to deal with but I'm not sure how to achieve that. One thought I have is writing a document which would be linked behind `- [ ] Update changelog in CHANGELOG.md` checklist item. Maybe some session interested SIG UX (changelogs are customer facing) & SIG operator (changelogs should be helpful when doing migrations) members would make sense.

kopiczko: Okay [reading this](https://github.com/giantswarm/apiextensions/pull/398/files#diff-ac354b58e960ffc4b699664e7f8a832dR14-R19) I think we should keep changelogs for docs because they are relevant from the user's perspective.

I should add that by the user I always mean the user of the library.

tfussell: I don't mind dropping it. I usually open a PR, see `Update changelog in CHANGELOG.md.` in the template, then go back and add my change to the changelog mostly regardless of what it was. Maybe we could change the wording to `Mention relevant changes, if any, in CHANGELOG.md.` We also have to make this rule work for non-library repos like operators in which case it's not as clear who the users are.

kopiczko: For me operator is similar to library with the difference that the contract is not Go API but the CR it reconciles. So again how we deploy it or if we added more tests — it doesn’t matter for the end user. What matters is behaviour and interface changes.

kopiczko:
I have my approvals so merging it now to establish  some direction. I’ll try to provide some document on how to write the changelog and why doing it like that. Then link it. I don’t want this thing to be annoying but helpful.

CHANGELOG.md
xh3b4sd: Should we have brief docs about this for Chapter PE or just something in `fmt`?

PR #422, opened 2020-04-20
Make releases CR cluster scoped

MarcelMue:
## Checklist


kopiczko:
Can you share the reasoning why that is?

MarcelMue:
I think @tfussell can explain it better but I can try:
We changed generation of our CRDs to be done fully in code - this causes all the yaml definitions to be generated from our code. 
The `kubebuilder` annotations are necessary to have everything in the `yaml` stuff - the other `gen` is there for the golang code gen.

tfussell:
Yep. This doesn't change anything (as long as `apiextensions` 0.3.0+ hasn't been used in an operator yet), just fixes a regression from the new CRD generation method. Releases were always supposed to be cluster-scoped.

kopiczko:
> Releases were always supposed to be cluster-scoped.

That was the sentence I was looking for :D Thanks!

PR #423, opened 2020-04-20
AWSControlPlane: add descriptions, remove unused status

marians:

This adds descriptions to properties.

Also it removes the unused status part.

## Checklist


marians:
There are still references to the removed `AWSControlPlaneStatus` in generated code it seems. How to get rid of all?

xh3b4sd:
I pushed a commit in here to fix the deep copy generation. This also fixed the problem with that `github.com/` directory which we deleted at the end of the script. That was generated and contained the deep copy functions. I just dropped the output base and use the default. I think this makes sense. FYI @tfussell

marians:
Thanks @xh3b4sd ! Now the PR is actually ready for review. :)

tfussell:
> I pushed a commit in here to fix the deep copy generation. This also fixed the problem with that `github.com/` directory which we deleted at the end of the script. That was generated and contained the deep copy functions. I just dropped the output base and use the default. I think this makes sense. FYI @tfussell

@xh3b4sd Nice one. The script looks cleaner now too.

PR #429, opened 2020-04-22
Check for NonStructuralSchema with kind

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/9386

It's possible to create a CRD which is valid (even with generated CRDs) but is not accepted by the Kubernetes API due to `NonStructuralSchema`. The recommended way to test this is to actually install the CRD on a running cluster so I have implemented this as a sort of integration test using kind.

## Checklist


.circleci/config.yml
```diff
@@ -3,9 +3,49 @@ version: 2.1
...
+          command: |
+            curl -sLo kind https://github.com/kubernetes-sigs/kind/releases/download/$KIND_VERSION/kind-linux-amd64
+      - run:
+          name: Download kubectl
+          command: |
+            curl -sLO https://storage.googleapis.com/kubernetes-release/release/$KUBERNETES_VERSION/bin/linux/amd64/kubectl && chmod +x kind kubectl
+      - run:
+          name: Create cluster
+          command: |
+            ./kind create cluster --image quay.io/giantswarm/kind-node:$KUBERNETES_VERSION --name apiextensions 2> /dev/null
```
MarcelMue, line 25: Is there some idea to align here between projects? I am not against this - just wondering as I fear divergence.

cc @giantswarm/wg-integration-testing

tfussell: Same here. I was having trouble determining the current state-of-the-art when it comes to testing with kind so I decided to take the straightforward approach. If we have some tools that can do this, that would be cool.

xh3b4sd: I think this goes into the direction of getting the future vision right which SIG RelEng may want to think about.

xh3b4sd (approved):
This should help a great deal already. I guess you did some testing and broke some CRDs purposefully to see the test job fail. So when this works then let's roll.

tfussell:
> This should help a great deal already. I guess you did some testing and broke some CRDs purposefully to see the test job fail. So when this works then let's roll.

Yes, it's actually quite difficult to create broken CRDs now with all of the code generation, unit tests, and CI, but here's is an example of a failure I induced using kustomize-based post-processing of CRDs to add validation to `.metadata.creationTimestamp` which isn't allowed https://circleci.com/gh/giantswarm/apiextensions/2309

PR #430, opened 2020-04-22
Remove YAML struct tags

tfussell:
Now that we use `sigs.k8s.io/yaml`, I believe we can get rid of `yaml` struct tags. This is one less thing to keep in sync and generally makes things cleaner.

## Checklist


tuommaki (approved):
Sure, why not if it works :+1:

kopiczko (approved):
<3

xh3b4sd (approved):

MarcelMue (approved):

PR #435, opened 2020-04-27
Support v1 and v1beta1 concurrently

tfussell:
As discussed in the SIG Operator meeting today about graduating CRDs, we need to be able to migrate individual CRDs rather than an entire package such as `provider`. To enable this, I thought it might make sense to generate all CRDs as both `v1` and `v1beta1`. In this way the `NewAzureConfigCRD()` function, for example, determines which version is loaded and we can transition one-by-one. When all CRDs have moved over, we can drop the versioned directories and `v1beta1` logic.

## Checklist


config/crd/v1/application.giantswarm.io_appcatalogs.yaml
```diff
@@ -0,0 +1,121 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
```
tfussell, line 3: Pinging @giantswarm/sig-ux because the CRDs are in a new location in the repo. I think it's fine to switch our CRD docs to `v1` now even though they won't all be migrated for a few weeks. The content of the CRD and particularly the `openAPIV3Schema` is the important part and it is unchanged between the two.

xh3b4sd (approved):
Fine with me.

fiunchinho:
I don't have knowledge enough to approve these changes. Can we get more approvals so we can merge? Thanks folks!!

xh3b4sd:
I can approve again if that helps. :trollface:

tfussell:
I'll go with this in that case.

PR #436, opened 2020-04-28
Prepare release of v0.3.3

marians:


calvix (approved):

xh3b4sd:
Just FYI we do not need to necessarily create a release for each and every change we make here.

marians:
> Just FYI we do not need to necessarily create a release for each and every change we make here.

Yeah, but we need one in order to publish CRD docs.

PR #437, opened 2020-04-30
make more values optional

anvddriesch:
We want to make the `replicas` and `AvailabilityZones` fields optional for defaulting and allow all replica values in the `status`.

## Checklist


pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -56,7 +56,8 @@ type AWSControlPlaneSpec struct {
 	// Configures which AWS availability zones to use by master nodes, as a list
 	// of availability zone names like e. g. `eu-central-1c`. We support either
 	// 1 or 3 availability zones.
-	AvailabilityZones []string `json:"availabilityZones"`
+	// +kubebuilder:validation:Optional
```
xh3b4sd, line 59: Let's please put the tags above the comments.

PR #440, opened 2020-05-01
Fix typo

marians:


marians:
The `test-install` check is failing and I don't understand why.

```
#!/bin/bash -eo pipefail
./kind create cluster --image quay.io/giantswarm/kind-node:$KUBERNETES_VERSION --name apiextensions 2> /dev/null

Exited with code exit status 1
```

PR #441, opened 2020-05-06
Use CRD v1

sslavic:
Aligns with https://github.com/giantswarm/k8sclient/pull/67 so they can be used together in operatorkit.

## Checklist

### DIFF
--- a/pkg/apis/core/v1alpha1/aws_cluster_types.go
+++ b/pkg/apis/core/v1alpha1/aws_cluster_types.go
@@ -1,7 +1,7 @@
 package v1alpha1
 
 import (
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/giantswarm/apiextensions/pkg/crd"
@@ -11,8 +11,8 @@ const (
 	kindAWSClusterConfig = "AWSClusterConfig"
 )
 
-func NewAWSClusterConfigCRD() *v1beta1.CustomResourceDefinition {
-	return crd.LoadV1Beta1(group, kindAWSClusterConfig)
+func NewAWSClusterConfigCRD() *v1.CustomResourceDefinition {
+	return crd.LoadV1(group, kindAWSClusterConfig)
 }
 
 // +genclient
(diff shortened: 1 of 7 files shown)


sslavic:
- k8sclient was switched to use CRD v1 via https://github.com/giantswarm/k8sclient/pull/67
- our apiextensions has somewhat related update "Support v1 and v1beta1 concurrently" https://github.com/giantswarm/apiextensions/pull/435
- trying to use both of those libs together in operatorkit does not compile https://circleci.com/gh/giantswarm/operatorkit/7104
- our apiextensions update had in description
> In this way the NewAzureConfigCRD() function, for example, determines which version is loaded and we can transition one-by-one.

Given all above, probably I'm missing something here why all `New*CRD()` did not get changed to CRD v1 in single PR like this. Here's that PR, looking forward to your feedback.

xh3b4sd:
You may want to check https://github.com/giantswarm/giantswarm/issues/9963#issuecomment-620048484. We discussed the necessary steps you need to take and recorded our session.

tfussell:
`KVMConfig`, `KVMClusterConfig`, `FlannelConfig`, and `StorageConfig` are being graduated here https://github.com/giantswarm/apiextensions/pull/497 `AWSClusterConfig` won't be graduated according to https://github.com/giantswarm/giantswarm/issues/9963 `CertConfig` and `DrainerConfig` are already done. That just leaves `ChartConfig` which should be handled by batman at some point. This can probably be closed.

rossf7:
> That just leaves ChartConfig which should be handled by batman at some point.

@tfussell chartconfig CRD is legacy it is replaced by the app and chart CRDs. We're waiting on customers to upgrade TCs so we can delete it but as we know that takes ages.

What is the impact if we don't migrate it?

tfussell:
> What is the impact if we don't migrate it?

CRD v1beta1 is being removed in Kubernetes 1.19 which will be released very soon. We won't be able to edit the CRD (without creating a special branch of apiextensions), use validation, or defaulting. If those are all okay, I think it doesn't need to be upgraded.

rossf7:
@tfussell Thanks for clarifying and in that case we're fine. 🎉 

chartconfig CRD is only installed in tenant clusters and from 9.0.1 onwards we only use chart CRDs.

xh3b4sd:
Do we want to go with this PR or can we close it?

tfussell:
> Do we want to go with this PR or can we close it?

We're close to being able to get rid of v1beta1 CRDs entirely at which point we can get rid of much of this backward compatibility. Moreover, we only use these CRDs in opsctl now so we could move some or all of the CRD management code there. In short, I'd say we should close it unless someone needs these V1 functions very soon.

rossf7:
Migration from v1beta1 is nearly done. We can finish that up in fresh PRs. Closing.

PR #443, opened 2020-05-06
Add conntrack to AWSCluster

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/10373

## Checklist


xh3b4sd:
I miss how this is AWS specific. I feel like this should not be added to the `AWSCluster` CR. Also why do we need to configure this per TC and not just once per installation?

tfussell:
> I miss how this is AWS specific. I feel like this should not be added to the `AWSCluster` CR.

It's not AWS specific per se. We just need it to be configurable for each TC and `AWSCluster` is where we store high-level cluster-specific info. Otherwise we could say the same of the OIDC settings for example.

> Also why do we need to configure this per TC and not just once per installation?

connection tracking is a tradeoff between memory and traffic. TCP has to keep state for each opened connection and each connection takes a few bytes of memory. On the flip side, clusters with many new connections per second will quickly hit the maximum number of entries (default is 32k per core IIRC). It can take several gigabytes of memory for machines with 10k+ requests per second so a reasonable value depends on the type of machine and the type of traffic it is serving. I believe this is sufficient reason to make it configurable per-cluster.

xh3b4sd:
Thanks for the feedback. Then I think we need to make any new field we introduce optional and omitempty. The transition will not work out well otherwise. Already made experiences with the pod CIDR.

tfussell:
> Thanks for the feedback. Then I think we need to make any new field we introduce optional and omitempty. The transition will not work out well otherwise. Already made experiences with the pod CIDR.

Thanks, good call. I made the `KubeProxy` sectional optional for validation. I don't think `omitempty` will do anything because it's a struct (not a pointer). Do you think this is sufficient? How can I test this with a real cluster?

xh3b4sd:
The result of making structs omitempty is `foo: {}` as opposed to this. 

```
foo: 
  bar: ""
```

I would like to make it as pretty and as clean as possible.

MarcelMue:
@xh3b4sd Is this now fine to you?

PR #447, opened 2020-05-07
Add v1alpha3 Cluster API types w/ MachinePool as well

tuommaki:
In order to start working on Cluster API support for Azure, we need the latest
CRDs as a base. This change also incorporates new experimental MachinePool type.

## Checklist


config/crd/v1/cluster.x-k8s.io_clusters.yaml
```diff
@@ -25,7 +25,7 @@ spec:
       jsonPath: .status.phase
       name: Phase
       type: string
-    name: v1alpha2
+    name: v1alpha3
```
tuommaki, line 28: I don't want to replace `v1alpha2`. We need both, but how can I achieve this? @tfussell do you have ideas?

scripts/gen.sh
```diff
@@ -110,6 +110,26 @@ for version in v1 v1beta1; do
   rm config/crd/$version/cluster.x-k8s.io_machines.yaml
   rm config/crd/$version/cluster.x-k8s.io_machinesets.yaml
 
+  pushd "$dir" > /dev/null
+  "$toolpath/controller-gen" \
+    crd \
+    paths=sigs.k8s.io/cluster-api/api/v1alpha3 \
+    output:dir="../config/crd/$version" \
+    crd:crdVersions="$version"
```
tuommaki, line 118: I'm not fully sure if CAPI types need `v1beta1` versions of CRD? Shouldn't `v1` be enough?

config/crd/v1/exp.cluster.x-k8s.io_machinepools.yaml
```diff
@@ -0,0 +1,466 @@
...
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.4
+  creationTimestamp: null
+  name: machinepools.exp.cluster.x-k8s.io
+spec:
+  group: exp.cluster.x-k8s.io
```
xh3b4sd, line 11: I think using even more experimental crap from upstream is a good thing. Think about the hustle we had already. I would vote for sticking with the MachineDeployments for now until there is more stability and clarity guaranteed on how the community moves forward. Nobody knows how much sense all of this makes. We have one working approach and should maybe just stick to it for now. This would get us some parity and not more divergence between providers.

tuommaki: I'm not so sure about this. I'd like to finally be semantically on the same page with upstream. Use of `MachineDeployment` was temporary hack we had to do when there was nothing better available.

Use of correct types semantically also help with getting on the same page architecturally with upstream design as for example we want to use bootstrapping with `ignition-operator`.

As we have learned with k8s upgrades in general, we will never get rid of the hassle. Things move forward and old things get deprecated. Just like with labels, CRD versions, validation, you name it. :slightly_smiling_face:

xh3b4sd: Disagree and commit. Just saying I will not advocate that move for AWS until I get a bit more comfortable with the maturity of the idea at hand.

scripts/gen.sh
tfussell: My plan for now is to generate `v1beta1` and `v1` for all CRDs. Then, once we finish the migration to v1 everywhere, we can drop `v1beta1` here for all of them. It simplifies CRD loading, documentation, and the migration process itself.

config/crd/v1/cluster.x-k8s.io_clusters.yaml
tfussell: kubebuilder supports multiple versions. Let me see if CAPI APIs have the correct labels for this. If they don't, we'll have to do it in a somewhat less elegant way, but we'll make it work.

tfussell: Done in https://github.com/giantswarm/apiextensions/pull/448

tfussell:
To be extra safe, you could install this CRD on an AWS installation with a branch of `opsctl ensure crds` and spin up a node pool cluster just to be sure, but as far as I understand it, this will have no effect on operators using `v1alpha2` provided `v1alpha2` is unchanged between CAPI releases v0.2.10 and v0.3.5.

scripts/gen.sh
```diff
@@ -102,13 +102,22 @@ for version in v1 v1beta1; do
   pushd "$dir" > /dev/null
   "$toolpath/controller-gen" \
     crd \
-    paths=sigs.k8s.io/cluster-api/api/v1alpha2 \
+    paths=sigs.k8s.io/cluster-api/api/... \
     output:dir="../config/crd/$version" \
     crd:crdVersions="$version"
   popd > /dev/null
-  # We only want Cluster and MachineDeployment for now, so delete the other two CAPI CRDs.
+  # We only want Cluster, MachineDeployment, and MachinePool for now, so delete the other two CAPI CRDs.
```
tfussell, line 109:
```suggestion
  # We only want Cluster, MachineDeployment, and MachinePool for now, so delete the other CAPI CRDs.
```

xh3b4sd:
For the record, Tuomas and I figured this breaks the installations. We fetch the latest upstream version to get the `v1alpha3` types so that we have all schema validation versions in the CRD. Now upstream already sets `v1alpha3` to `served: true` and `storage: true`, which breaks our TCs because there is some migration required to actually change the stored version. So getting all updates is cool but we need more control over the active schema validation version in order to graduate the CRDs to `v1`. The migration issue for the stored version is separate and we need to figure this later as well. /cc @tfussell

tuommaki:
`master` went through bigger overhaul. Getting these changes back is now easy with gained knowledge so I'll provide new PR later. Dropping this one.

PR #449, opened 2020-05-07
Add load balancer type

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/9332

This PR adds a new field in the `ClusterKubernetesIngressController` struct to allow customer to choose the load balancer type they want to use for the ingress controller (either public or internal)


tfussell:
Still working on this one? I can give a hand with translating it to the new structure if so.

PR #455, opened 2020-05-11
Add CRD categories

tfussell:
This PR uses kubebuilder comments to add categories to all CRDs. Categories are supported by `kubectl` so this allows us to list all Giant Swarm CRs with `kubectl get giantswarm`, AWS-related CRs with `kubectl get aws` etc.

I'm also creating a function to list all CRDs so we can potentially install CRDs with `opsctl` by category as well like `opsctl ensure crds --category aws`. This also means we don't need to keep a static list of all CRDs in the loading test.


MarcelMue:
Also some more reviewers would be okay as we touch everyones CRDs here :D

tfussell:
Pinging @giantswarm/sig-ux too as we might be able to leverage this for CRD docs, pulling the `provider` and `topic` from these categories.

config/crd/v1/core.giantswarm.io_awsclusterconfigs.yaml
```diff
@@ -10,6 +10,9 @@ metadata:
 spec:
   group: core.giantswarm.io
   names:
+    categories:
+    - aws
```
kopiczko, line 14: I'm thinking if `aws` isn't too generic? Shouldn't we prefix everything with `giantswarm-`?

tfussell: I was thinking we could get all GS CRs with `kubectl get giantswarm`. If we end up with other third-party CRDs also in the `aws` category, it seems okay if they show up in `kubectl get aws`. If you feel that it's useful, I don't see a problem with adding it in addition to plain `aws`. Also we have to type these repeatedly for operations work so I think it's good to keep them shorter if possible (lots of people alias `kubectl` to `k` as you know). I even thought of changing `giantswarm` to `gs`. What do you think?

tfussell: This would be more important if we started running more CRDs with categories and especially if we start running cross-provider installations like `aws-operator` on a KVM installation.

kopiczko: Tough choice. I'm leaning towards `giantswarm-aws` + `aws`. I think the previous may be useful for some scripting. But I'm not sure if this is important to have it now. I think we can add it later on when needed since we're working on smooth CRDs updates anyway. All good from my side 👍

pkg/crd/crd.go
```diff
@@ -64,72 +59,92 @@ func find(crdKind schema.GroupVersionKind, crGroup, crKind string) (interface{},
...
+
+// LoadV1 loads a v1 CRD from the virtual filesystem
+func LoadV1(group, kind string) *v1.CustomResourceDefinition {
+	for _, crd := range List() {
+		if crd.Spec.Names.Kind == kind && crd.Spec.Group == group {
+			return &crd
+		}
 	}
-	return crd
+	panic(microerror.Mask(notFoundError))
```
whites11, line 149: Do we allow a panic to be thrown?

kopiczko: That's a bit off I agree. If anything we should do `panic(microerror.JSON(microerror.Mask(notFoundError)))`. I'm surprised it even compiles. I thought panic takes string.

tfussell: `recover()` returns `interface{}`. I think the top level panic handler will stringify the panic so using a `string` ensures that the format is correct, but I think it's normal to panic with an `error`.

kopiczko: I mean when it compiles it compiles. I'm not sure about panic usage in general here.

tfussell: The reason I'm using `panic` here is that existing functions like `NewAppCatalogCRD` don't return an error and I didn't want to change the signature. I'm planning to drop those functions and use `crd.List` directly in `opsctl ensure crds` so I can probably drop the panic here and keep it in `LoadV1/LoadV1Beta1` only. I'll fix that. Thanks.

PR #457, opened 2020-05-12
Add external snat configuration

paurosello:
## Checklist


anvddriesch (approved):

marians:
Do we really want this optional? Or should it be mandatory and have a default value?

xh3b4sd:
New fields added to existing schema validation versions must be optional in order to be able to introduce them.

xh3b4sd:
I talked with Pau about this and I think this goes into the wrong direction. I see us adding new fields every other week and the CR becomes messy. I would like to thrive for a more generic approach that enables the customer to do more without Giant Swarm being a bottleneck. One idea was to add an environment file that users can create and configure to the CR like we do already with the credential secret. I would like to get more opinions on that.

kopiczko:
In the first place we should think if this can't be configured per installation. Because it's easier to remove flags from aws-operator (or any other operataor) itself than fields from the CR.

config/crd/v1/infrastructure.giantswarm.io_awsclusters.yaml
```diff
@@ -123,6 +123,12 @@ spec:
                         description: Subnet size, expresses as the count of leading
                           1 bits in the subnet mask of this subnet.
                         type: string
+                      externalSNAT:
+                        description: When set to false, pod connections outside the
+                          VPC where the pod is located will be NATed through the node
+                          primary IP. When set to true, all connections will use the
+                          pod IP.
```
pipo02mix, line 130: Which is the default? I would add it in the description

paurosello:
> In the first place we should think if this can't be configured per installation. Because it's easier to remove flags from aws-operator (or any other operataor) itself than fields from the CR.

The problem of having this attributes in the installation is that then the installation settings can never be changed. I think that installation values should be treated as default and for the attributes that cannot be changed we should pin them down in the CR.

config/crd/v1/infrastructure.giantswarm.io_awsclusters.yaml
paurosello: The default will be decided in the installation, so I might not add it here

kopiczko:
> > In the first place we should think if this can't be configured per installation. Because it's easier to remove flags from aws-operator (or any other operataor) itself than fields from the CR.
> 
> The problem of having this attributes in the installation is that then the installation settings can never be changed. I think that installation values should be treated as default and for the attributes that cannot be changed we should pin them down in the CR.

They can be changed. You update draughtsman and redeploy relevant apps. I disagree they should be defaults (or at least not most of the time). They should be global settings for installation.

pipo02mix:
But the major problem we have is some installations are multi-tenant, so our customers have internal customer which could have different ways to setup their clusters ( running in different accounts)

paurosello:
Also that some attributes cannot be changed, for example if you change the podCIDR on the installations repository you will still be able to create new clusters but old clusters will fail to upgrade due to this change as it's not doable on AWS

kopiczko:
I understand that:

- Not everything can be configured with flags.
- Not everything configured with flags can be changed.
- Not everything configured with CR can be changed.

But we should always _consider_ if the setting is global and can be configured as the operator flag for the sake of simplicity as a general rule. That's what I meant.

config/crd/v1/infrastructure.giantswarm.io_awsclusters.yaml
kopiczko: What does "external" mean here? Outside VPC? I feel the name is not descriptive enough but it may be just me. Maybe `snatMode: node|pod` or something similar would make more sense.

kopiczko: Maybe we can take inspiration from CAPI?

marians (approved):

xh3b4sd (approved):

PR #458, opened 2020-05-12
Add generation of CRD storageversion if multiple versions are present

tuommaki:
// +kubebuilder:storageversion marks the given CRD version as the one that
should be used for storage (i.e. persisting in etcd). When there are no other
versions of given CRD, it doesn't generate anything. When new version is added,
one must decided case-by-case which version should be stored in etcd.

This is preparation to introduce `infrastructure/v1alpha3` for Azure
Cluster API types and Node Pools.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add code generation directive to set CRD storage version when multiple
-  versions for given type are present.
+- Add code generation directive (`+kubebuilder:storageversion`) to set CRD
+  storage version when multiple versions for given type are present.
 
 ### Changed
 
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add code generation directive to set CRD storage version when multiple
+  versions for given type are present.
```
xh3b4sd, line 15: Let's maybe add the kubebuilder tag here.

tuommaki:
> I'm not sure if it's necessary when there is only a single version for these API groups, but it doesn't hurt to be explicit.

No, it's not necessary and that's why it doesn't generate anything in the first place. My reasoning here was more towards preventing problems when someone is adding new versions for some type. Without this directive the code generator gives weird error and if both versions have the directive I expect k8s to reject the CRD so it should make life easier in the future. That's my intention at least :slightly_smiling_face:

PR #460, opened 2020-05-12
Introduce infrastructure/v1alpha3 types for Azure

tuommaki:
Introduce initial types for infrastructure/v1alpha3 types for Azure. This
change intentionally leaves AWS types out since those are not changing at this
point. Common types are still included for completeness and further
adjustments.

## Checklist


tuommaki:
`master` went through bigger overhaul. Getting these changes back is now easy with gained knowledge so I'll provide new PR later. Dropping this one.

PR #462, opened 2020-05-14
Ludacris managed CRDs to v1

MarcelMue:
This is a collection PR to graduate all CRDs managed by ludacris to v1.

## Checklist


MarcelMue:
I will go with these now - there are still some more missing from our backlog but we can do them in a second batch.

PR #468, opened 2020-05-18
Make more attributes optional

anvddriesch:
Somehow this wasn't an issue earlier, but now it seems like I can't set optional attributes `null` for defaulting with `OPA`. This should make it possible.

## Checklist


pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
```diff
@@ -51,13 +51,16 @@ type AWSControlPlane struct {
 }
 
 type AWSControlPlaneSpec struct {
+	// +nullable
 	// +kubebuilder:validation:Optional
 	// Configures which AWS availability zones to use by master nodes, as a list
 	// of availability zone names like e. g. `eu-central-1c`. We support either
 	// 1 or 3 availability zones.
 	AvailabilityZones []string `json:"availabilityZones,omitempty"`
+	// +nullable
```
xh3b4sd, line 60: What kind of syntax is this and where is it coming from?

anvddriesch: This will add `nullable: true` in the CRD when you run the code generation. It's an openAPI thing: https://swagger.io/docs/specification/data-models/data-types/#null
And I found it here: https://github.com/giantswarm/apiextensions/blob/36e595787bd155286e4cfa64e3548ddd3a78d8b9/pkg/apis/provider/v1alpha1/azure_types.go#L37-L39

anvddriesch:
If we want to set the default value in case attributes are `null`, we need to allow to set these attributes to `null`.

pkg/apis/infrastructure/v1alpha2/aws_control_plane_types.go
xh3b4sd: I see. Just looks a bit unofficial to me without the tag namespacing. Anyway, this makes sense for arrays. Why do we need it for strings and ints? That seems unnecessary.

anvddriesch: While testing I encountered this:
```
The AWSCluster "g8kw3" is invalid: spec.provider.master.instanceType: Invalid value: "null": spec.provider.master.instanceType in body must be of type string: "null"
```
I couldn't set that string to `null`.

xh3b4sd: Why should we set the instance type to `"null"` anyway? It should either be `""` in case it should be defaulted or a proper instance type value.

anvddriesch: Ah, that is valuable information. I was working on the assumption that it's `null` if it should be defaulted. I can change that in OPA.
However, what about int? Wouldn't they be `null` if they are not set? And why is it better to use "" for strings that aren't set?

xh3b4sd: We should operator under the assumption of zero values. See https://golang.org/ref/spec#The_zero_value. An `int` is by default `0`. When we want to be able to differentiate between the user wanting to configure a zero value for good reasons or the user not having configured anything at all we need to work with pointers in case we want to support optional fields. Otherwise fields need to be required and validated accordingly.

anvddriesch:
So I removed all the `nullable` tags and used `optional` instead. It works for the validation and if nobody minds I would like to merge it. :)

xh3b4sd:
Antonia, did you notice we are about to release something and that I wanted to get Pau's feedback before merging?

anvddriesch:
Oh I'm really sorry. For some reason I missed the comment. :( I hope it didn't cause any problem.

xh3b4sd:
Don't know. We have to figure out now.

anvddriesch:
I can revert it if it helps

PR #470, opened 2020-05-18
Update example application group CRDs to include version labels

rossf7:
Example app CR doesn't include the version label. This means that app-operator cannot process it and will return a `did not find resource set` error.

## Checklist


tfussell:
@rossf7 The example CRs are defined in code for the time being so the tests are failing due to files being out of sync. You can add the labels to the object here https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/application/v1alpha1/app_types_test.go#L62 and then run `go test ./... -update` to get the CRs in sync. I'll try to fix this annoyance soon.

rossf7:
@tfussell Thanks for the pointer. There has been a lot of changes here and I haven't been following as closely as I'd like.

I really like the improvements and having these labels in the examples is important.

PR #471, opened 2020-05-18
Install CRs in kind test

MarcelMue:
## Checklist


tfussell:
Looks good. Just curious though--will we see the failure in CircleCI if we try to create an invalid CR? I imagine `kubectl create` will exit with a non-zero status causing the job to fail in this case.

MarcelMue:
> Looks good. Just curious though--will we see the failure in CircleCI if we try to create an invalid CR? I imagine `kubectl create` will exit with a non-zero status causing the job to fail in this case.

It did before but I am not sure with the `for loop` now. Not sure how that works with exit codes in bash :D

PR #473, opened 2020-05-19
Create float64 wrapper

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/10611. `kubebuilder` doesn't allow `float64` for CRD fields because it is not rount-trippable. To get around this, I changed the type of `float64` fields to `string`, but I realize now that this is not a backwards compatible change and should have been handled more elegantly. My new solution is to create custom marshal functions. In the long term, we should use `Quantity` type for these fields.

This has been successfully tested in `kvm-operator`. I will test it in `kubernetesd` before I merge it to master there.

## Checklist


config/crd/v1/core.giantswarm.io_kvmclusterconfigs.yaml
```diff
@@ -54,8 +54,10 @@ spec:
                         id:
                           type: string
                         memorySizeGB:
+                          format: double
```
MarcelMue, line 57: What happens to old versions of `kvm-operator` when this CRD is applied? What happens to existing CRs?
I feel like this can go wrong badly.

tfussell: Good point. My understanding is that this should be equivalent to the float64 type we are already using. It's worth testing with an old cluster/operator to see what happens though.

tfussell:
I'm testing this now. My plan is to start with a clean installation, create a cluster with the latest release (11.3.1), wait for it to be ready, apply this modified CRD with `kubectl apply -f`, create another 11.3.1 cluster, deploy a kvm-operator branch which uses this apiextensions branch, create an 11.3.2 (WIP) cluster, and check for errors in k8s API, operators, and CRD status.

tfussell:
This has been tested following the above plan on `geckon` in `kvm-operator` and `kubernetesd`. I had to make `nodeIndexes` optional and otherwise everything else looked good. I will test the components more fully before releasing them, but I think this change is okay for now.

PR #478, opened 2020-05-20
Support running within $GOPATH

tfussell:
Fixes https://github.com/giantswarm/giantswarm/issues/11095

## Checklist


tfussell:
> Can’t we simply export GOMODULES=on ?

I tried it with `GO111MODULE` set to `on`, `auto`, and `off`. It fails when set to `off` and no change with `on` or `auto`. If I set `--output-base .` it works correctly outside of `$GOPATH` and creates a `github.com` directory in the root of the repo inside `$GOPATH`. Conversely, if I remove `--output-base .`, it works inside `$GOPATH` and generates the files in `$GOPATH/pkg` when outside of `$GOPATH`. I'd be happy for a more elegant solution--I just can't figure one out.

kopiczko:
What if you unset GOPATH for the process?

pkg/apis/infrastructure/v1alpha2/common_cluster_status_spec.go
```diff
@@ -5,6 +5,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// +kubebuilder:object:generate=false
```
kopiczko, line 8: For my interest: why do we need that?

tfussell: I was getting these two errors:
`github.com/giantswarm/apiextensions/pkg/apis/infrastructure/v1alpha2:-: invalid type interface{github.com/giantswarm/apiextensions/pkg/apis/infrastructure/v1alpha2.CommonClusterStatusGetSetter; k8s.io/apimachinery/pkg/apis/meta/v1.Object; k8s.io/apimachinery/pkg/runtime.Object}`
`github.com/giantswarm/apiextensions/pkg/apis/infrastructure/v1alpha2:-: invalid type interface{GetCommonClusterStatus() github.com/giantswarm/apiextensions/pkg/apis/infrastructure/v1alpha2.CommonClusterStatus; SetCommonClusterStatus(ccs github.com/giantswarm/apiextensions/pkg/apis/infrastructure/v1alpha2.CommonClusterStatus)}`
It seems that this interface was being interpreted as a Kubernetes object, but I don't think that's correct so I'm explicitly disabling the generation of DeepCopy functions on these two types.

tfussell:
As you can see, I gave up on `deepcopy-gen` and tried `controller-gen` since this is what `cluster-api` uses to generate their DeepCopy functions. It seems to work everywhere and is significantly faster.

PR #480, opened 2020-05-21
Update example with externalSNAT

paurosello:
## Checklist


pkg/apis/infrastructure/v1alpha2/aws_cluster_types_test.go
```diff
@@ -103,7 +105,8 @@ func newAWSClusterExampleCR() *AWSCluster {
 				Namespace: "example-namespace",
 			},
 			Pods: AWSClusterSpecProviderPods{
-				CIDRBlock: "10.2.0.0/16",
+				CIDRBlock:    "10.2.0.0/16",
+				ExternalSNAT: &externalSNAT,
```
kopiczko, line 109: Do we really need to have a pointer for bool?

paurosello: this is used to check if the user actually set the value in the CR or no, otherwise it's defaulted to false by Go :(

kopiczko: I see.

PR #481, opened 2020-05-21
Add CR constructors

tfussell:
This PR adds CR constructor functions which inject `TypeMeta` and `ObjectMeta` so that new CRs have expected fields. These are defaulted when written to the Kubernetes API so this only affects CRs created client-side. This also allows the injection of documentation annotations on CRs in a uniform way rather than defining them throughout the codebase. Operators/services will need to create CRs using the new CR constructors to get these benefits.

## Checklist


tfussell:
> I am not sure if dropping it is the right time already. We have not transitioned all operators / services over yet. This would basically mean that we need to transition them in case we want to bugfix something through apiextensions.

I guess you're right. I will revert the new CRD function removal and make this PR just for CRs when I have time.

kopiczko:
Is it rebased?

tfussell:
Too many conflicts. I'll redo this later.

PR #484, opened 2020-05-25
fix: Appcatalog as cluster-scope

tomahawk28:
We didn't indicate appcatalog as cluster-scoped resources. It looks like we forgot to do that when we switched to `controller-gen`.

## Checklist


pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -39,7 +39,7 @@ func NewAppCatalogCR() *AppCatalog {
 // +genclient
 // +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-// +kubebuilder:resource:categories=common;giantswarm
+// +kubebuilder:resource:categories=common;giantswarm,scope=Cluster
```
tomahawk28, line 42: This is the part we forgot to add it.

rossf7 (approved):
Good catch. LGTM

PR #486, opened 2020-05-27
Add util functions for working with multiple CRD versions

tuommaki:
Introduce two functions that can be used to work with different CRD versions on
different platforms.

Main objective here is to provide functionality for `opsctl` so that AWS can
use one version of Cluster API CRDs while Azure uses different one. This way
one can specify which versions to get and which one to set as storage version.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `ExtractV1CRDVersions()` and `WithStorageVersion()` functions to work with
+  multi-versioned CRDs easier on different platforms.
+
 ## [0.4.3] 2020-05-25
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `ExtractV1CRDVersions()` and `WithStorageVersion()` functions to work with
+- Add `crd.ExtractV1CRDVersions()` and `crd.WithStorageVersion()` functions to work with
   multi-versioned CRDs easier on different platforms.
 
 ## [0.4.3] 2020-05-25
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -165,3 +165,63 @@ func LoadV1(group, kind string) *v1.CustomResourceDefinition {
 	}
 	panic(microerror.Mask(notFoundError))
 }
+
+// ExtractV1CRDVersions takes instance of CRD and variable number of versions
+// and returns CRD with those versions. This function does not guarantee that
+// all defined versions are present in returned CRDs.
+//
+// Example:
+//	crd contains versions for v1alpha1 and v1alpha3.
+//	versions is v1alpha3.
+// Returned:
+//	crd with only v1alpha3 version.
+//
+func ExtractV1CRDVersions(crd *v1.CustomResourceDefinition, versions ...string) *v1.CustomResourceDefinition {
+	crd = crd.DeepCopy()
+
+VERSION_LOOP:
+	for i := 0; i < len(crd.Spec.Versions); i++ {
+		v := crd.Spec.Versions[i]
+		for _, name := range versions {
+			if v.Name == name {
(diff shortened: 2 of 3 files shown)


tuommaki:
My intention here is to allow `opsctl ensure crds` to ensure Cluster API `v1alpha2` on AWS while using `v1alpha3` on Azure.

With these, we can still pull upstream CRDs easily with existing tooling as-is.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `ExtractV1CRDVersions()` and `WithStorageVersion()` functions to work with
```
kopiczko, line 14:
```suggestion
- Add `crd.ExtractV1CRDVersions()` and `crd.WithStorageVersion()` functions to work with
```

pkg/crd/crd.go
```diff
@@ -165,3 +165,63 @@ func LoadV1(group, kind string) *v1.CustomResourceDefinition {
...
+
+VERSION_LOOP:
+	for i := 0; i < len(crd.Spec.Versions); i++ {
+		v := crd.Spec.Versions[i]
+		for _, name := range versions {
+			if v.Name == name {
+				// Keep current version and proceed to next.
+				continue VERSION_LOOP
+			}
+		}
```
kopiczko, line 190: I'd prefer to extract a private/inline function here instead of using the label.

pkg/crd/crd.go
```diff
@@ -165,3 +165,63 @@ func LoadV1(group, kind string) *v1.CustomResourceDefinition {
...
+		// Remove current element since its version was not specified in versions.
+		crd.Spec.Versions = append(crd.Spec.Versions[:i], crd.Spec.Versions[i+1:]...)
+		i--
+	}
+
+	return crd
+}
+
+// WithStorageVersion takes CRD and version and sets storage flag for this
+// version as true while setting others to false.
```
kopiczko, line 201: Shouldn't we rather have that properly defined here in the repo? Why would we want to change that dynamically? The question applies also to the versions thing.

tuommaki: Nope. The same version of `apiextensions` must be usable and functional for all environments. When upstream introduced incompatible version changes for Cluster API types, we ran into troubles because we can't have new types in AWS nor old types in Azure because only one version is possible to exist in etcd (the storage version).

Therefore we need a mechanism where the client can decide which version is the one that is stored. I'd then use this function further in the `opsctl ensure crds` for provider specific functions to use correct storage version.

tfussell:
What do you think about putting this logic in `opsctl ensure crds` instead? I'm just thinking that this library should export standard full CRDs and we could transform them in consumers of the library like `opsctl`. Not blocking this work--just wanted to mention.

tuommaki:
> What do you think about putting this logic in opsctl ensure crds instead? I'm just thinking that this library should export standard full CRDs and we could transform them in consumers of the library like opsctl. Not blocking this work--just wanted to mention.

I was writing it there first but then got thinking that since `apiextensions` already has `crd` package which provides functionality for CRDs, it felt like a logical place for me here.

I don't have strong opinions on this though.

tfussell:
> > What do you think about putting this logic in opsctl ensure crds instead? I'm just thinking that this library should export standard full CRDs and we could transform them in consumers of the library like opsctl. Not blocking this work--just wanted to mention.
> 
> I was writing it there first but then got thinking that since `apiextensions` already has `crd` package which provides functionality for CRDs, it felt like a logical place for me here.
> 
> I don't have strong opinions on this though.

My main argument for moving it to `opsctl` would be to speed up development and reduce the surface area of the exported API of `apiextensions`. I find it a little annoying to modify the library, test the branch in the consumer (e.g. `opsctl`), merge the change in the library, create a library release, create a tag, then use the tag in the consumer.

tuommaki:
> My main argument for moving it to opsctl would be to speed up development and reduce the surface area of the exported API of apiextensions. I find it a little annoying to modify the library, test the branch in the consumer (e.g. opsctl), merge the change in the library, create a library release, create a tag, then use the tag in the consumer.

Ok. I'll move this to `opsctl`.

pkg/crd/crd.go
kopiczko: I see. That's sad.

PR #492, opened 2020-06-01
omitempty optional fields

xh3b4sd:
## Checklist


config/crd/v1/infrastructure.giantswarm.io_awsmachinedeployments.yaml
```diff
@@ -122,9 +122,6 @@ spec:
                         maximum: 100
                         minimum: 0
                         type: integer
-                    required:
-                    - onDemandBaseCapacity
-                    - onDemandPercentageAboveBaseCapacity
```
xh3b4sd, line 127: I am not sure why this changed now. Is the generation paying respect to the JSON tags on the structs?

tfussell: There's some info here https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required The important part seems to be this: 
> Note that for backward compatibility, any field that has the omitempty struct tag will be considered to be optional, but this may change in the future and having the +optional comment tag is highly recommended.

xh3b4sd: Interesting. Do you think we should add the optional kubebuilder tags to the fields that have the omitempty JSON tags?

tfussell: I think so, yes. If others agree, next steps would be:
- Update all relevant fields in `apiextensions`.
- Update documentation (which? `fmt` maybe).
- Optionally add a CI check to enforce this.

xh3b4sd: +1 on the CI check. Everything else than automatic enforcement will not work well. I guess I am good to go here for now. We can chat how to roll with the CI check. Maybe you have an idea already.

PR #493, opened 2020-06-02
Add OpenAPI spec generation

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/8622

### Purpose
To be able to generate a Kubernetes client which knows our api groups and types, and could be written in a programming language that is not Go.

This means we can generate a TypeScript client that we can use in `happa`, but it also means that customers would be able to easily generate their own clients based on our OpenAPI spec.

### How to make the generator see a type
To make a type visible to the generator, you must add the `// +k8s:openapi-gen=true` macro before its declaration: [Example](https://github.com/giantswarm/apiextensions/blob/74b166f5dce7962cb1a949277e12512d4e78235e/pkg/apis/core/v1alpha1/flannel_types.go#L23)
> Please note that even sub-types must be marked with the macro (such as the `spec` or the `status`)

> Also, the `List` types must not be marked with the macro, they will be handled by the generator.

> Groups must also be marked: [Example](https://github.com/giantswarm/apiextensions/blob/add-openapi-spec-generator/pkg/apis/example/v1alpha1/doc.go)


marians:
I'm very biased of course, as you presented this in chapter front end already and it seems to do what we want.

Just ran the `make` command and in the `Generating OpenAPI models` step I get more than 1700 complaints like

```
API rule violation: list_type_missing,github.com/giantswarm/apiextensions/pkg/apis/tooling/v1alpha1,AzureToolList,Items
API rule violation: list_type_missing,k8s.io/api/core/v1,AvoidPods,PreferAvoidPods
API rule violation: list_type_missing,k8s.io/api/core/v1,Capabilities,Add
API rule violation: list_type_missing,k8s.io/api/core/v1,Capabilities,Drop
API rule violation: list_type_missing,k8s.io/api/core/v1,CephFSPersistentVolumeSource,Monitors
API rule violation: list_type_missing,k8s.io/api/core/v1,CephFSVolumeSource,Monitors
API rule violation: names_match,github.com/giantswarm/apiextensions/pkg/apis/tooling/v1alpha1,AzureToolStatus,WorkspaceStatus
API rule violation: names_match,k8s.io/api/core/v1,AzureDiskVolumeSource,DataDiskURI
API rule violation: names_match,k8s.io/api/core/v1,ContainerStatus,LastTerminationState
API rule violation: names_match,k8s.io/api/core/v1,DaemonEndpoint,Port
...
```

axbarsan:
That's part of the generation hell... Even `kubernetes` [has a bunch of these](https://github.com/kubernetes/kubernetes/blob/36981002246682ed7dc4de54ccc2a96c1a0cbbdb/api/api-rules/violation_exceptions.list). They're just warnings, they don't break generation in any way.

go.mod
```diff
@@ -6,11 +6,14 @@ require (
 	github.com/evanphx/json-patch v4.5.0+incompatible // indirect
 	github.com/giantswarm/microerror v0.2.0
 	github.com/go-openapi/errors v0.19.4
+	github.com/go-openapi/spec v0.19.8
```
kopiczko, line 9: Do we need those dependencies here? Can this be avoided?

axbarsan: Unfortunately, yes. At least with the way the project is structured now. They're required by each `openapi_generated.go` file.
https://github.com/giantswarm/apiextensions/blob/74b166f5dce7962cb1a949277e12512d4e78235e/pkg/apis/tooling/v1alpha1/openapi_generated.go#L26

kopiczko: Would it be an option to generate the client outside the repo then? Or can we create a separate module in one of the directories? I'd like to avoid pulling those dependencies in all our operators.

axbarsan: That'a really valid question. I extracted the OpenAPI spec generation to a [separate repo](https://github.com/giantswarm/cp-api-spec), and only left the macros for the generator in this PR

marians: Feeling much better with this being separate!

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add macros for OpenAPI spec generation.
```
kopiczko, line 16: I think it's irrelevant from the user's perspective.

```suggestion
```

More detailed info: https://github.com/giantswarm/fmt/blob/master/releases/changelog.md

PR #494, opened 2020-06-11
Add new comment on app status

tomahawk28:
## Checklist


tomahawk28:
Just a comment; going with a ping.

sslavic (approved):
LGTM

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Update the `status` comment in `App` CRs.
```
kopiczko, line 16: IMO this should not be in the changelog. Also this is `Added` and not `Changed`. See https://github.com/giantswarm/fmt/blob/master/releases/changelog.md

PR #488, opened 2020-05-29
lastDeployed in app & chart CRs as nullable field

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11261

Making lastDeployed as nullable. 

## Checklist

- [x] Consider SIG UX feedback.
- [x] Update changelog in CHANGELOG.md.
- [ ] If adding a new type, ensure that it is added to the [CRD unit tests](https://github.com/giantswarm/apiextensions/blob/d78aba91d578d56ef49f58e336035d35edc4e1b0/pkg/crd/crd_test.go#L9).

PR #496, opened 2020-06-16
Make fields optional for App and AppCatalog CRDs

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/11556

See https://gigantic.slack.com/archives/CR99LSZ1N/p1592311735122800. 

Mea culpa I think I approved the PR 🤦‍♂️ but the current openapi validation is too strict and isn't helping.

## Checklist


config/crd/v1/application.giantswarm.io_appcatalogs.yaml
```diff
@@ -73,9 +73,6 @@ spec:
                     - name
                     - namespace
                     type: object
-                required:
-                - configMap
-                - secret
```
rossf7, line 78: It's valid for an appcatalog to have no configuration. I think that applies to the releases catalog.

config/crd/v1/application.giantswarm.io_apps.yaml
```diff
@@ -115,9 +112,7 @@ spec:
                     - namespace
                     type: object
                 required:
-                - context
                 - inCluster
-                - secret
```
rossf7, line 120: Only inCluster is required. For CP app CRs we set this to true.

config/crd/v1beta1/application.giantswarm.io_apps.yaml
```diff
@@ -76,9 +76,6 @@ spec:
                   - name
                   - namespace
                   type: object
-              required:
-              - configMap
-              - secret
```
rossf7, line 81: Same for app CRs. Its valid to have no config or userConfig.

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -45,7 +45,7 @@ func NewAppCatalogCR() *AppCatalog {
 
 type AppCatalog struct {
 	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata"`
```
rossf7, line 48: Same change as https://github.com/giantswarm/apiextensions/pull/492.

Using omitempty for the optional fields is a lot nicer when using kubectl IMO.

marians (approved):

tomahawk28 (commented):
From the last PR https://github.com/giantswarm/apiextensions/pull/488/files, I learn that `nullable` property is also necessary.

rossf7:
> From the last PR https://github.com/giantswarm/apiextensions/pull/488/files, I learn that nullable property is also necessary.

@tomahawk28 Thanks good catch. I'll get that fixed.

rossf7:
@tomahawk28 Should be good now. PTAL

tomahawk28 (approved):

tomahawk28:
Sorry for late, LGTM!

tomahawk28:
I found a PR a year ago.
> The problem is the operator changes got big and it needs logic to handle nils which is easy to miss
https://github.com/giantswarm/apiextensions/pull/285

I think we don't have those cases to miss anymore, is that right?

rossf7:
> Sorry for late, LGTM!

Np :) ❤️ 

Tested on `geckon`.

rossf7:
> I found a PR a year ago.

> The problem is the operator changes got big and it needs logic to handle nils which is easy to miss
#285

@tomahawk28 Thanks for posting this. I think we do need these changes because for example `.Spec.Config` can be nil when getting `.Spec.Config.ConfigMap.Name`.

But now that app CRs are customer facing I think it's important we get rid the empty string fields. So I'll re-implement the changes in https://github.com/giantswarm/app-operator/pull/136

PR #500, opened 2020-06-25
Add ref releases

MarcelMue:
## Checklist


config/crd/v1beta1/release.giantswarm.io_releases.yaml
```diff
@@ -76,9 +77,18 @@ spec:
               description: Components describes components used in this release.
               items:
                 properties:
+                  catalog:
```
kopiczko, line 80: Do we need this?

config/crd/v1beta1/release.giantswarm.io_releases.yaml
```diff
@@ -76,9 +77,18 @@ spec:
...
+                    default: control-plane-catalog
+                    description: Catalog specifies the name of the app catalog that
+                      this component belongs to.
+                    type: string
                   name:
                     description: Name of the component.
                     type: string
+                  reference:
+                    description: Reference is the git tag or SHA specifying the component
```
kopiczko, line 89: I guess you don't mean SHA. It's always the version in catalog. In test catalog you'll have versions like `0.0.1-921feaa96a9882b974755c87eb8d852afb6e173a`.

MarcelMue: Yes - description is not accurate yet :D

config/crd/v1beta1/release.giantswarm.io_releases.yaml
MarcelMue: So current thinking is we do need this - it is more clear then to define if a reference is from a different catalog. Without this we would need some implicit magic to determine the correct catalog which we already do in opsctl. So I would not like to reimplement that somewhere else.

config/crd/v1beta1/release.giantswarm.io_releases.yaml
stone-z: Updated

stone-z:
Since this is only additions, I _believe_ it is backwards-compatible

MarcelMue:
Please fix merge conflicts, then this should be ready to go

stone-z:
Oh boy this merge might be ugly. Here goes nothing

stone-z:
Huh. everything-went-better-than-expected.png

MarcelMue:
We should add something in the changelog afterwards :D

stone-z:
Whoops. https://github.com/giantswarm/apiextensions/pull/521

PR #503, opened 2020-07-03
Add a component flag for the release operator

yulianedyalkova:
The flag specifies if the component should be deployed by the release-operator automatically. 

Default value is false. This will make migration and testing easier, because we won't have to manually edit all the releases to exclude the components that we don't want.


pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -92,6 +92,9 @@ type ReleaseSpecComponent struct {
 	// +kubebuilder:validation:Optional
 	// Reference is the git tag or SHA specifying the component source.
 	Reference string `json:"reference,omitempty"`
+	// +kubebuilder:validation:Optional
```
MarcelMue, line 95: Should we still add `// +kubebuilder:default=false` or is that unnecessary?
I am not sure how it shows up in the generated docs.

yulianedyalkova: Which docs? In the release resource it just omits it if it's empty which ends up being treated as false.

MarcelMue: Here: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
But I don't think we show defaults there, so it should be fine like this.

PR #505, opened 2020-07-06
Allow suffixes in release names

yulianedyalkova:
This is part of the [release-operator implementation](https://github.com/giantswarm/giantswarm/issues/11716) and it will allow people to create test releases by simply adding a suffix to the name. So I will be able to call my test release `v13.3.0-yulia` instead of coming up with some random numbers.


config/crd/patches/v1/release.giantswarm.io_releases/patch.yaml
```diff
@@ -4,5 +4,5 @@
     properties:
       name:
         type: string
-        pattern: ^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$
+        pattern: ^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-[1-9a-z]*)?$
```
tfussell, line 7: You might want to allow `0` in the suffix so that a version like `1.0.0-beta.10` is accepted.

MarcelMue:
Change is LGTM some TODO's for testing:
- vendor this into `opsctl`
- run `opsctl ensure crds` against some test installation
- see if new validation explodes

yulianedyalkova:
> Change is LGTM some TODO's for testing:
> 
> * vendor this into `opsctl`
> * run `opsctl ensure crds` against some test installation
> * see if new validation explodes

I did this already. I even tried creating a cluster with a release version with a suffix on azure to double check if the operator is going to explore (because some of them use it).

PR #510, opened 2020-07-08
Release notes url in wide output

ubergesundheit:
This PR adds a new column to `kubectl get releases -o wide` output.

```bash
kubectl get release -o wide
NAME      KUBERNETES VERSION   STATE    AGE                    RELEASE NOTES
v11.2.0   1.16.3               active   2020-03-03T11:12:13Z   https://github.com/giantswarm/releases/tree/master/kvm/v11.2.0
```

I am unsure why the `AGE` column is different to output on live installations. I've tested this with a local kind cluster (k8s 1.18.2)

## Checklist


pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -8,9 +8,11 @@ import (
...
 
 const (
-	crDocsAnnotation         = "giantswarm.io/docs"
-	kindRelease              = "Release"
-	releaseDocumentationLink = "https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/"
+	crDocsAnnotation            = "giantswarm.io/docs"
+	crReleaseNotesURLAnnotation = "giantswarm.io/release-notes"
+	kindRelease                 = "Release"
+	releaseDocumentationLink    = "https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/"
+	releaseNotesLink            = "https://github.com/giantswarm/releases/tree/master/kvm/v11.2.0"
```
tfussell, line 15: Maybe the default should be an empty string or `https://github.com/giantswarm/releases`.

MarcelMue: I would prefer to just link to the repo as well :)

ubergesundheit: Ok just so that we're on the same page: Once I change the value here, I also need to change it in [`config/crd/v1/release.giantswarm.io_releases.yaml`](https://github.com/giantswarm/apiextensions/pull/510/files#diff-47531fe5a991d9cd7e1b8cecc5260012) to satisfy the tests.

Is the value here a default value for the release CR?

tfussell: Yes, exactly. We create release CRs with helm+kustomize instead of operators so this is unlikely to come up in the near future anyways (i.e. this code won't be executed anywhere except in the test).

ubergesundheit: Thanks for the explanation. I'll change this now.

PR #512, opened 2020-07-09
Adds deprecation notices for AWSConfig and StorageConfig

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/10599

## Checklist


anvddriesch (approved):

cornelius-keller (approved):

tfussell:
Are we really deprecating StorageConfig? I haven't heard about a plan for its replacement.

tfussell:
Since the CRDs are generated from code now, you'll need to add this notice as a comment on the struct rather than in the CRD YAML. Add here https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/provider/v1alpha1/aws_types.go#L47 and then run `make`. The result should be the same.

config/crd/v1/core.giantswarm.io_storageconfigs.yaml
```diff
@@ -22,6 +22,7 @@ spec:
   - name: v1alpha1
     schema:
       openAPIV3Schema:
+        description: StorageConfig used to provide storage for Giant Swarm API microservices. Deprecated.
```
kopiczko, line 25: What do we use instead?

marians (approved):

marians:
@JosephSalisbury Please run the `make` command to adapt generated CRD YAML files and then commit the result.

marians:
I'll take over

tfussell:
Why is StorageConfig deprecated?

MarcelMue:
> Why is StorageConfig deprecated?

+1 Two people asked this and then it just got merged?

PR #514, opened 2020-07-10
Use `date` as age column type in release CRD

ubergesundheit:
This PR changes the `type` of the `age` column to `date` in the release CRD.

After ensuring the latest release of apiextensions (containing #510) to gaia I noticed the format of the `AGE` column did change from relative to an absolute format.

Unfortunately I couldn't find any change whch could have caused this to change.

## Checklist


ubergesundheit:
I think this only affects the formatting of the output of `kubectl get`. But to be sure I'll try to play around with `format` too

ubergesundheit:
- `format=date,type=string` prints the age column in RFC3339 format
- `format=date,type=date` prints the age column in relative format

so I think I'll leave this to just `type=date` since `format seems to be optional.

PR #517, opened 2020-07-13
Update changelog to prepare for release 0.4.13

marians:
Changelog update


xh3b4sd:
I think we should use the release workflows here. Then the release process is automated and you do not have to do all of this. This obviously expects everyone to write their changelogs in the first place when a change is being made.

marians:
How is it done? Where should I read?

anvddriesch:
The docs + video are here: https://intranet.giantswarm.io/docs/dev-and-releng/how-to-release-a-project/

PR #515, opened 2020-07-10
Update Cluster API dependencies

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/9

We need new `sigs.k8s.io/cluster-api` version with `MachinePool` `.Spec.FailureFomains` field. Updated dependencies:
- `sigs.k8s.io/cluster-api` v0.3.5 -> v0.3.7-rc.1
- `sigs.k8s.io/cluster-api-provider-azure` v0.4.3 -> v0.4.5

PR #520, opened 2020-07-14
Apply upstream k8s 1.18.5

tomahawk28:
Towards giantswarm/giantswarm#12180

Applying k8s 1.18 into apiextensions first so we can align with k8s 1.18 changes from other operators. 

## Checklist


pkg/clientset/versioned/typed/application/v1alpha1/app.go
```diff
@@ -38,15 +39,15 @@ type AppsGetter interface {
...
-	Create(*v1alpha1.App) (*v1alpha1.App, error)
-	Update(*v1alpha1.App) (*v1alpha1.App, error)
-	UpdateStatus(*v1alpha1.App) (*v1alpha1.App, error)
-	Delete(name string, options *v1.DeleteOptions) error
-	DeleteCollection(options *v1.DeleteOptions, listOptions v1.ListOptions) error
-	Get(name string, options v1.GetOptions) (*v1alpha1.App, error)
-	List(opts v1.ListOptions) (*v1alpha1.AppList, error)
-	Watch(opts v1.ListOptions) (watch.Interface, error)
-	Patch(name string, pt types.PatchType, data []byte, subresources ...string) (result *v1alpha1.App, err error)
+	Create(ctx context.Context, app *v1alpha1.App, opts v1.CreateOptions) (*v1alpha1.App, error)
```
tomahawk28, line 42: As you see, we add `context.Context` as the first parameter.

hack/tools/client-gen/go.mod
```diff
@@ -2,4 +2,4 @@ module github.com/giantswarm/apiextensions/hack/tools/client-gen
 
 go 1.14
 
-require k8s.io/code-generator v0.16.6
+require k8s.io/code-generator v0.18.0
```
tomahawk28, line 5: code generator also need to follow v0.18.0

hack/go.mod
```diff
@@ -3,14 +3,18 @@ module github.com/giantswarm/apiextensions/hack
...
 replace (
-	k8s.io/api => k8s.io/api v0.16.6
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.16.6
-	k8s.io/apimachinery => k8s.io/apimachinery v0.16.6
-	k8s.io/apiserver => k8s.io/apiserver v0.16.6
-	k8s.io/client-go => k8s.io/client-go v0.16.6
+	k8s.io/api => k8s.io/api v0.18.0
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.18.0
+	k8s.io/apimachinery => k8s.io/apimachinery v0.18.0
+	k8s.io/apiserver => k8s.io/apiserver v0.18.0
```
tomahawk28, line 18: Updating all upstream k8s repo to 0.18.0

go.mod
```diff
@@ -7,10 +7,9 @@ require (
...
 	github.com/google/go-cmp v0.4.0
-	k8s.io/api v0.16.6
-	k8s.io/apiextensions-apiserver v0.16.6
-	k8s.io/apimachinery v0.16.6
-	k8s.io/client-go v0.16.6
-	k8s.io/utils v0.0.0-20200324210504-a9aa75ae1b89 // indirect
+	k8s.io/api v0.18.0
+	k8s.io/apiextensions-apiserver v0.18.0
+	k8s.io/apimachinery v0.18.0
+	k8s.io/client-go v0.18.0
```
tomahawk28, line 13: Here are upgrading part, too.

hack/go.mod
tfussell: Why not go with the latest, v0.18.5?

tomahawk28: You're right but the helm version that we are trying to introduce is v3.2.4 which is the latest at the moment and they are still using v0.18.0. 
https://github.com/helm/helm/blob/0ad800ef43d3b826f31a5ad8dfbb4fe05d143688/go.mod#L41
I'm not sure we can just put `replace` directives from helmclient (redirect client-go v0.18.0 to v0.18.4) and make other repos to 0.18.5. Should we try it?

tfussell: I think we can replace it. I've tried to standardize repos that I touch to k8s v1.16.6 even if a dependency uses a different version and haven't run into any issues yet. I'm not sure if it matters that much either way--just nice to be consistent.

tomahawk28: Then I would vote to use 1.18.5 everywhere and replace it only for helmclient. @rossf7 WDYT?

rossf7: Yes consistency is important for me too. If we can use 1.18.5 everywhere except helmclient and that works then I'm happy.

Helm `v3.3.0-rc1` shipped a week ago but its still only using 1.18.4. So either way we'll be behind.

https://github.com/helm/helm/blob/5c2dfaad847df2ac8f289d278186d048f446c70c/go.mod#L41

hack/go.mod
```diff
@@ -3,14 +3,18 @@ module github.com/giantswarm/apiextensions/hack
 go 1.14
 
 require (
-	sigs.k8s.io/cluster-api v0.3.5
-	sigs.k8s.io/cluster-api-provider-azure v0.4.3 // indirect
+	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
+	github.com/hashicorp/golang-lru v0.5.4 // indirect
+	github.com/imdario/mergo v0.3.8 // indirect
+	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
+	sigs.k8s.io/cluster-api v0.3.6
```
nprokopic, line 10: @giantswarm/team-celestial needs newer CAPI here (for Node Pools story). There is a PR for upgrading to `v0.3.7-rc.1` https://github.com/giantswarm/apiextensions/pull/515

nprokopic: I will update that PR to use `v0.3.7` which was released this morning.

tomahawk28: Thanks for the update, I will hold this PR until you merged.

config/crd/v1/exp.cluster.x-k8s.io_machinepools.yaml
```diff
@@ -402,8 +402,31 @@ spec:
...
+                    and require a REST mapping.  In most cases, the dependency is
+                    on the group,resource tuple     and the version of the actual
+                    struct is irrelevant.  5. We cannot easily change it.  Because
+                    this type is embedded in many locations, updates to this type     will
+                    affect numerous schemas.  Don''t make new APIs embed an underspecified
+                    API type they do not control. Instead of using this type, create
+                    a locally provided and used type that is well-focused on your
+                    reference. For example, ServiceReferences for admission registration:
+                    https://github.com/kubernetes/api/blob/release-1.17/admissionregistration/v1/types.go#L533
+                    .'
```
rossf7, line 429: I guess they are fine but do we know where these changes are coming from?

I'm assuming its upstream docs from CAPI?

rossf7 (commented):
Direction is good but I lack CAPI knowledge. So good to hear more opinions here.

config/crd/v1/exp.cluster.x-k8s.io_machinepools.yaml
fiunchinho: I guess so too

tomahawk28: This is from the latest CAPI https://github.com/giantswarm/cluster-api/blob/master/config/crd/bases/exp.cluster.x-k8s.io_machinepools.yaml#L9

hack/go.mod
tomahawk28: @nprokopic Because In case you concern, v0.3.7 was released a few days ago, I updated it to 0.3.7 in this PR (although by a forked repo)

nprokopic: We ended up using `v0.3.7-rc.1` in order not to wait for CAPI release. There isn't any difference that's relevant for us in `v0.3.7` when comapred to `v0.3.7-rc.1`, as we only use CRDs.

So feel free to bump `v0.3.7-rc.1` to `v0.3.7` if you want. :)

config/crd/v1/exp.cluster.x-k8s.io_machinepools.yaml
nprokopic: Yes, CRD yaml files for Cluster API and Cluster API CRDs are generated from the upstream types.

rossf7 (approved):
LGTM

PR #526, opened 2020-07-17
create CRs

xh3b4sd:
## Checklist

## Context

I want to consolidate functionality. The code in this PR comes from https://github.com/giantswarm/kubectl-gs. We want to use the functionality used to create the CRs also in https://github.com/giantswarm/awscnfm and maybe in other places as well. So we should have one place to create the CRs. Since we define the structure and schema already in `apiextensions` it would make sense to just have the helper functions here. Do you think this goes into the right direction? Note that the code does not fully compile. I just want to show off an idea and get quick feedback without wasting time on polishing anything. When we get the go Team Firecracker will iterate on this.


anvddriesch:
I think it would be very helpful to have these helper functions since we want to create CRs in quite a few different places.

tfussell:
I like it. I had a similar idea a few months ago and created this PR https://github.com/giantswarm/apiextensions/pull/481 which adds new CR functions for all CRs along with a documentation annotation. Maybe we can combine our work and get this into `master` soon. Over time, we could then use these functions to construct CRs in operators/services rather than starting from an empty struct.

kopiczko:
I'm fine with that too.

xh3b4sd:
Fishing for LGTM.

xh3b4sd:
Using these changes in https://github.com/giantswarm/kubectl-gs/pull/103.

pkg/id/id.go
```diff
@@ -0,0 +1,44 @@
...
+
+const (
+	// IDChars represents the character set used to generate cluster IDs.
+	// (does not contain 1 and l, to avoid confusion)
+	IDChars = "023456789abcdefghijkmnopqrstuvwxyz"
+	// IDLength represents the number of characters used to create a cluster ID.
+	IDLength = 5
+)
+
+func Generate() string {
```
njuettner, line 18: In case we generate an already existing id, do we check for that?

docs/cr/infrastructure.giantswarm.io_v1alpha2_awsmachinedeployment.yaml
```diff
@@ -2,14 +2,21 @@ apiVersion: infrastructure.giantswarm.io/v1alpha2
 kind: AWSMachineDeployment
 metadata:
   annotations:
-    giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/awsclusters.infrastructure.giantswarm.io/
```
njuettner, line 5: good catch

pkg/id/id.go
xh3b4sd: We don't and I think we can't. The chances are also super low and on top the IDs can still be configured upfront.

PR #527, opened 2020-07-17
Initial type for pre-ignition-operator simulation of CAPI bootstrap process

tuommaki:
## Checklist


tfussell:
Do you have an issue explaining what this CRD is? Just curious.

fiunchinho:
> Do you have an issue explaining what this CRD is? Just curious.

We are implementing [this proposal](https://github.com/giantswarm/giantswarm/issues/10808) since implementing the ignition operator would be too much work, and that would delay releasing node pools in Azure. Basically, we want something like the `Ignition` CR, but we didn't want to mess with the existing definition. We also figured it would be easier to just change to a different CRD when the ignition operator would be ready than having to deal with different versions.

Does that help?

tfussell:
> Does that help?

Makes sense. Thanks! We were discussing ignition-operator today so I mainly wanted to see how we could collaborate on it in the coming weeks/months. I'll keep an eye on the linked epic and this PR.

PR #530, opened 2020-07-21
Graduate application CRDs to v1

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/9963

## Checklist


xh3b4sd:
Are you sure this is good enough? I would have expected CRD schema changes because the validation is different and when we start to graduate CRDs we should have the versioned scheme in place, so that we have e.g. `v1alpha2` and `v1alpha3` and so on.

rossf7:
> Are you sure this is good enough? I would have expected CRD schema changes because the validation is different.

@xh3b4sd No I'm not sure. 😅  How can I learn more about this? The meet recording?

I'll compare with the CAPI CRs to see if I've missing something. We already have OpenAPI validation rules in place because the CRDs are customer facing.

> when we start to graduate CRDs we should have the versioned scheme in place, so that we have e.g. v1alpha2 and v1alpha3 and so on.

At the moment we plan to stay on v1alpha1 but we will definitely be making structural changes in future. So yes we need to get this right.

tfussell:
> Are you sure this is good enough? I would have expected CRD schema changes because the validation is different and when we start to graduate CRDs we should have the versioned scheme in place, so that we have e.g. `v1alpha2` and `v1alpha3` and so on.

This is a good point. The safest way to have done CRD graduation would be to have kept the existing API version without a schema (using `x-kubernetes-preserve-unknown: true`) and then creating a new version with a full schema. Unfortunately, we didn't have a good story in place for dealing with multiple versions in a CRD, using conversion webhooks, or generating CRDs like this using kubebuilder, but we are working on iit. With this in mind, I think it's fine to migrate these CRDs as we have for the others simply by making the implicit schema (from marshalling/unmarshalling the CR structs) explicit and looking for validation errors. For future breaking CRD changes, we should definitely use API versioning.

xh3b4sd:
I just had a look at https://github.com/giantswarm/apiextensions/blob/master/config/crd/v1/application.giantswarm.io_apps.yaml which looks about right. So when you do not get any weird status conditions after ensuring on a control plane this should be good.

rossf7:
@tfussell @xh3b4sd Yes we added the schema quite early because customers create apps CRs directly with the CP API. We had some bugs but they should be fixed. We're testing on ginger to see if we missed any.

We will want to change the CRD schema in future but there is no urgent need IMO. I was expecting we'd need a conversion webhook for that. So I think we're aligned. Thx!

rossf7:
Tested on `ginger`.

We're going to roll out in the morning if our overnight test looks OK. Merging so we can do the release prep.

PR #531, opened 2020-07-22
Use annotations from annotation package in example CRs

marians:
The goal here is to use a single source of truth for our annotation keys.

## Checklist


pkg/annotation/annotation.go
```diff
@@ -4,3 +4,7 @@ package annotation
 // documentation. This aims to help understanding all the moving parts within
 // the system and how they relate to each other.
 const Docs = "giantswarm.io/docs"
+
+// ReleaseNotes is the annotation key for a URL to release notes, to learn
+// more about details of a release.
+const ReleaseNotes = "giantswarm.io/release-notes"
```
xh3b4sd, line 10: When I saw this my assumption was we add it to all CRs. Below I figured we only put this on the `Release` CRs. If this is the intention we should make this more clear in the godoc. Above the docs annotation godoc says `is the docs annotation put into all CRs`. IMO it is super important to make clear what it is used for and where to expect it.

CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Annotation key constant `ReleaseNotes`.
+
+### Changed
+
+- Use annotations from `annotation` packe in example CRs.
```
kopiczko, line 18: Irrelevant for the end user IMO. And surely not worth next major.

```suggestion
```

marians: What is an end user when it comes to apiextensions?

pkg/annotation/annotation.go
marians: OK, makes sense. Since I cannot predict where this annotation key will be used in the future, and I don't want to limit where it should be used, I will rephrase the description to fit more use cases.

marians:
@kopiczko Apart from your comment on the changelog, do you think this PR is worth your approval?

CHANGELOG.md
kopiczko: Someone using the library. From the outside it's transparent.

kopiczko:
> @kopiczko Apart from your comment on the changelog, do you think this PR is worth your approval?

Code-wise LGTM

marians:
Had to change the annotation name, as somebody already created the same annotation with name `annotation.ReleaseNotesURL` and I didn't want to break the API.

CHANGELOG.md
marians: Anyway, removed.

PR #532, opened 2020-07-22
Add annotations

marians:
This should add all annotations defined in https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md to the repo.

## Checklist


pkg/annotation/machinedeployment.go
```diff
@@ -0,0 +1,6 @@
+package annotation
+
+// MachineDeploymentSubnet is used to specify the IPv4 CIDR used for a
+// node pool defined by a MachineDeployment or AWSMachineDeployment
+// resource.
```
xh3b4sd, line 5: I think we should stick to established terminology and not derive from what people are used to. So either name it `CR` or `custom resource`. 

```suggestion
// node pool defined by a MachineDeployment or AWSMachineDeployment CR.
```

marians:
> I found the PR description a bit misleading. It tells me all labels and annotations in the fmt ...

Actually this PR here is only about annotations, and the title and description say so. There is a separate PR for labels.

PR #533, opened 2020-07-22
Add label definitions

marians:
This adds all the labels defined in https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md which are not third party labels to the repo.

## Checklist


pkg/label/app.go
```diff
@@ -0,0 +1,11 @@
+package label
+
```
kopiczko, line 2: Shouldn't we add to changelog that those things were added?

rossf7:
We're going to merge #520 later today with K8s 1.18 support. This will be a major version bump to v2.0.0.

@marians Do you want to release this before we do that?

stone-z:
@marians I'm planning to do an apiextensions release shortly, would you like to include these?

PR #535, opened 2020-07-23
Use values from pkg/annotation in release_types

ubergesundheit:
This PR adds the release notes annotation key to `pkg/annotation`

## Checklist


marians:
For future annotations I would suggest to align the constant name with the key string.

E. g. key string `giantswarm.io/release-notes` should result in `ReleaseNotes`.

PR #539, opened 2020-07-23
Add EndOfLifeDate to Release CRD

marians:
Towards https://github.com/giantswarm/giantswarm/issues/12224

This adds an End of Life date to the Release CRD, with the goal to be able to communicate when support for a tenant cluster using this release will end.

## Checklist


pkg/apis/release/v1alpha1/release_types.go
```diff
@@ -74,11 +74,21 @@ type Release struct {
...
+
 	// Date that the release became active.
 	Date *metav1.Time `json:"date"`
+
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// EndOfLifeDate is the date and time when support for a tenant cluster using
+	// this release ends. This may not be set at the time of release creation
+	// and can be specififed later.
+	EndOfLifeDate *metav1.Time `json:"endOfLifeDate"`
```
tfussell, line 90:
```suggestion
	EndOfLifeDate *metav1.Time `json:"endOfLifeDate,omitempty"`
```
This hasn't really been documented yet, but I think it's better to use `omitempty` whenever we make a field optional.

marians: Which only means that Go will not write this attribute when creating a JSON representation, in the case the value is `nil`. Do we consider this cleaner?

tfussell: It's cleaner, but there are more good reasons here https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

config/crd/v1/release.giantswarm.io_releases.yaml
```diff
@@ -112,6 +112,13 @@ spec:
                 description: Date that the release became active.
                 format: date-time
                 type: string
+              endOfLifeDate:
+                description: EndOfLifeDate is the date and time when support for a
```
xh3b4sd, line 116: I would like to emphasize what end of support means and what it doesn't. So I am not sure if EOL is the right term here. Or at least the description could be more elaborative on the fact that your workloads can continue running but that you are effectively responsible yourself for all consequences when not upgrading anymore, meaning, we are not going to do anything about it anymore. At the same time, we are not assholes and want to be there for our customers.

pkg/apis/release/v1alpha1/release_types.go
tfussell: I think this is relevant to the percent spot instances field you were working on recently too.

config/crd/v1/release.giantswarm.io_releases.yaml
marians: I think the details should not be defined in the CRD, but instead in some public documentation article.

marians: Once we have that, we can even link out from that description.

config/crd/v1/release.giantswarm.io_releases.yaml
```diff
@@ -112,6 +112,13 @@ spec:
                 description: Date that the release became active.
                 format: date-time
                 type: string
+              endOfLifeDate:
```
MarcelMue, line 115: Just wondering:
Why not make this an annotation? I am thinking about doing the same to the creation time.

marians: What's the benefit of an annotation?

MarcelMue: A benefit would be that we keep the CRD a bit more clean. I have been thinking about it more recently to try and justify why something should be in the CRD instead of just metadata on the objects.

I might be wrong here - so not blocking.

pkg/apis/release/v1alpha1/release_types.go
marians: There we removed `omitempty` again because it caused us trouble. But that was an integer field and it was not declared as a pointer. So `omitempty` resulted in the attribute being omitted whenever the value was zero. Here the case is different.

marians: Added `omitempty` here

calvix (approved):
LGTM< but maybe others have opinion as well

njuettner (approved):
lgtm

marians:
I'd like an approval from @giantswarm/team-ludacris, too.

@MarcelMue Is your concern a blocker?

tfussell (approved):

stone-z:
Not blocking, but just to echo Marcel's concern -- the EOL date seems like it is mostly for display to humans. It also seems like the kind of thing which might need to be changed on a given release for various reasons in the future (I know we don't want to but it happens). It feels weird to me to have to then update the otherwise static release resource. Are two identical releases with different EOLs equivalent?

xh3b4sd (approved):
Defining end of life in the spec makes sense to me.

marians:
> Are two identical releases with different EOLs equivalent?

I am not sure what two identical releases are. If there would be two releases that only differ with regard to their EndOfLifeDate, something unrelated to this date must have gone wrong.

anvddriesch (approved):
I like the idea of having the EOL as part of the spec as well instead of just an annotation. This emphasizes that there is a best-before and clusters shouldn't be kept on outdated releases forever.
That will help a lot to move our focus towards new releases.

marians:
> The EOL date seems like it is mostly for display to humans.

That is one purpose. The goal is to make something which is currently hidden in contractual prosa much more concrete and visible. There will likely be automation evolving relying on this new attribute. That's why I think it's important to have it as a "first class citizen" in the schema.

stone-z:
Just a final thought on this, I realize it is already merged.

IMO, a release defines a specific state of the software. Releases should be immutable.
An EOL date is a future intention. Realistically, EOL dates will change, no matter how much we wish they wouldn't.

> The goal is to make something which is currently hidden in contractual prosa much more concrete and visible. There will likely be automation evolving relying on this new attribute

Contractual considerations are part of our human relationship with the customer. Force-deleting clusters when their EOL expires would be insane, so -- as we do now -- we will inevitably choose ignore the EOL when it suits our human processes.

At some point, we'll have to either change the date on a given release (to satisfy hypothetical automation) or ignore it. If we change it, releases aren't immutable. If we ignore it, then I'm even more convinced it is a purely human concern*. Improving EOL visibility and automation could be achieved via other means without making a release mutable.

But, I see I'm the minority, I disagree and commit

\* (and if automation exists, we'd likely have to pacify it with some exceptions list or override configuration, which seems like more work)

marians:
Regarding "Force-deleting": nobody recommends anything like this. Please read https://github.com/giantswarm/giantswarm/issues/12224 for context.

I acknowledge that in fact with the EOL date we are adding a mutible attribute to the CRD. But looking at the `state` attribute, that changes over the lifecycle, too (wip, active, deprecated). I think the eol date is very much in line with the state attribute.

PR #547, opened 2020-08-09
Add monitoring labels

mig4:
These labels can be applied to `Service` resources to configure how
Giant Swarm Prometheus instances scrape those services.

Towards https://github.com/giantswarm/giantswarm/issues/11720

Towards https://github.com/giantswarm/giantswarm/issues/11724

## Checklist


mig4:
The links are 404 because the PR adding the doc is still in progress - https://github.com/giantswarm/giantswarm/pull/12512

rossf7:
We're going to merge https://github.com/giantswarm/apiextensions/pull/520 later today with K8s 1.18 support. This will be a major version bump to v2.0.0.

@mig4 Do you want to release this before we do that?

pkg/label/monitoring.go
```diff
@@ -0,0 +1,27 @@
...
+const MonitoringPort = "giantswarm.io/monitoring-port"
+
+// MonitoringPath is an annotation that tells Prometheus the URL path a service
+// exposes metrics on. It can be used on `Service` resources and will be picked
+// up by `simple-service` `ServiceMonitor` to configure Prometheus.
+//
+// The default is to use the path `/metrics`.
+//
+// Spec: https://intranet.giantswarm.io/docs/architecture-specs-adrs/specs/configuration-of-targets-in-tc-prometheus/
+const MonitoringPath = "giantswarm.io/monitoring-path"
```
kopiczko, line 27: Maybe endpoint is a better name than a path. No big deal though.

mig4: These are already in use in managed apps so while I agree the naming could be better, changing them at this point probably wouldn't be worth the effort.

kopiczko: Understood :+1:

PR #549, opened 2020-08-13
make chart.config nullable 

tomahawk28:
We found out `chart` CRD needs the same changes as https://github.com/giantswarm/apiextensions/pull/496 to make it less rigorous from passing empty values. 

## Checklist


rossf7 (approved):
LGTM

marians (approved):

PR #551, opened 2020-08-14
Add managed-by label and version labels for app-operator and chart-operator

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/11748

## Checklist


pkg/label/id.go
```diff
@@ -17,6 +17,10 @@ const MachineDeployment = "giantswarm.io/machine-deployment"
 // giantswarm.io/machine-pool=de19f-h94vd.
 const MachinePool = "giantswarm.io/machine-pool"
 
+// ManagedBy is the name of the operator managing the object. E.g.
+// giantswarm.io/managed-by=app-operator.
+const ManagedBy = "giantswarm.io/managed-by"
```
rossf7, line 22: We use this label in both app-operator and chart-operator.

PR #561, opened 2020-09-16
Initial proposal for the NetworkPool CRD

paurosello:
Towards https://github.com/giantswarm/giantswarm/issues/10862

Spec: https://intranet.giantswarm.io/docs/architecture-specs-adrs/specs/aws_flexible_ipam/

## Checklist


paurosello:
@marians do you have some time today to have a look at this? :)

config/crd/v1beta1/infrastructure.giantswarm.io_awsclusters.yaml
```diff
@@ -114,6 +114,14 @@ spec:
                       description: AWS EC2 instance type to use for the master node.
                       type: string
                   type: object
+                nodes:
+                  description: Node network configuration.
+                  properties:
+                    ipPool:
+                      description: IP Pool represents the collection of IPs that master
+                        and workers subnets will use.
+                      type: string
```
anvddriesch, line 123: I wasn't involved in the discussion about this, so I'm curious to understand the relationship between the `awscluster` and `awsippool` resources. Will the `ipPool` attribute contain some sort of reference / name that connects the two?

paurosello: yes, this attribute will reference an IPPool and that makes me think IPPool should be a Cluster Wide resource :)

xh3b4sd: Shouldn't we use a real object reference instead some magic string that we treat like an object reference?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add `IPPool` for AWS Clusters
```
kopiczko, line 12:
```suggestion
### Added

- Add `IPPool` for AWS Clusters
```

config/crd/v1beta1/infrastructure.giantswarm.io_awsclusters.yaml
anvddriesch: The thing about the object reference is that the object has to exist before you can reference it when you create the `awcluster` CR, or you need to update it later which can be awkward in some cases.
Who is going to create the `awsippool` CR?

xh3b4sd: What you describe is a requirement imposed by some upstream controller. I think we could just use the type and data structure so that we at least align with the way how objects are referenced. How the `aws-operator` deals with the referenced information is up to the controller.

marians: To me it appears quite logical that the IPPool has to exist in order to be used by the cluster. So a proper object reference would make sense.

anvddriesch: I agree that an object reference would make sense. It would also include the namespace.

paurosello: AWSIPPool will need to exist before the cluster is created, so it's fine with putting a reference and assuming it exits when the cluster is created.

Customers are going to create this resources

paurosello:
Pending questions that came out of the call with @marians :
- [ ] Does this concept of IPPool need to be provider independent? Other providers like KVM or Azure might want to implement something similar.
- [ ] We need to either make the CR namespaced or clusterwide. Namespaced will allow better authorization (IPPool will belong to organizations), clusterwide seems to align more with the "uniqueness" that we need to ensure anyway.

Other considerations:
- This CR will have more attributes in the future, which is why we created a new CRD
- The naming is still a bit wonky

I think we need to involve SIG Architecture to make this decisions @alex-dabija

alex-dabija:
I do think the IPPool concept needs to be provider independent because a similar feature could be requested or implemented for at least Azure. At the moment the CRD is simple enough (just has CIRD block) to be independent. Let's talk today about this in SIG-Architecture SWH.

I would scope the IPPool CRs to the organization namespace, because:
- it's easier to authorize access with RBAC ( as you mentioned);
- we have more flexibility in the future if organizations are used to separate multiple teams using the same installation (which probably is the case at the moment).

In both situations we still have to ensure that the CIDR blocks declared in IPPools don't overlap globally on a Control Plane by implementing a validation via an admission controller (or similar). The only difference being the query which gets executed:
- give me all IPPools (if globally scoped);
- give me all IPPools from all namespaces (if organization namespaced scoped).

alex-dabija:
Could we compile a list of alternative names for IPPool to discuss in SIG Architecture and in SIG UX?

Some ideas:
- IPBlock
- CIDR
- TCIPConfig (Tenant Cluster IP configuration) (we might introduce additional fields like workersSubnetSize, etc.)

teemow:
Can you also check if there are similar thoughts in the upstream cluster api and if not at least run the idea against them?

alex-dabija:
Unfortunately, the upstream CAPI doesn't have a concept similar to IPPools. From what I have seen, the network configuration is defined statically on the Cluster and AWSCluster CRs. They don't seem to have a way of dynamically figuring out what the VPC CIDR should be. I still have to look into the actual implementation because VPCSpec and Subnets fields are both optional and it could be that there is some code to specify defaults or determine the values in some other way.

@paurosello also looked into this and has reached a similar conclusion.

JosephSalisbury:
+1 to provider independent. 

Not sure about namespaced vs cluster-scoped.

alex-dabija:
Today, in SIG Architecture we agreed on having the IPPool concept provider independent and organization namespace scoped.

There were also a few naming suggestions: 
- SubnetPool
- NetworkPool

@paurosello is going to talk @marians to agree on the name. In the meeting we were leaning towards NetworkPool .

paurosello:
And we have decisions: 
- It will be namespaced, aligned with organizations and shared controlplanes....
- It will be provider independent all providers will benefit from it

The last one, naming. There are 2 options left:
- SubnetPool
- NetworkPool
- IPPool is "discarded" as calico is already using it, might create confusion and it's already present in our clusters

config/crd/v1beta1/infrastructure.giantswarm.io_awsclusters.yaml
paurosello: This will be namespaced, but it will always be in the same namespace as the cluster exists, do we still want to make it a full reference?

pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go
```diff
@@ -156,6 +159,14 @@ type AWSClusterSpecProviderPods struct {
 	ExternalSNAT *bool `json:"externalSNAT,omitempty"`
 }
 
+// AWSClusterSpecProviderNodes Nodes network configuration.
+// +k8s:openapi-gen=true
+type AWSClusterSpecProviderNodes struct {
+	// +kubebuilder:validation:Optional
+	// IP Pool represents the collection of IPs that master and workers subnets will use.
```
marians, line 166:
```suggestion
	// NetworkPool represents a range of IP addresses to chose chunks from for master and worker node subnets.
```

config/crd/v1/infrastructure.giantswarm.io_awsclusters.yaml
```diff
@@ -115,6 +115,14 @@ spec:
                         description: AWS EC2 instance type to use for the master node.
                         type: string
                     type: object
+                  nodes:
+                    description: Node network configuration.
+                    properties:
+                      networkPool:
+                        description: NetworkPool represents a range of IP addresses
+                          to chose chunks from for master and worker node subnets.
```
pipo02mix, line 123:
```suggestion
                          to choose chunks from for master and worker node subnets.
```

config/crd/v1beta1/infrastructure.giantswarm.io_awsclusters.yaml
```diff
@@ -114,6 +114,14 @@ spec:
                       description: AWS EC2 instance type to use for the master node.
                       type: string
                   type: object
+                nodes:
+                  description: Node network configuration.
+                  properties:
+                    networkPool:
+                      description: NetworkPool represents a range of IP addresses
+                        to chose chunks from for master and worker node subnets.
```
pipo02mix, line 122:
```suggestion
                        to choose chunks from for master and worker node subnets.
```

pipo02mix (approved):

paurosello:
I will merge this at the end of the day if there is no additional suggestions

anvddriesch:
Did we come to an agreement on the string vs object reference thing? :)

config/crd/v1beta1/infrastructure.giantswarm.io_awsclusters.yaml
kopiczko: So with the namespaces being organizations clusters will have to be in the same org to share IP pool. Is it ok?

paurosello: Yes, it was decided that this was the case. If you want the cluster in the same NetworkPool on a different org you can still create the same NetworkPool in multiple orgs.

paurosello:
As the NetworkPool will be in the same NS as the Cluster we decided to avoid the reference for now

PR #567, opened 2020-09-22
Marked nullable fields in etcd types

whites11:
This PR adds some annotations in the ETCDBackup type to allow for some fields to be nullable


pkg/apis/backup/v1alpha1/etcd_backup_types.go
```diff
@@ -39,12 +39,18 @@ type ETCDBackupSpec struct {
...
 // +k8s:openapi-gen=true
 type ETCDBackupStatus struct {
+	// +kubebuilder:validation:Optional
+	// +nullable
 	// map containing the state of the backup for all instances
 	Instances map[string]ETCDInstanceBackupStatusIndex `json:"instances,omitempty"`
 	// Status of the whole backup job (can be 'Pending', 'Running'. 'Completed', 'Failed')
 	Status string `json:"status"`
+	// +kubebuilder:validation:Optional
+	// +nullable
```
tfussell, line 49: I'm not totally sure, but I think `+nullable` is only required for reference types (maps and slices) and maybe pointers where you might pass the literal value `null`. `omitempty` is sufficient for value types I believe. It would be cool to nail this down and document it if you have a little time to experiment.

whites11: nope, without the nullable it fails:
```
ETCDBackup.backup.giantswarm.io "etcd-backup-20200922142731" is invalid: [status.finishedTimestamp: Invalid value: "null": status.finishedTimestamp in body must be of type string: "null", status.startedTimestamp: Invalid value: "null": status.startedTimestamp in body must be of type string: "null"] kind:unknown 
```

tfussell: Interesting. Thanks for trying. I'll be sure to dig into this more later on.

pkg/apis/backup/v1alpha1/etcd_backup_types.go
```diff
@@ -63,18 +69,32 @@ type ETCDInstanceBackupStatusIndex struct {
...
 	// Timestamp when the first attempt was made
 	StartedTimestamp metav1.Time `json:"startedTimestamp,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +nullable
 	// Timestamp when the last (final) attempt was made (when the Phase became either 'Completed' or 'Failed'
 	FinishedTimestamp metav1.Time `json:"finishedTimestamp,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +nullable
```
tfussell, line 81: I believe at least these `string` and `int64` values don't need `nullable`. I checked the rest of our CRDs and we only use nullable for structs (including `time.Time`), maps, and slices.

whites11: ok let's try

whites11: confirmed, it works for strings and int64s.

tfussell: Good to hear. Nice to avoid unnecessary clutter.

tfussell (approved):
LGTM.

PR #569, opened 2020-09-23
Delete CRDs for Azure managed AKS clusters

tuommaki:
We don't use nor support tenant clusters using AKS and hence we don't need
related CRDs.

## Checklist


tuommaki:
We discussed this. Nothing controversial. Going with a ping.

PR #572, opened 2020-09-24
Add AppCatalogEntry CRD to the application group

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/10930

Spec: https://intranet.giantswarm.io/docs/architecture-specs-adrs/specs/managed-apps/2020-05-05-app-versions-representation/#how-to-make-the-data-available-in-the-control-plane-api

Adds first iteration of the new appcatalogentry CRD that just uses fields from index.yaml in the catalog repo. I'd like to get a review from Batman first and then will open up to SIG UX.

## Checklist


docs/cr/application.giantswarm.io_v1alpha1_AppCatalogEntry.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: giantswarm-nginx-ingress-controller-app-1.9.2
+spec:
+  appName: nginx-ingress-controller-app
+  appVersion: v0.35.0
+  catalog: ""
+  chart:
+    home: https://github.com/giantswarm/nginx-ingress-controller-app
+    icon: https://upload.wikimedia.org/wikipedia/commons/nginx-logo.svg
+  dateUpdated: "2020-09-02T09:40:39Z"
+  version: 1.9.2
```
rossf7, line 14: baby_yoda

@giantswarm/team-batman @puja108 @pipo02mix A new baby CRD

docs/cr/application.giantswarm.io_v1alpha1_AppCatalogEntry.yaml
```diff
@@ -0,0 +1,14 @@
...
+  creationTimestamp: null
+  name: giantswarm-nginx-ingress-controller-app-1.9.2
+spec:
+  appName: nginx-ingress-controller-app
+  appVersion: v0.35.0
+  catalog: ""
+  chart:
+    home: https://github.com/giantswarm/nginx-ingress-controller-app
+    icon: https://upload.wikimedia.org/wikipedia/commons/nginx-logo.svg
+  dateUpdated: "2020-09-02T09:40:39Z"
```
puja108, line 13: I'm guessing the reason why this is not in status is that the "updating" is an update that happens outside of the operator/cr loop, e.g. in CI/github?

rossf7: > I'm guessing the reason why this is not in status is that the "updating" is an update that happens outside of the operator/cr loop, e.g. in CI/github?

@puja108 Yes for this first dev iteration the CRD will only using fields from the catalog `index.yaml`. So I think the date should be in the CR spec rather than the status.

In index.yaml this field is called `created` but in the spec it's named `dateUpdated`. @piontec I'm fine with the name but what was the reason for that?

docs/cr/application.giantswarm.io_v1alpha1_AppCatalogEntry.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: application.giantswarm.io/v1alpha1
+kind: AppCatalogEntry
+metadata:
+  creationTimestamp: null
+  name: giantswarm-nginx-ingress-controller-app-1.9.2
+spec:
+  appName: nginx-ingress-controller-app
+  appVersion: v0.35.0
+  catalog: ""
```
tomahawk28, line 9: Is this `catalog` could be empty like this? I'm asking since it's `required` field from CRD spec.

rossf7: Thanks I missed setting this in the unit test that generates the example. 

It's there now and I also set the namespace since the new CRD is namespace scoped.

docs/cr/application.giantswarm.io_v1alpha1_AppCatalogEntry.yaml
puja108: the `updated` does confuse a bit, a tag/version of an app gets created once, it should never get updated, should it?

rossf7: The app itself should not be updated.  But once we add the metadata files they will get updated. e.g. We run compatibility tests for a new release.

Do we need created and updated in the CR? So we can show in the UX when this version was created.

piontec: So, as you say, `updated` is because metadata will change now and I think it's really important to check when was the last update of the CR and be able to compare that with what we have in app catalog's repo (debugging mainly). As for `created`... what value does it bring? I mean, sure, we can include it just because it's in the `index.yaml`... Maybe it's useful to show the date when the app version was introduced for the first time (created) and last updated? I'm not sure...

piontec: [1 min later] OK, actually I think it's useful to some extent: if the app version was created 1 year ago, but updated by some build tools 5 mins ago, it's still the old version from 1 year ago, not something fresh releases 5 mins ago. So, it's useful, let's use it :+1:

piontec (approved):
LGTM.
@rossf7 , do you think it already makes sense to include `catalogNamespace` and make it optional for now? Or mandatory with default value of `""`? The idea is to avoid resources migration later when we switch to namespace things

config/crd/v1/application.giantswarm.io_appcatalogentries.yaml
```diff
@@ -0,0 +1,94 @@
...
+  group: application.giantswarm.io
+  names:
+    categories:
+    - common
+    - giantswarm
+    kind: AppCatalogEntry
+    listKind: AppCatalogEntryList
+    plural: appcatalogentries
+    singular: appcatalogentry
+  scope: Cluster
```
pipo02mix, line 20: But if App Catalog is namespaced, should not be entry namespaced too?

piontec: Yeah, should be - smae thing. We know we'll have to move both to namespace CRs, but not just yet. And I'm just wondering if it's worth doing some preparation steps now already.

config/crd/v1/application.giantswarm.io_appcatalogentries.yaml
```diff
@@ -0,0 +1,94 @@
...
+                    nullable: true
+                    type: string
+                type: object
+              dateUpdated:
+                description: DateUpdated is when this entry was last updated. e.g.
+                  2020-09-02T09:40:39.223638219Z
+                format: date-time
+                type: string
+              version:
+                description: Version is the version of the app for this entry. e.g.
```
pipo02mix, line 74: What about `app package` or `app chart` to discriminate of `appVersion`

pipo02mix (commented):
Some comments and I have also a doubt. We define URL, name,home,... but what about description. We will need that for displaying more info about the app entry

puja108:
> do you think it already makes sense to include catalogNamespace and make it optional for now? Or mandatory with default value of ""? The idea is to avoid resources migration later when we switch to namespace things

Might be good to either have the namespacing prepared and have this namespaced already. Maybe even could combine it with a new version of catalog CRD that is namespaced (without the migration for now, just to already have it and not need migration for everything).

as for a `catalogNamespace` field, would it not make sense to have it like ingress in that it just references its own namespace? like and Entry always lives in the same namespace as its catalog? I cannot think about a use case where catalog and entry should be in separate namespaces, would make things very weird.

rossf7:
Thanks for the feedback. I created a spec PR with the changes. https://github.com/giantswarm/giantswarm/pull/13529

I think its easier to review there and I want to keep the spec in sync. 

> do you think it already makes sense to include catalogNamespace and make it optional for now? Or mandatory with default value of ""? The idea is to avoid resources migration later when we switch to namespace things

> Might be good to either have the namespacing prepared and have this namespaced already. Maybe even could combine it with a new version of catalog CRD that is namespaced (without the migration for now, just to already have it and not need migration for everything).

@piontec @puja108 See https://github.com/giantswarm/giantswarm/pull/13529/files#r497349555
I've added separate fields for the name and namespace like we have in the other CRDs.

I think that prepares us for making the appcatalog CRD namespace scoped. But changing that even without a migration is a lot of changes so I don't think we should do that now.

> We define URL, name,home,... but what about description. We will need that for displaying more info about the app entry

@pipo02mix I want to keep this first dev iteration lean because we don't have much time to work on it. I checked Happa and it looked like icon was the only field used from index.yaml. I'd like to sync with @oponder when he's back on whether we need more fields.

docs/cr/application.giantswarm.io_v1alpha1_AppCatalogEntry.yaml
rossf7: Both dates are there now.

rossf7 (commented):
All changes made and the CRD is now **namespace** scoped not cluster scoped :)

PTAL

config/crd/v1/application.giantswarm.io_appcatalogentries.yaml
```diff
@@ -0,0 +1,111 @@
...
+              version:
+                description: Version is the version of the app chart for this entry.
+                  e.g. 1.9.2
+                type: string
+            required:
+            - appName
+            - appVersion
+            - catalog
+            - dateCreated
+            - dateUpdated
```
puja108, line 98: could `dateUpdated` be empty? Like when a tag got created and the app is there but no additional CI steps ran to update the entry?

puja108 (commented):
minor question, otherwise LGTM

config/crd/v1/application.giantswarm.io_appcatalogentries.yaml
piontec: I assume at the beginning `dateCreated` == `dateUpdated`, which indicates "creation was the latest update of the state"

rossf7: My idea was for new entries to set it to the same as dateCreated.

So dateCreated never changes and matches the created timestamp in index.yaml. dateUpdated gets updated if the metadata file get updated.

WDYT?

rossf7:
Thanks @puja108 

If we're OK on the 2 date fields then this should be good to go. @tomahawk28 Could you take another look?

config/crd/v1/application.giantswarm.io_appcatalogentries.yaml
rossf7: Our updates crossed. What Lukasz said :)

puja108 (approved):

docs/cr/application.giantswarm.io_v1alpha1_AppCatalogEntry.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: application.giantswarm.io/v1alpha1
+kind: AppCatalogEntry
+metadata:
+  creationTimestamp: null
+  name: giantswarm-nginx-ingress-controller-app-1.9.2
```
tomahawk28, line 5: Is that convention that we put {namespace}-{app-name}-{appVersion} format as metadata name?

tomahawk28 (approved):
LGTM, just one question from my side.

docs/cr/application.giantswarm.io_v1alpha1_AppCatalogEntry.yaml
rossf7: Yes this was my idea on naming. As I think it will prevent any collisions. WDYT?

There will be a new appcatalogentry resource in app-operator that will control this and it replaces the current dummy resource in the appcatalog controller.

rossf7:
> - [ ] Consider SIG UX feedback.

I missed this step but its a new customer facing CRD. So I think it's important to get feedback from the SIG.

@giantswarm/sig-ux PTAL 🙏 

I'd like to merge this by EOD Monday. In the meantime I can start using this in app-operator.

docs/cr/application.giantswarm.io_v1alpha1_AppCatalogEntry.yaml
tomahawk28: > There will be a new appcatalogentry resource in app-operator that will control this and it replaces the current dummy resource in the appcatalog controller.

👍

pipo02mix (approved):
I am happy by now

rossf7:
Going with this. Thanks for the feedback.

PR #577, opened 2020-10-01
Add networkpool support on AWS Clusters

paurosello:
## Checklist

### DIFF
--- a/pkg/apis/infrastructure/v1alpha2/aws_cluster_types_test.go
+++ b/pkg/apis/infrastructure/v1alpha2/aws_cluster_types_test.go
@@ -42,6 +42,11 @@ func Test_GenerateAWSClusterYAML(t *testing.T) {
 		},
 		{
 			category: "cr",
+			name:     fmt.Sprintf("%s_%s_awscluster.yaml", group, version),
+			resource: newAWSClusterEmptyNetworkPoolCR(),
+		},
+		{
+			category: "cr",
 			name:     fmt.Sprintf("%s_%s_awscluster_networkpool.yaml", group, version),
 			resource: newAWSClusterNetworkPoolCR(),
 		},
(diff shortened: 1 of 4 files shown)


xh3b4sd:
Some thoughts. 

* Are there some tests we can extend or add? 
* Should we have additional validation? 
* How is the optionality of network pools still ensured?

paurosello:
I created some tests, thanks for reminding :+1: 

I am checking that setting an empty value for the NetworkPool produces same output as not even specifying the attribute in order to test optionality.

The second test I added is to check that the CR is correctly generated when specifying a NetworkPool

pkg/apis/infrastructure/v1alpha2/aws_cluster_types_test.go
```diff
@@ -144,3 +154,86 @@ func newAWSClusterExampleCR() *AWSCluster {
 
 	return cr
 }
+
+func newAWSClusterEmptyNetworkPoolCR() *AWSCluster {
+	awscluster := newAWSClusterExampleCR()
```
xh3b4sd, line 159: Just FMI why do we have `newAWSClusterExampleCR` and `NewAWSClusterCR`? 🙈

xh3b4sd: I see in https://github.com/giantswarm/apiextensions/blob/58f996a5fea29796572cf3909214cabffbd0b37c/pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go#L27-L37 that the spec is empty when using `NewAWSClusterCR`. /shrug

xh3b4sd (approved):

pkg/apis/infrastructure/v1alpha2/aws_cluster_types_test.go
paurosello: yup, not sure about why do we have that empty CR function

PR #580, opened 2020-10-07
Update k8s related dependencies

fiunchinho:
The `cluster-api` dependency on version `v0.3.7` was bringing k8s dependencies using `0.17.*` versions. Updating it so we use only `0.18.*`.

## Checklist


fiunchinho:
This should not affect @giantswarm/team-firecracker-engineers since they use old `v1alpha2` types which are not being developed.

fiunchinho:
going with a ping

PR #582, opened 2020-10-08
Prepare next major, add warning to changelog

fiunchinho:
For some reason that we don't know about yet, projects depending on latest release don't compile anymore because they try to use `cluster-api` `v0.3.7` together with k8s 1.18 dependencies.

We don't know what it started to happen in the last release, since we already did that change in the prior releases.

To fix this issue, projects depending on this library need to explicitly add a `replace` directive to their `go.mod` files. According to [the docs](https://github.com/golang/go/wiki/Modules#gomod)
> exclude and replace directives only operate on the current (“main”) module. exclude and replace directives in modules other than the main module are ignored when building the main module. The replace and exclude statements, therefore, allow the main module complete control over its own build, without also being subject to complete control by dependencies.

That's why it's not enough specifying the `replace` here in `apiextensions`.

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/giantswarm/apiextensions/v2
 go 1.14
 
 require (
-	github.com/giantswarm/microerror v0.2.0
+	github.com/giantswarm/microerror v0.2.1
 	github.com/go-openapi/errors v0.19.4
 	github.com/google/go-cmp v0.4.1
 	k8s.io/api v0.18.9
@@ -14,7 +14,4 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
-replace (
-	sigs.k8s.io/cluster-api v0.3.10 => github.com/giantswarm/cluster-api v0.3.10-gs
-	sigs.k8s.io/cluster-api v0.3.7 => github.com/giantswarm/cluster-api v0.3.7-gs
-)
+replace sigs.k8s.io/cluster-api => github.com/giantswarm/cluster-api v0.3.10-gs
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -14,4 +14,4 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
-replace sigs.k8s.io/cluster-api v0.3.10 => github.com/giantswarm/cluster-api v0.3.10-gs
+replace sigs.k8s.io/cluster-api => github.com/giantswarm/cluster-api v0.3.10-gs
```
rossf7, line 17: OK sorry but maybe I get this now :(

An explanation for this is a newer dependency is trying to use a version other than `v0.3.10`.

IMO this is now correct. As we always want to use our fork and this version.

rossf7 (approved):
LGTM

But will dependabot create PRs for the major bump? Doing it manually is possible but kind of a pain.

fiunchinho:
> LGTM
> 
> But will dependabot create PRs for the major bump? Doing it manually is possible but kind of a pain.

I think it will, but the `replace` needs to be manually added to `go.mod`.

rossf7:
> I think it will, but the replace needs to be manually added to go.mod.

Thanks and if the WG can help with some of the busy work we can do that. Happy to take some PRs.

We're also using the CAPI fork in aws-operator and cluster-operator. So I think we should try and keep them in sync. WDYT?

fiunchinho:
AFAIK it's not just those repositories but *all repositories that depend on `apiextensions`*.

PR #586, opened 2020-10-09
Bump go module version to v3

fiunchinho:
Since we released the v3.0.0 we need to also increase the go module version.

## Checklist


rossf7 (approved):

nprokopic (approved):
LGTM

PR #588, opened 2020-10-13
set lastDeployed, revision as nullable

tomahawk28:
`lastDeployed`, `revision` is useful only if they contain the value from helm release. 
This PR is to hiding these fields if it's not present. 

Current:
```yaml
appVersion: ""
release:
  lastDeployed: null
  revision: 0
  status: ""
version: ""
```

Desired:
```yaml
appVersion: ""
release:
  status: ""
version: ""
```

## Checklist


rossf7 (approved):
LGTM

The chartcrd resource in app-operator should apply this. Not sure about CPs. We might need to use opsctl for that.

tomahawk28:
> Not sure about CPs. We might need to use opsctl for that. 

I think it's up to `opsctl`, I will test this in ginger by opsctl actually :)

tomahawk28:
Damn, I need to update change the `opsctl` dependencies which are not easy. I will merge this and test on the tenant cluster.

PR #596, opened 2020-10-26
Basic Ready conditions for AzureCluster

nprokopic:
## Checklist

## New conditions and condition reasons for `Cluster` and `AzureCluster` CRs

What's in the box:
- `Cluster` conditions:
    - `ProviderInfrastructureReady`: `True` when `AzureCluster` is ready
    - `NodePoolsReady`: `True` when all node pools are ready
- `Cluster` `Upgrading` condition reasons:
    - `UpgradeCompleted`: used when `Upgrading` is set to `False` because the upgrade has been completed
    - `UpgradeNotStarted`: used when `Upgrading` is set to `False` because the upgrade has not been started
- `Cluster` `Creating` condition reasons:
    - `CreationCompleted`: used when `Creating` is set to `False` because the creation has been completed
    - `ExistingCluster`: used when `Creating` is set to `False` because an older cluster (created without Conditions support) is upgraded to newer release that has conditions.
- Helper struct/functions for setting/getting structured message from `Cluster `Upgrading` condition


fiunchinho (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
```
kopiczko, line 11: Header is missing here. I think it's `### Added`.

nprokopic: Thanks, added it.

PR #599, opened 2020-10-27
aws-annotations-for-update-control

calvix:
towards
https://github.com/giantswarm/roadmap/issues/216
and
https://github.com/giantswarm/roadmap/issues/217

We want to give a customer/SEs and option to configure the update policy of ASG per a cluster, more specific the control over MaxBatchSize and PauseTime in ASG UpdatePolicy - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html

The naming is just something I came up with, so feel free to suggest better names

Its AWS specific

## Checklist

### DIFF
--- a/pkg/annotation/aws.go
+++ b/pkg/annotation/aws.go
@@ -2,8 +2,13 @@ package annotation
 
 // AWSUpdateMaxBatchSize is the aws update annotation used for configuring
 // maximum batch size for instances during ASG update.
+// The value can be either a whole number specifying the number of instances
+// or a percentage of total instances as decimal number ie: `0.30` for 30%.
+// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate-maxbatchsize
 const AWSUpdateMaxBatchSize = "aws.giantswarm.io/update-max-batch-size"
 
 // AWSUpdatePauseTime is the aws update annotation used for configuring
 // time pause between rolling a single batch during ASG update.
+// The value must be in ISO_8601 format.
+// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate-pausetime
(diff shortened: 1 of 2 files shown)


pkg/annotation/aws.go
```diff
@@ -0,0 +1,14 @@
+package annotation
+
+// AWSUpdateMaxBatchSize is the aws update annotation used for configuring
+// maximum batch size for instances during ASG update.
+// The value can be either a whole number specifying the number of instances
+// or a percentage of total instances as decimal number ie: `0.30` for 30%.
```
marians, line 6:
```suggestion
// or a percentage of total instances as decimal number ie: `0.3` for 30%.
```

pkg/annotation/aws.go
```diff
@@ -0,0 +1,14 @@
...
+// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate-maxbatchsize
+const AWSUpdateMaxBatchSize = "aws.giantswarm.io/update-max-batch-size"
+
+// AWSUpdatePauseTime is the aws update annotation used for configuring
+// time pause between rolling a single batch during ASG update.
+// The value must be in ISO_8601 format.
```
marians, line 12: Since the ISO 8601 standard is long and the duration format is not exactly intuitive, I would definitely give an example here that might work for some people.

```suggestion
// The value must be in ISO_8601 format, e. g. "PT5M" for five minutes.
```

marians:
Naming looks good to me. I made a suggestion how to improve the documentation.

marians (approved):

pipo02mix:
Where are going to be these annotations? In AWSMachineDeployment CR would make more flexible the rolling configuration, so customers can have aggressive values for one node pool meanwhile they set more conservative values for other node pool.

calvix:
> Where are going to be these annotations? In AWSMachineDeployment CR would make more flexible the rolling configuration, so customers can have aggressive values for one node pool meanwhile they set more conservative values for other node pool.

the first idea is cluster CR to have this but adding machine deployment specific should be quite easy one all this is on place (machine deployment value would override over cluster if its set)

PR #602, opened 2020-10-30
Add catalog and k8s labels and notes annotation

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/11656

Moves metadata from app-operator so it can be reused in app-admission-controller.

## Checklist

### DIFF
--- a/pkg/annotation/general.go
+++ b/pkg/annotation/general.go
@@ -8,7 +8,7 @@ package annotation
 const Docs = "giantswarm.io/docs"
 
 // Notes is for informational messages for resources managed by operators. Such
-// as whether they resources should be edited.
+// as whether the resource may or may not be edited.
 const Notes = "giantswarm.io/notes"
 
 // ReleaseNotesURL defines where to find release notes about the CR at hand.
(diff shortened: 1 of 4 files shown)


pkg/label/kubernetes.go
```diff
@@ -0,0 +1,11 @@
+package label
+
+// AppKubernetesInstance is a unique name identifying an instance of an app.
+const AppKubernetesInstance = "app.kubernetes.io/name"
```
pipo02mix, line 4:
```suggestion
const AppKubernetesInstance = "app.kubernetes.io/instance"
```

rossf7: @pipo02mix Thanks missed the copy pasta bug.

pkg/annotation/general.go
```diff
@@ -7,6 +7,10 @@ package annotation
 // the system and how they relate to each other.
 const Docs = "giantswarm.io/docs"
 
+// Notes is for informational messages for resources managed by operators. Such
+// as whether they resources should be edited.
```
oponder, line 11:
```suggestion
// as whether the resource may or may not be edited.
```

oponder (approved):
LGTM but a little comment suggestion

pkg/annotation/general.go
rossf7: That's much nicer, thanks.

pipo02mix (approved):
LGTM

PR #603, opened 2020-10-30
Add subnet and metadata annotations

paurosello:
## Checklist

### DIFF
--- a/pkg/annotation/aws.go
+++ b/pkg/annotation/aws.go
@@ -22,7 +22,6 @@ const AWSUpdatePauseTime = "alpha.aws.giantswarm.io/update-pause-time"
 // requests. In this state, retrieving the IAM role credentials always returns the version 2.0 credentials; the
 // version 1.0 credentials are not available.
 // Default: 'optional'
-// The value can be either "optional" or "required". Default is "optional".
 // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html#cfn-ec2-launchtemplate-launchtemplatedata-metadataoptions-httptokens
 const AWSMetadataV2 = "alpha.aws.giantswarm.io/metadata-v2"
 
--- a/pkg/annotation/aws.go
+++ b/pkg/annotation/aws.go
@@ -15,8 +15,8 @@ const AWSUpdatePauseTime = "alpha.aws.giantswarm.io/update-pause-time"
 
 // AWSMetadataV2 is the aws update annotation used for configuring
 // the state of token usage for your instance metadata requests.
-// The value can be either optional or required.
-// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html#cfn-ec2-launchtemplate-launchtemplatedata-metadataoptions-httpendpoint
+// The value can be either "optional" or "required". Default is "optional".
+// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html#cfn-ec2-launchtemplate-launchtemplatedata-metadataoptions-httptokens
 const AWSMetadataV2 = "alpha.aws.giantswarm.io/metadata-v2"
 
 // AWSSubnetSize is the aws update annotation used for configuring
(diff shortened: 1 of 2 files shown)


calvix (approved):

xh3b4sd (approved):

pipo02mix (approved):
LGTM nice

pkg/annotation/aws.go
```diff
@@ -12,3 +12,14 @@ const AWSUpdateMaxBatchSize = "alpha.aws.giantswarm.io/update-max-batch-size"
 // The value must be in ISO 8601 duration format, e. g. "PT5M" for five minutes or "PT10S" for 10 seconds.
 // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate-pausetime
 const AWSUpdatePauseTime = "alpha.aws.giantswarm.io/update-pause-time"
+
+// AWSMetadataV2 is the aws update annotation used for configuring
+// the state of token usage for your instance metadata requests.
+// The value can be either optional or required.
+// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html#cfn-ec2-launchtemplate-launchtemplatedata-metadataoptions-httpendpoint
+const AWSMetadataV2 = "alpha.aws.giantswarm.io/metadata-v2"
```
marians, line 20: I don't understand the description completely.

> The value can be either optional or required.

Does that mean it can be either the string`"optional"` or `"required"`? If yes, please set the values in quotes.

The docs link points to a setting named `HttpEndpoint` which allows the value `disabled` or `enabled`. I don't understand how this explains the meaning of this annotation.

Should that link point to `HttpTokens` instead? Then please use the URL https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html#cfn-ec2-launchtemplate-launchtemplatedata-metadataoptions-httptokens

marians (changes requested):

pkg/annotation/aws.go
paurosello: Yep, could get the URL correctly previously but that looks better, thanks for pointing out

pkg/annotation/aws.go
```diff
@@ -12,3 +12,14 @@ const AWSUpdateMaxBatchSize = "alpha.aws.giantswarm.io/update-max-batch-size"
 // The value must be in ISO 8601 duration format, e. g. "PT5M" for five minutes or "PT10S" for 10 seconds.
 // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate-pausetime
 const AWSUpdatePauseTime = "alpha.aws.giantswarm.io/update-pause-time"
+
+// AWSMetadataV2 is the aws update annotation used for configuring
+// the state of token usage for your instance metadata requests.
```
marians, line 17: I still don't understand this sentence:

> AWSMetadataV2 is the aws update annotation used for configuring the state of token usage for your instance metadata requests.

Maybe just copy and adapt the entire description from upstream? Lines are for free.

```suggestion
// AWSMetadataV2 configures token usage for your instance metadata requests.
// If the value is 'optional', you can choose to retrieve instance metadata with or without a signed token
// header on your request. If you retrieve the IAM role credentials without a token, the version 1.0 role
// credentials are returned. If you retrieve the IAM role credentials using a valid signed token, the version
// 2.0 role credentials are returned.
// If the state is 'required', you must send a signed token header with any instance metadata retrieval
// requests. In this state, retrieving the IAM role credentials always returns the version 2.0 credentials; the
// version 1.0 credentials are not available.
// Default: 'optional'
```

pkg/annotation/aws.go
```diff
@@ -12,3 +12,14 @@ const AWSUpdateMaxBatchSize = "alpha.aws.giantswarm.io/update-max-batch-size"
...
+// The value can be either "optional" or "required". Default is "optional".
```
marians, line 18: With my suggestion above, line 18 can be dropped.

pkg/annotation/aws.go
```diff
@@ -13,8 +13,15 @@ const AWSUpdateMaxBatchSize = "alpha.aws.giantswarm.io/update-max-batch-size"
...
+// AWSMetadataV2 configures token usage for your AWS EC2 instance metadata requests.
+// If the value is 'optional', you can choose to retrieve instance metadata with or without a signed token
+// header on your request. If you retrieve the IAM role credentials without a token, the version 1.0 role
+// credentials are returned. If you retrieve the IAM role credentials using a valid signed token, the version
+// 2.0 role credentials are returned.
+// If the state is 'required', you must send a signed token header with any instance metadata retrieval
+// requests. In this state, retrieving the IAM role credentials always returns the version 2.0 credentials; the
+// version 1.0 credentials are not available.
+// Default: 'optional'
 // The value can be either "optional" or "required". Default is "optional".
```
marians: As I wrote, this line is redundant now

paurosello: yes, it was removed

marians: Thanks! I looked at an outdated diff without noticing.

PR #608, opened 2020-11-03
Add release.giantswarm.io/last-deployed-version annotation

nprokopic:
Add `release.giantswarm.io/last-deployed-version` annotation that is put into Cluster CR to define which Giant Swarm release version was last successfully deployed during cluster creation or upgrade.

Having this annotation will make setting of `Cluster` `Creating` and `Upgrading` conditions update implementation more reliable and more straightforward.

## Checklist


pipo02mix (approved):
LGTM

PR #615, opened 2020-11-10
Adding restrictions field into AppCatalogEntry CRD

tomahawk28:
## Checklist

Toward https://github.com/giantswarm/giantswarm/issues/13973


pkg/apis/application/v1alpha1/app_catalog_entry_types.go
```diff
@@ -101,6 +105,18 @@ type AppCatalogEntrySpecChart struct {
 	Icon string `json:"icon,omitempty"`
 }
 
+// +k8s:openapi-gen=true
+type AppCatalogEntrySpecRestrictions struct {
+	// ClusterSingleton is a flag for whether this chart can be installed at most once per cluster.
```
rossf7, line 110:
```suggestion
	// ClusterSingleton is a flag for whether this app can be installed at most once per cluster.
```

This is the text from the spec but I wonder whether we should use app rather than chart here?

pkg/apis/application/v1alpha1/app_catalog_entry_types.go
```diff
@@ -101,6 +105,18 @@ type AppCatalogEntrySpecChart struct {
...
+	ClusterSingleton bool `json:"cluster_singleton"`
+	// NamespaceSingleton is a flag for whether this chart can be installed at most once per namespace
```
rossf7, line 112:
```suggestion
	// NamespaceSingleton is a flag for whether this app can be installed at most once per namespace.
```

pkg/apis/application/v1alpha1/app_catalog_entry_types.go
```diff
@@ -101,6 +105,18 @@ type AppCatalogEntrySpecChart struct {
...
+	NamespaceSingleton bool `json:"namespace_singleton"`
+	// FixedNamespace is the namespace which this chart can be installed only.
```
rossf7, line 114:
```suggestion
	// FixedNamespace is the namespace which this app must be installed in.
```

pkg/apis/application/v1alpha1/app_catalog_entry_types.go
```diff
@@ -101,6 +105,18 @@ type AppCatalogEntrySpecChart struct {
...
+	FixedNamespace string `json:"fixed_namespace,omitempty"`
+	// GpuInstances is a flag for whether this chart requires GPU instances to run.
```
rossf7, line 116:
```suggestion
	// GpuInstances is a flag for whether this app requires GPU instances to run.
```

pkg/apis/application/v1alpha1/app_catalog_entry_types.go
```diff
@@ -73,6 +73,10 @@ type AppCatalogEntrySpec struct {
 	// DateUpdated is when this entry was last updated.
 	// e.g. 2020-09-02T09:40:39.223638219Z
 	DateUpdated *metav1.Time `json:"dateUpdated"`
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// Restrictions is metadata from Chart.yaml of the app this entry requires to validate them.
```
rossf7, line 78:
```suggestion
	// Restrictions is metadata from Chart.yaml for this app and is used to validate app CRs.
```

rossf7 (commented):
Looks good but it needs a changelog and also good to hear from SIG UX.

Just some minor stuff. My main query is whether we should use app terminology instead of chart.

pkg/apis/application/v1alpha1/app_catalog_entry_types.go
tomahawk28: Right, `app` is more likely suitable to the word `app catalog Entry`

pkg/apis/application/v1alpha1/app_catalog_entry_types.go
tomahawk28: Nice, 👍

rossf7 (approved):
LGTM

marians:
I see you ticked the `Consider SIG UX feedback` but I don't think anyone from SIG UX has given feedback. That's not how it's supposed to work.

marians:
Some things I have to note:

- The whole `restrictions` attribute is `nullable`, but the three sub attributes `cluster_singleton`, `gpu_instances`, and `namespace_singleton` are required. That does not make sense to me. If `restrictions` is `null` or not given, there must be defaults in place for these settings, right? So they should ideally be optional. Also we should give information regarding what the default is.
- Is (A) a combination of `cluster_singleton: true` and `namespace_singleton: true` possible? Or is it (B) always either one or the other or none? If (B), then I would find it simpler to combine the two into one attribute `singleton` with possible values `"CLUSTER"` and `"NAMESPACE"` and a default value `null`.

tomahawk28:
> I see you ticked the Consider SIG UX feedback but I don't think anyone from SIG UX has given feedback. 

I'm sorry I was in rush, will continue to get your feedback though in this thread.

tomahawk28:
> If restrictions is null or not given, there must be defaults in place for these settings, right? So they should ideally be optional. Also we should give information regarding what the default is.

You are right, It should all optional if `restrictions` is optional as well. 

I think (A) combination of settings is right.  We should enforce the app in either per namespace or per cluster. cc> @piontec Can you double-check?

piontec:
Sorry for late response. Yes, it makes sense to make them all optional.

PR #620, opened 2020-11-13
Graduate Organization CRD

fiunchinho:
I want to graduate the `Organization` CRD from `v1beta1` to `v1`. The weird thing is that the YAML file with the CRD definition in `config/crd/v1/security.giantswarm.io_organizations.yaml` was already defining it as `v1`. I don't understand how this can be possible if the go object was defined using `v1beta1`. Actually, running `make` doesn't update any files.

To finish the graduation, after merging this we need to release a new version of `apiextensions`, and use that new version from clients using the `Organization` CRD like `opsctl`, `organization-operator` or some Celestial components. Is there anything else to do to complete the graduation?

## Checklist

- [X] If adding a new type, ensure that it is added to the [CRD unit tests](https://github.com/giantswarm/apiextensions/blob/d78aba91d578d56ef49f58e336035d35edc4e1b0/pkg/crd/crd_test.go#L9).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Graduate `Organization` CRD to v1.
+
 ### Added
 
 - Add `Restrictions` metadata to `AppCatalogEntry` CRD.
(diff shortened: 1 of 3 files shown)


kubasobon (approved):
LGTM.

tfussell:
The CRD loading function is just for our convenience and to help us keep track of whether we want to use v1 or v1beta1 CRD API for each. Controller-gen only looks at the CR structs and their comments to generate CRDs. We could technically drop all of these "New*CRD" functions and just use crd.LoadV1 directly in opsctl if desired. Once everything is graduated, we will drop all of the v1beta1 stuff here.

fiunchinho:
Ok, so it's not really necessary. I'll close this then.

PR #624, opened 2020-11-20
Add catalog reference to apps in releases

MarcelMue:
## Checklist


stone-z (approved):

MarcelMue:
Tested as working on ginger

config/crd/v1/release.giantswarm.io_releases.yaml
```diff
@@ -71,6 +71,11 @@ spec:
                 description: Apps describes apps used in this release.
                 items:
                   properties:
+                    catalog:
+                      default: default
```
rossf7, line 75: Change makes sense to me. Way off topic but I have naming regret. The catalog should be called `preinstalled` 🤦‍♂️ but naming is hard.

rossf7 (approved):
LGTM

config/crd/v1/release.giantswarm.io_releases.yaml
MarcelMue: I can relate :D

PR #631, opened 2020-11-30
make nodepool AZs optional

anvddriesch:
Towards: https://github.com/giantswarm/giantswarm/issues/14429

We want to make the nodepool AZs optional so they can be defaulted. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Make `availabilityZones` attribute in `AWSMachineDeployment` optional.
+
 ## [3.9.0] - 2020-11-24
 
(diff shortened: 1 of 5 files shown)


njuettner (approved):
lgtm

kopiczko:
Changelog is missing.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Make `availabilityZones` attribute in `AWSMachineDeployment` optional.
+
```
marians, line 15: Would be great to add information regarding the default. As a user, when I don't set this attribute, what will happen?

anvddriesch: Good point! :) This will be defaulted by the `aws-admission-controller` if it is not set.

PR #636, opened 2020-12-03
Add annotations for apps

tomahawk28:
`LatestConfigMapVersion`, `LatestSecretVersion` is needed by two operators at the moment. 
It makes sense to move these to apiextensions. 

## Checklist


rossf7 (approved):
LGTM, thx!

PR #653, opened 2021-01-05
Change 'tenant cluster' to 'workload cluster'

marians:
Towards https://github.com/giantswarm/giantswarm/issues/14961

Only affects comments.

## Checklist


config/crd/v1/provider.giantswarm.io_azureconfigs.yaml
```diff
@@ -305,7 +305,7 @@ spec:
                             type: object
                           kubeProxy:
                             description: ClusterKubernetesNetworkSetupKubeProxy describes
-                              values passed to the kube-proxy running in a tenant
+                              values passed to the kube-proxy running in a workload
```
MarcelMue, line 308: Not sure where this is coming from but why is it not `workload cluster` here?
I can not find where this was generated from in this PR

marians: It looks all right to me. The work `cluster` is in the next line. Please check again.

MarcelMue (approved):

PR #657, opened 2021-01-12
Introduce ability to skip CRD installation with App/Chart Operator

yasn77:
Towards giantswarm/giantswarm#13746

Add ability to define if the installation of CRD should be skipped.

## Checklist


pkg/apis/application/v1alpha1/chart_types.go
```diff
@@ -76,6 +76,8 @@ type ChartSpec struct {
 
 // +k8s:openapi-gen=true
 type ChartSpecConfig struct {
+	// Whether to install CRDs that might be supplied with the chart
+	SkipCRDInstall bool `json:"skipCrdInstall"`
```
tomahawk28, line 80: I think `SkipCRD` is enough WDYT?

```suggestion
	SkipCRD bool `json:"skipCrdInstall"`
```

tomahawk28 (commented):
Please execute `make` from the top of the directory and commit auto-generated files.

docs/cr/application.giantswarm.io_v1alpha1_chart.yaml
```diff
@@ -16,6 +16,7 @@ spec:
       name: f2def-chart-values
       namespace: f2def
       resourceVersion: ""
+    skipCrdInstall: false
```
rossf7, line 19:
```suggestion
  install:
    skipCRDs: false
```

@yasn77 So far we've done these type of config extensions by adding annotations like for skipping forced upgrades. But I think we should start extending the CRD which is fine as long as we don't make breaking changes.

I compared this with the helmrelease CRD as I was interested to see how Flux do this. I like how they structure this. As its aligned with the helm option.  WDYT?

https://toolkit.fluxcd.io/components/helm/helmreleases/

Also I think we need to make both install and skipCRDs optional. As in most cases we don't need to set this option.

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -85,6 +85,8 @@ type AppSpec struct {
 
 // +k8s:openapi-gen=true
 type AppSpecConfig struct {
+	// Whether to install CRDs that might be supplied with the chart
+	SkipCRDInstall bool `json:"skipCrdInstall"`
```
rossf7, line 89:
```suggestion
	SkipCRDInstall bool `json:"skipCrdInstall,omitempty"`
```

Adding omitempty will make the field optional as per my other comment.

docs/cr/application.giantswarm.io_v1alpha1_chart.yaml
yasn77: Yeh I like adding `Install.SkipCRDs`, it gives us an option of adding additional install time options such as skipping hooks (similar to fluxcd approach)

yasn77:
> Please execute `make` from the top of the directory and commit auto-generated files.

Should this be a pre-commit step then?

pkg/apis/application/v1alpha1/app_types.go
yasn77: I am not sure about making it optional.... Doesn't a zero value boolean in go default to `false`? which is exactly what we want. It would conditional logic in the code similar right? something like...
```golang
if cr.Install.SkipCRDs {
...
}
```
Rather trying to test for the existence of the bool then test for true/false

pkg/apis/application/v1alpha1/chart_types.go
yasn77: I am going to go with Ross suggestion and add an `Install` struct. Thanks 👍🏽

pkg/apis/application/v1alpha1/app_types.go
rossf7: Yes in the operator the boolean will default to false. But one thing to watch is the install struct as it could be nil.

The omitempty is about keeping the app CRs cleaner when they are viewed in kubectl etc. So the setting will only appear if the user has requested CRDs to be skipped.

docs/cr/application.giantswarm.io_v1alpha1_chart.yaml
```diff
@@ -16,6 +16,9 @@ spec:
...
```
tomahawk28, line 19: I think this should be removed. right?

yasn77: Yep... removing

config/crd/v1/application.giantswarm.io_apps.yaml
```diff
@@ -94,6 +94,17 @@ spec:
                     - namespace
                     type: object
                 type: object
+              install:
+                description: Install is the config sed to deploy the app and is passed
```
pipo02mix, line 98: I dont get the phrase, do you meant `set`?

pipo02mix:
```suggestion
                description: Install is the config used to deploy the app and is passed
                to Helm.
```

pipo02mix: copy from other block ^^

yasn77: Yeh, was a typo.. Meant to be 'used'... Fixed now

pipo02mix:
I tend to think just `crds`it is a better naming. Since it is under `install` property I would rather just use a name and not verb+name. Trying to keep consistency, WDYT?

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -65,6 +65,10 @@ type AppSpec struct {
 	// +nullable
 	// Config is the config to be applied when the app is deployed.
 	Config AppSpecConfig `json:"config,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// Install is the config used to deploy the app and is passed to chart operator.
```
rossf7, line 70:
```suggestion
	// Install is the config used when installing the app.
```

chart-operator is an implementation detail and may even change to the Flux helm-controller in the future. So can we drop that part?

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -97,6 +101,14 @@ type AppSpecConfig struct {
 	Secret AppSpecConfigSecret `json:"secret,omitempty"`
 }
 
+// +k8s:openapi-gen=true
+type AppSpecInstall struct {
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// SkipCRDs decides whether to install CRDs that might be supplied with the chart
+	SkipCRDs bool `json:"skipCrds,omitempty"`
```
rossf7, line 109:
```suggestion
	SkipCRDs bool `json:"skipCRDs,omitempty"`
```

The JSON field should match the Go struct apart from the first char.

rossf7 (commented):
> I tend to think just crds it is a better naming. Since it is under install property I would rather just use a name and not verb+name.

@pipo02mix I also prefer keeping the CRs minimal but without the skip I think there isn't enough context. I also like that its aligned with the helm flag naming. WDYT?

```
helm install -h | grep skip-crds
      --skip-crds                    if set, no CRDs will be installed. By default, CRDs are installed if not already present
```

yasn77:
> I tend to think just `crds`it is a better naming. Since it is under `install` property I would rather just use a name and not verb+name. Trying to keep consistency, WDYT?

@pipo02mix, In general, I prefer shorter things :) and yeh, it's a bit much when verb+name is already implied. However in this case, it matches with then what is actually passed to helm client.

I know that is a bit of an implementation detail, but we are not likely to move away from Helm any time soon and I don't expect Helm to drop `--skip-crd` in Helm 3. So I feel from a UX point of view, it's less of a cognitive load to use `skipCrd`, as the user (who most likely already is familiar with Helm) will know immediately what the functionality of that config will do.  It matches very closely with what they already know about Helm install.

marians:
@yasn77 As a reviewer, I'd find a description of the goal of this PR in the main title and comment helpful. Then I could look at the changes and check whether they make sense with respect to that goal.

yasn77:
> @yasn77 As a reviewer, I'd find a description of the goal of this PR in the main title and comment helpful. Then I could look at the changes and check whether they make sense with respect to that goal.

Your right... My bad. I should have added more detail and not just used the branch name. I was experimenting with a new IDE to create the PR and missed what I usually take for granted.

pkg/apis/application/v1alpha1/chart_types.go
```diff
@@ -74,8 +78,18 @@ type ChartSpec struct {
 	Version string `json:"version"`
 }
 
+// +k8s:openapi-gen=true
+type ChartSpecInstall struct {
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// SkipCRDs decides whether to install CRDs that might be supplied with the chart
+	SkipCRDs bool `json:"skipCrds,omitempty"`
```
marians, line 86: There is a bit of a chance of a misinterpretation here. "skip" to me indicates that a CRD _will not_ be installed. However the description seems to indicate the opposite.

This could be resolved by making the description more explicit. Or by renaming the field.

Also it should be mentioned what the default is (if not set). This may be obvious in a Go context, but not in a CRD context.

```suggestion
	// SkipCRDs when true decides that CRDs which are supplied with the chart are not installed. Default: false.
	SkipCRDs bool `json:"skipCrds,omitempty"`
```

or

```suggestion
	// InstallCRDs decides that CRDs which are supplied with the chart will be installed (default). Set to false to omit CRD installation.
	InstallCRDs bool `json:"installCrds,omitempty"`
```

pkg/apis/application/v1alpha1/chart_types.go
marians: Also see my comment https://github.com/giantswarm/apiextensions/pull/657/files#r558190783 which I think also applies here.

marians (approved):

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -97,6 +101,14 @@ type AppSpecConfig struct {
 	Secret AppSpecConfigSecret `json:"secret,omitempty"`
 }
 
+// +k8s:openapi-gen=true
+type AppSpecInstall struct {
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// SkipCRDs when true decides that CRDs which are supplied with the chart are not installed. Default: false.
+	SkipCRDs bool `json:"skipCrds,omitempty"`
```
rossf7, line 109:
```suggestion
	SkipCRDs bool `json:"skipCRDs,omitempty"`
```

@yasn77 Could we make the JSON tag upper case as well? We don't have many abbreviations in CRDs but I think they're usually upper case.

rossf7 (approved):
LGTM just one query on the JSON tag

PR #662, opened 2021-01-18
Fix name for AppCatalogEntry example.

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/15025

## Checklist


pkg/apis/application/v1alpha1/app_catalog_entry_types_test.go
```diff
@@ -44,7 +44,7 @@ func Test_GenerateAppCatalogEntryYAML(t *testing.T) {
 	}{
 		{
 			category: "cr",
-			name:     fmt.Sprintf("%s_%s_AppCatalogEntry.yaml", group, version),
+			name:     fmt.Sprintf("%s_%s_appcatalogentry.yaml", group, version),
```
rossf7, line 47: The example is there but I think the name is wrong.

@marians Is any other wiring needed to make the example show up here?

https://docs.giantswarm.io/reference/management-api/appcatalogentries.application.giantswarm.io/

pipo02mix (approved):
lgtm

PR #664, opened 2021-01-20
add-aws-cni-annotations

calvix:
towards https://github.com/giantswarm/giantswarm/issues/14441

## Checklist

### DIFF
--- a/pkg/annotation/aws.go
+++ b/pkg/annotation/aws.go
@@ -1,14 +1,14 @@
 package annotation
 
-// AWSCNIMinimumIpTarget is the annotation to configure value for MINIMUM_IP_TARGET for AWS CNI
+// AWSCNIMinimumIPTarget is the annotation to configure value for MINIMUM_IP_TARGET for AWS CNI
 // see https://github.com/aws/amazon-vpc-cni-k8s#cni-configuration-variables
 // and https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/eni-and-ip-target.md
-const AWSCNIMinimumIpTarget = "alpha.aws.giantswarm.io/cni-minimum-ip-target"
+const AWSCNIMinimumIPTarget = "alpha.aws.giantswarm.io/cni-minimum-ip-target"
 
-// AWSCNIWarmIpTarget is the annotation to configure value for WARM_IP_TARGET for AWS CNI
+// AWSCNIWarmIPTarget is the annotation to configure value for WARM_IP_TARGET for AWS CNI
 // see https://github.com/aws/amazon-vpc-cni-k8s#cni-configuration-variables
 // and https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/eni-and-ip-target.md
-const AWSCNIWarmIpTarget = "alpha.aws.giantswarm.io/cni-warm-ip-target"
+const AWSCNIWarmIPTarget = "alpha.aws.giantswarm.io/cni-warm-ip-target"
 
 // AWSUpdateMaxBatchSize is the aws update annotation used for configuring
 // maximum batch size for instances during ASG update.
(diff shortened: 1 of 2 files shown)


pipo02mix (approved):
LGTM but I would use IP as it is an acronym too

paurosello (approved):
LGTM

njuettner (approved):

pkg/annotation/aws.go
```diff
@@ -1,5 +1,15 @@
...
 
+// AWSCNIMinimumIPTarget is the annotation to configure value for MINIMUM_IP_TARGET for AWS CNI
+// see https://github.com/aws/amazon-vpc-cni-k8s#cni-configuration-variables
+// and https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/eni-and-ip-target.md
+const AWSCNIMinimumIPTarget = "alpha.aws.giantswarm.io/cni-minimum-ip-target"
+
+// AWSCNIWarmIPTarget is the annotation to configure value for WARM_IP_TARGET for AWS CNI
+// see https://github.com/aws/amazon-vpc-cni-k8s#cni-configuration-variables
+// and https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/eni-and-ip-target.md
+const AWSCNIWarmIPTarget = "alpha.aws.giantswarm.io/cni-warm-ip-target"
```
xh3b4sd, line 11: What about this structure? I feel like this is a more appropriate structure for the domain namespace. 

```suggestion
const AWSCNIWarmIPTarget = "cni.alpha.aws.giantswarm.io/warm-ip-target"
```

calvix: @giantswarm/sig-ux @marians  WDYT?

pkg/annotation/aws.go
```diff
@@ -1,5 +1,15 @@
 package annotation
 
+// AWSCNIMinimumIPTarget is the annotation to configure value for MINIMUM_IP_TARGET for AWS CNI
```
marians, line 3:
```suggestion
// AWSCNIMinimumIPTarget is an annotation to configure the value for MINIMUM_IP_TARGET for AWS CNI
```

pkg/annotation/aws.go
```diff
@@ -1,5 +1,15 @@
...
+// see https://github.com/aws/amazon-vpc-cni-k8s#cni-configuration-variables
+// and https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/eni-and-ip-target.md
+const AWSCNIMinimumIPTarget = "alpha.aws.giantswarm.io/cni-minimum-ip-target"
+
+// AWSCNIWarmIPTarget is the annotation to configure value for WARM_IP_TARGET for AWS CNI
```
marians, line 8:
```suggestion
// AWSCNIWarmIPTarget is an annotation to configure the value for WARM_IP_TARGET for AWS CNI
```

pkg/annotation/aws.go
marians: I don't think we have system regarding annotation grouping, and I can't judge if we need this group.

Just noting, the prefix could be either of

- `cni.alpha.aws.giantswarm.io/`
- `alpha.cni.aws.giantswarm.io/`

calvix: `alpha.cni.aws.giantswarm.io/` would make more sense for me

calvix: but uncertain since we don't use it anywhere else so far

marians: Then I vote for `alpha.cni.aws.giantswarm.io/`, too

calvix:
RFR with the changes to the annotation structure

xh3b4sd (approved):

PR #667, opened 2021-01-21
Set LastHeartbeatTime when creating Conditions for DrainerConfigs

stone-z:
It is invalid to create a `Condition` with a `nil` `LastHeartbeatTime`, so this adds one. I think this was just missed when the equivalent timestamp for `LastTransitionTime` was added.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Set a `LastHeartbeatTime` when creating new `Condition`s for `DrainerConfig`s.
+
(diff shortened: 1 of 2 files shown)


MarcelMue (approved):
Makes sense to me - can we ask some others for review? Maybe @kopiczko or @tuommaki can see if this makes sense.

kopiczko (approved):
https://gigantic.slack.com/archives/C2MS2VB37/p1574100113071300?thread_ts=1574095202.071100&cid=C2MS2VB37

https://github.com/kubernetes/community/blob/712590c108bd4533b80e8f2753cadaa617d9bdf2/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

So AIUI whenever `LastTransitionTime` is updated `LastHeartbeatTime` should be updated by definition.

PR #687, opened 2021-02-18
Allow to add the same version again in the cluster status to support rollbacks

anvddriesch:
Related: https://github.com/giantswarm/giantswarm/issues/15468
We now support rolling back upgrades, which means that we want to be able to see the same version again in the status history.

## Checklist


calvix (approved):

xh3b4sd (approved):

anvddriesch:
Tested this on `gauss` and it works.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
```
kopiczko, line 10: IMO this is Fixed

anvddriesch: Can we use `Fixed` also for things that were not broken? We only just started supporting the version rollbacks, But I am honestly not sure what to pick (we may have also `Added` the ability to roll back or `Removed` the limitation...)

PR #689, opened 2021-02-23
Add appCatalogEntry into registry

tomahawk28:
We need to add new CRDs into the registry to use it from unittest. 
Otherwise, we get an error below. 
```
    --- FAIL: Test_ValidateMetadataConstraints/case_1:_fixed_namespace_constraint (0.00s)
panic: no kind is registered for the type v1alpha1.AppCatalogEntry in scheme "pkg/runtime/scheme.go:101" [recovered]
	panic: no kind is registered for the type v1alpha1.AppCatalogEntry in scheme "pkg/runtime/scheme.go:101"
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Add `AppCatalogEntry` into the registry.
+
 ## [3.18.2] - 2021-02-18
 
 ### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Add `AppCatalogEntry` into the registry.
```
rossf7, line 12:
```suggestion
- Register `AppCatalogEntry` CRD as a known type.
```

Changelog is fine but what about this?

rossf7 (approved):

CHANGELOG.md
tomahawk28: Much better, Thanks!

PR #692, opened 2021-02-24
Add example tag provider

pipo02mix:
## Checklist

### DIFF
--- a/docs/cr/cluster.x-k8s.io_v1alpha2_cluster.yaml
+++ b/docs/cr/cluster.x-k8s.io_v1alpha2_cluster.yaml
@@ -4,7 +4,7 @@ metadata:
   annotations:
     giantswarm.io/docs: https://pkg.go.dev/sigs.k8s.io/cluster-api/api/v1alpha2?tab=doc#Cluster
   labels:
-    tag.provider.giantswarm.io/Environment: PROD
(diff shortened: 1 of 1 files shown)


pipo02mix:
@marians is this file auto-generated? or I can add it manually?

marians:
Can we make the example all lowercase?

I see no additional value in encouraging users to use mixed case labels. However I see a potential of getting confused when users try to use mixed case labels in selectors and don't know about their case sensitivity.

pipo02mix:
It is copied from an actual customer case. It is up to them choosing the format but I am fine with not encourgint them to use uppercase. I change it

docs/cr/cluster.x-k8s.io_v1alpha2_cluster.yaml
```diff
@@ -3,7 +3,8 @@ kind: Cluster
 metadata:
   annotations:
     giantswarm.io/docs: https://pkg.go.dev/sigs.k8s.io/cluster-api/api/v1alpha2?tab=doc#Cluster
-  creationTimestamp: null
+  labels:
+    tag.provider.giantswarm.io/Environment: PROD
```
pipo02mix, line 7:
```suggestion
    tag.provider.giantswarm.io/environment: prod
```

marians (approved):
Thanks!

PR #697, opened 2021-03-10
Add CAPI labels that will be used to filter CR's by version

fiunchinho:
We need to filter which CRs are watched by each deployed CAPI controller. The filtering is done via label.

Kubernetes label max length is 63 chars, we are safe.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add Cluster API labels that will be used to filter CR's by version.
+
 ### Changed
 
 - Terminology update to use 'workload cluster release' consistently.
--- a/pkg/label/version.go
+++ b/pkg/label/version.go
@@ -30,6 +30,30 @@ const AWSOperatorVersion = "aws-operator.giantswarm.io/version"
 // that there is an azure-operator release v4.1.0.
 const AzureOperatorVersion = "azure-operator.giantswarm.io/version"
 
+// CAPIVersion is the version label put into provider independent CRs
+// to define which cluster-api-core version should reconcile the given CR.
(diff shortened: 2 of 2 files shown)


fiunchinho:
We are experimenting with this "upgrade operator" that takes care of updating the labels for all the CR's. It knows which version to use because it reads that from the `Release` CR referenced by the `Cluster` CR release label. So basically, customers only need to use the release label in the Cluster CR, and the operator takes care of the rest.

kubasobon:
Bonus question: what's our approach to `matchPolicy` ([documentation](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy))? Are labels granular enough, or do we need to redirect webhook calls based on CR version (`v1alphaX`).

fiunchinho:
> enough

I don't really know how this works yet. But I'd say that a specific capi component release is coupled to a api version of the CRD, so filtering by label would be enough?

mig4:
Stupid question, perhaps I'm confusing something, but according to https://github.com/kubernetes-sigs/cluster-api/pull/4119 shouldn't the label key always be `cluster.x-k8s.io/watch-filter`?

fiunchinho:
> Stupid question, perhaps I'm confusing something, but according to [kubernetes-sigs/cluster-api#4119](https://github.com/kubernetes-sigs/cluster-api/pull/4119) shouldn't the label key always be `cluster.x-k8s.io/watch-filter`?

Oh so it's always the same label? I thought it was configurable. Closing this then.

mig4:
> > Stupid question, perhaps I'm confusing something, but according to [kubernetes-sigs/cluster-api#4119](https://github.com/kubernetes-sigs/cluster-api/pull/4119) shouldn't the label key always be `cluster.x-k8s.io/watch-filter`?
> 
> Oh so it's always the same label? I thought it was configurable. Closing this then.

Double-check if I'm right, but yeah it seems that way.

PR #698, opened 2021-03-12
Refactor ID generation to block ids starting with digits

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/16193

This PR makes generated IDs not be able to start with digits. It also simplifies the previous implementation.

`api` and `kubectl-gs` should be updated to use this new generator.

### DIFF
--- a/pkg/id/id_test.go
+++ b/pkg/id/id_test.go
@@ -24,7 +24,7 @@ func TestGenerate(t *testing.T) {
 			}
 
 			if !unicode.IsLetter(idChars[0]) {
-				t.Fatalf("generated id starts does not start with a letter: %q", id)
+				t.Fatalf("generated id does not start with a letter: %q", id)
 			}
 
 			if lettersOnlyRegexp.MatchString(id) {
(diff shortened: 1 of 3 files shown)


pkg/id/id_test.go
```diff
@@ -0,0 +1,43 @@
...
+	validCharsRegexp := regexp.MustCompile("^[02-9a-km-z]")
+	lettersOnlyRegexp := regexp.MustCompile("^[a-z]+$")
+	digitsOnlyRegexp := regexp.MustCompile("^[0-9]+$")
+
+	for i := 0; i < attempts; i++ {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			id := Generate()
+			idChars := []rune(id)
+
+			if !validCharsRegexp.MatchString(id) {
```
whites11, line 22: Am I wrong or you're not testing the leading digit case here?

axbarsan: That is tested lower down https://github.com/giantswarm/apiextensions/blob/68d61e975367a19b5d266262b264b6b7d47eec31/pkg/id/id_test.go#L26

It will panic if the first character is not a letter

whites11 (approved):

PR #699, opened 2021-03-15
Add label ui.giantswarm.io/display

marians:
Towards https://github.com/giantswarm/roadmap/issues/281

## Checklist

### DIFF
--- a/pkg/label/ui.go
+++ b/pkg/label/ui.go
@@ -1,0 +1,11 @@
+package label
+
+// I the ui.giantswarm.io label namespace we define labels that
+// exist in order to affect the user experience in user
+// interfaces, like the Giant Swarm web UI or CLIs.
+
+// Affects whether or not a resource is intended for display in
+// a user interface. For example, it can be used to hide
+// irrelevant system roles from users by default. The value can
+// either be "true" or "false".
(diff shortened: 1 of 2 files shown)


pkg/label/ui.go
```diff
@@ -0,0 +1,11 @@
+package label
+
+// I the ui.giantswarm.io label namespace we define labels that
```
ubergesundheit, line 3:
```suggestion
// In the ui.giantswarm.io label namespace we define labels that
```

pkg/label/ui.go
```diff
@@ -0,0 +1,11 @@
...
+// exist in order to affect the user experience in user
+// interfaces, like the Giant Swarm web UI or CLIs.
+
+// Affects whether or not a resource is intended for display in
+// a user interface. For example, it can be used to hide
```
ubergesundheit, line 8:
```suggestion
// an user interface. For example, it can be used to hide
```

ubergesundheit (approved):

axbarsan (approved):

pkg/label/ui.go
marians: A Google search tells me that "a user interface" is correct

alex-dabija (approved):

PR #700, opened 2021-03-15
Adding spec for target namespace in app CRD

tomahawk28:
~Toward https://github.com/giantswarm/giantswarm/issues/15375~
Toward https://github.com/giantswarm/giantswarm/issues/16209

Adding `NamespaceConfig` to the app CRD spec so we can add labels/annotates to the target namespace.  

## Checklist


pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -80,6 +80,10 @@ type AppSpec struct {
 	Namespace string `json:"namespace"`
 	// +kubebuilder:validation:Optional
 	// +nullable
+	// NamespaceConfig is the namespace config to be applied to the target namespace when the app is deployed.
+	NamespaceConfig AppSpecNamespaceConfig `json:"namespaceConfig,omitempty"`
```
tomahawk28, line 84: I'm making NamespaceConfig as nullable since it's not always necessary.

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -160,6 +164,18 @@ type AppSpecKubeConfigSecret struct {
 	Namespace string `json:"namespace"`
 }
 
+// +k8s:openapi-gen=true
+type AppSpecNamespaceConfig struct {
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// Labels references a string map the app is applying to the label of target namespace.
+	Labels map[string]string `json:"labels,omitempty"`
```
tomahawk28, line 172: The string map type is the right match to the namespace configs.

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -160,6 +164,18 @@ type AppSpecKubeConfigSecret struct {
 	Namespace string `json:"namespace"`
 }
 
+// +k8s:openapi-gen=true
+type AppSpecNamespaceConfig struct {
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// Labels references a string map the app is applying to the label of target namespace.
```
rossf7, line 171:
```suggestion
	// Labels is a string map of labels to apply to the target namespace.
```

pkg/apis/application/v1alpha1/app_types.go
rossf7: Nice just suggested some edits for the comments. I think the fields should be in alpha order.

rossf7 (commented):
Direction is good. Just some minor stuff.

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -160,6 +164,18 @@ type AppSpecKubeConfigSecret struct {
...
+	Labels map[string]string `json:"labels,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// Annotations references a string map the app is applying to the annotation of target namespace.
```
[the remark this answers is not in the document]
tomahawk28: Great, it looks better.

tomahawk28:
@giantswarm/sig-ux would you please one of you can review this? Thanks!

pkg/apis/application/v1alpha1/app_types_test.go
```diff
@@ -98,6 +98,11 @@ func newAppExampleCR() *App {
 				Namespace: "f2def",
 			},
 		},
+		NamespaceConfig: AppSpecNamespaceConfig{
+			Labels: map[string]string{
```
rossf7, line 102:
```suggestion
			Annotations: map[string]string{
```

I like the real world example but it should be an annotation.

rossf7 (approved):
LGTM

pkg/apis/application/v1alpha1/app_types.go
axbarsan, line 84: If this is `nullable`, shouldn't it be a pointer type?

tomahawk28: No, it's okay even it's not pointer type. k8s apiserver will hide NamespaceConfig when it is an empty struct. This is the same for `AppSpecInstall`.

axbarsan (approved):

PR #707, opened 2021-03-30
Document original-organization-name annotation

oponder:
We want to migrate fully to Organization CR's as the source of truth for Organizations, while delaying the task of figuring out how to migrate related resources.

Unfortunately, some organisations have a name that is incompatible with Organization CR's validation for names. That means, in order not to lose data, we will store that name into an annotation. This will let our UI's operate as normal, and further the transition to the MAPI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Added the `ui.giantswarm.io/original-organization-name` annotation
+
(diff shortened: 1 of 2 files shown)


marians (approved):
Works for me

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Added the `ui.giantswarm.io/original-organization-name` annotation
```
kopiczko, line 10: It sounds like `legacy-organization-name` from the description. But not a big deal for me.

PR #693, opened 2021-03-03
Add all CAPI CRDs

MarcelMue:
WE DON'T NEED CHECKLISTS WHERE WE ARE GOING

towards https://github.com/giantswarm/giantswarm/issues/15997

PR #712, opened 2021-04-19
Improve CRD generation

tfussell:
## Checklist


tfussell:
This would subsume https://github.com/giantswarm/apiextensions/pull/693 by the way.

hack/build-charts.go
```diff
@@ -0,0 +1,365 @@
...
+			},
+			ConversionReviewVersions: []string{
+				"v1",
+				"v1beta1",
+			},
+		},
+	}
+}
+
+func patchCAPAWebhook(crd *v1.CustomResourceDefinition) {
```
MarcelMue, line 155: Maybe I am missing something but is there no CAPZ webhook? Or should this be added later?

hack/build-charts.go
```diff
@@ -0,0 +1,365 @@
...
+		if *decodedGVK != crdGVK {
+			continue
+		}
+		crds = append(crds, crd)
+	}
+
+	return crds, nil
+}
+
+func downloadReleaseAssetCRDs(ctx context.Context, client *github.Client, asset releaseAssetFileDefinition) ([]v1.CustomResourceDefinition, error) {
```
MarcelMue, line 103: Currently this file seems pretty complex for a single file hack script. For the first iteration this is fine to me but we should think about splitting this into multiple files. Maybe also something we could contribute upstream?

hack/build-charts.go
```diff
@@ -0,0 +1,365 @@
...
+
+func patchCAPIWebhook(crd *v1.CustomResourceDefinition) {
+	port := int32(9443)
+	crd.Spec.Conversion = &v1.CustomResourceConversion{
+		Strategy: v1.WebhookConverter,
+		Webhook: &v1.WebhookConversion{
+			ClientConfig: &v1.WebhookClientConfig{
+				Service: &v1.ServiceReference{
+					Namespace: "giantswarm",
+					Name:      "cluster-api-core-unique-webhook",
```
MarcelMue, line 141: Hardcoding this name should be okay for now. Maybe we should document this a bit as it's non obvious where this would be coming from.

hack/build-charts.go
```diff
@@ -0,0 +1,365 @@
...
+		repo:     "cluster-api-provider-vsphere",
+		version:  "v0.7.6",
+		file:     "infrastructure-components.yaml",
+		provider: "vmware",
+	},
+	{
+		owner:    "Azure",
+		repo:     "aad-pod-identity",
+		version:  "v1.7.4",
+		file:     "deployment.yaml",
```
MarcelMue, line 66: All the other names are "agreed" to by upstream. Do we have some written agreement by azure folks that this name will stay the same?

MarcelMue (commented):
Generally okay with me - some comments

njuettner (approved):
Any ideas how can we merge this? I fixed gitleaks issue for the example credentials in AWSStaticIdentityRole CRD. 

For me it seems good, we can iterate on it and once Celestial also wants to include their webhooks they can easily add it 🙂.

tfussell:
> Any ideas how can we merge this? I fixed gitleaks issue for the example credentials in AWSStaticIdentityRole CRD.
> 
> For me it seems good, we can iterate on it and once Celestial also wants to include their webhooks they can easily add it 🙂.

I will address Marcel's comments this afternoon and then we can merge it tomorrow.

hack/build-charts.go
tfussell: I can't see this changing for a published release. It may change in future versions, but I think that can be handled when we update the version here.

PR #713, opened 2021-04-20
cabundle

njuettner:
This pr adds some minor changes to @tfussell PR. It will change the CABundle from conversion like we in [aws-admission-controller](https://github.com/giantswarm/aws-admission-controller/blob/master/helm/aws-admission-controller/templates/webhook.yaml#L20). Leave it more or less empty to let cert-manager inject it.

This PR also changes the inject-ca from the corresponding webhook in the `giantswarm` namespace instead of capi / capa namespaces. For now we assume the capa webhook is called `cluster-api-provider-aws-unique-webhook` which is responsible for **CRD** migration.

/cc @anvddriesch 

## Checklist

### DIFF
--- a/hack/build-charts.go
+++ b/hack/build-charts.go
@@ -132,6 +132,9 @@ func downloadReleaseAssetCRDs(ctx context.Context, client *github.Client, asset 
 
 func patchCAPIWebhook(crd *v1.CustomResourceDefinition) {
 	port := int32(9443)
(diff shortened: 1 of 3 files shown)


anvddriesch (approved):
nice!

njuettner:
@tfussell will merge this now into your helm PR 🙂

PR #730, opened 2021-05-17
Creating Catalog CRD

tomahawk28:
Toward giantswarm/roadmap#219

Adding `catalogs.application.giantswarm.io/v1alpha1`

## Checklist


pkg/apis/application/v1alpha1/catalog_types.go
```diff
@@ -0,0 +1,116 @@
...
+// +kubebuilder:printcolumn:name="Catalog URL",type=string,JSONPath=`.spec.storage.URL`,description="URL of the catalog"
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`,description="Time since created"
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:categories=common;giantswarm
+// +kubebuilder:storageversion
+// +k8s:openapi-gen=true
+// Catalog represents a catalog of managed apps. It stores general information for potential apps to install.
+// It is reconciled by app-operator.
+type Catalog struct {
```
tomahawk28, line 42: We are now using the new name `catalog` instead of `appcatalog`

pkg/apis/application/v1alpha1/catalog_types.go
```diff
@@ -0,0 +1,116 @@
...
+			},
+		},
+		TypeMeta: NewCatalogTypeMeta(),
+	}
+}
+
+// +kubebuilder:printcolumn:name="Catalog URL",type=string,JSONPath=`.spec.storage.URL`,description="URL of the catalog"
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`,description="Time since created"
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
tomahawk28, line 36: The only difference between `catalog` and `appcatalog` is `catalog` is namespace-scoped where `appcatalog` is cluster-scoped.

marians:
I'd welcome an addition to the `appcatalog` CRD main description saying that it's deprecated and being replaced with `catalog`.

pkg/apis/application/v1alpha1/catalog_types_test.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func newCatalogExampleCR() *Catalog {
+	cr := NewCatalogCR()
+
+	cr.ObjectMeta = metav1.ObjectMeta{
+		Name: "my-playground-catalog",
+		Labels: map[string]string{
+			"app-operator.giantswarm.io/version": "1.0.0",
+		},
```
rossf7, line 70:
```suggestion
		Name: "my-playground-catalog",
		Namespace: metav1.NamespaceDefault,
```

I'd like us to get rid of the version label from our docs. We should default it rather than asking the customer to set it.

We don't actually set a label selector in the current appcatalog controller. So we can just remove it. 🎉

rossf7 (approved):
LGTM

There is a just a change to the example CR I'd like us to make to drop the version label.

pkg/apis/application/v1alpha1/catalog_types_test.go
tomahawk28: Good catch, sure, I will remove those labels.

tomahawk28:
@marians Added deprecated. PTAL :)

pkg/apis/application/v1alpha1/app_catalog_types.go
```diff
@@ -36,6 +36,7 @@ func NewAppCatalogCR() *AppCatalog {
 // +kubebuilder:resource:categories=common;giantswarm,scope=Cluster
 // +kubebuilder:storageversion
 // +k8s:openapi-gen=true
+// Deprecated, Use Catalog CRD instead.
```
marians, line 39:
```suggestion
// Deprecated, use Catalog CRD instead.
```

marians (approved):
Little nitpick, but LGTM in general

marians (approved):

PR #734, opened 2021-05-18
add catalogNamespace into app CR

tomahawk28:
Toward giantswarm/roadmap#219

## Checklist

### DIFF
--- a/pkg/apis/application/v1alpha1/app_types.go
+++ b/pkg/apis/application/v1alpha1/app_types.go
@@ -57,6 +57,11 @@ type AppSpec struct {
 	Catalog string `json:"catalog"`
 	// +kubebuilder:validation:Optional
 	// +nullable
+	// CatalogNamespace is the namespace of the catalog this app belongs to.
+	// e.g. giantswarm
+	CatalogNamespace string `json:"catalogNamespace,omitempty"`
+	// +kubebuilder:validation:Optional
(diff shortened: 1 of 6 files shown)


pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -57,6 +57,11 @@ type AppSpec struct {
 	Catalog string `json:"catalog"`
 	// +kubebuilder:validation:Optional
 	// +nullable
+	// CatalogNamespace is the namespace of the catalog this app belongs to.
```
tomahawk28, line 60: Note that this belongs to `catalog` CRs, not `appCatalog`.

rossf7 (approved):
LGTM

Should we add catalogNamespace to the example CR in the UT?

rossf7:
@tomahawk28 Also, I think we should have a changelog entry for this. Since app CRs are customer-facing.

pkg/apis/application/v1alpha1/app_types.go
marians: You could make that more obvious by writing `... the namespace of the Catalog CR this ...`

marians:
> Should we add catalogNamespace to the example CR in the UT?

Always good to have the example as complete as possible

marians (approved):

PR #760, opened 2021-07-20
GS AWS v1alpha3 CRDs

njuettner:
I'd like to enable `v1alpha3` for AWS GiantSwarm CRD's to start preparing migration from `v1alpha2` to `v1alpha3`. So far we only enable `v1alpha3` but `v1alpha2` will be still be around for a while. `v1alpha3` will be set as the stored version. 

No changes are made on any field, so no CRD webhook is needed.

## Checklist


calvix (approved):
lgtm but its hard to fully review everything :p

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `v1alpha3` GS AWS CRD's.
```
marians, line 12:
```suggestion
- Add `v1alpha3` version for Giant Swarm AWS CRDs.
```

pkg/apis/infrastructure/v1alpha3/aws_cluster_types.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:categories=aws;cluster-api;giantswarm
+// +k8s:openapi-gen=true
+
+// AWSCluster is the infrastructure provider referenced in upstream CAPI Cluster
```
marians, line 40: I would prefer `Cluster API` over `CAPI` in general

marians:
Glanced over it. Can't really judge this, as I don't even understand why we are adding this version if everything is identical to v1alpha2. 🤷

pkg/apis/infrastructure/v1alpha3/aws_cluster_types.go
njuettner, line 40:
```suggestion
// AWSCluster is the infrastructure provider referenced in upstream Cluster API
```

CHANGELOG.md
```diff
@@ -21,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add deprecation info to CRD docs metadata.
 
+
```
njuettner, line 28:
```suggestion
```

njuettner:
> Glanced over it. Can't really judge this, as I don't even understand why we are adding this version if everything is identical to v1alpha2. 🤷

Good point, I'll try to explain it.

On AWS we ran `v1alpha2` for quite a while now. So far we updated upstream CRD's to `v1alpha3` but we never touched AWS Giantswarm CRD's. At some point we need to that if we want to try `v1alpha4` but we cannot do that because CRD conversion only supports two versions. If we want to use CAPI/CAPA with `v1alpha4` we need to get rid off `v1alpha2`, even for our CRD's because we use `infrastructureRef: v1alpha2`. 

There's no reason updating or changing any fields because the main goal would be still replacing GS CRD's with CAPA CRD's just to unblock us from trying out `v1alpha4`.

axbarsan (approved):
Exciting stuff

PR #763, opened 2021-07-22
Conversion webhook removed from upstream CAPZ CRDs

fiunchinho:
The CAPZ CRDs have a conversion webhook defined that it's not deployed on any Azure installations. If we ensure these CRDs, we will break the cluster because CRs wouldn't be converted. I'm removing it until we figure out how to deploy the conversion webhook.

I'm not sure how this repository works, but I believe upstream manifests are pulled remotely, then checkout in the repository, so I can just modify them and commit those changes.

## Checklist

- [X] Consider SIG UX feedback.


whites11 (approved):
yes please

MarcelMue (approved):

MarcelMue:
Please check while the makefile wants to add it back

njuettner (approved):

fiunchinho:
> Please check while the makefile wants to add it back

IIUC from the README, this repository tries to fetch whatever is released upstream, then it uses `make verify` to run `git diff --exit-code` and fail if there are local changes compared to upstream. But in this case, we want changes. So I'd say let's ignore that check until we deploy the conversion webhook?

fiunchinho:
I pushed a new commit that removes the call to `make verify` from CI. This call was validating that there were no local changes on the manifests downloaded from upstream releases. But we need to modify the `CAPZ` CRDs to remove the conversion webhook until we deploy it everywhere.

After deploying it, we can add the CI validation again.

PTAL.

MarcelMue (approved):

MarcelMue:
Please make a note in your team to add the check back later :)

fiunchinho:
I think I finally understood how this repository works. I shouldn't modify the generated manifests. We are using `kustomize` patches to modify upstream manifests, so instead of modifying the generated manifests I added a new patch to do the change I needed. I ended up not removing any checks from CI.

MarcelMue:
I agree this looks much cleaner now!

PR #768, opened 2021-07-28
Add capi/capz experimental types as if they were our own

fiunchinho:
In the more recent CAPI/CAPZ releases, the `MachinePool` and `AzureMachinePool` experimental types have changed their api group. If we would ensure the CRDs on the new releases, the CRD belonging to the old group may be deleted but our azure-operator still reconciles CRs on that api group. That's why we need to keep the CRD's on the old group, while the CAPZ controllers would reconcile CRs on the new one.

## Checklist


fiunchinho:
superseeded by https://github.com/giantswarm/apiextensions/pull/771

PR #771, opened 2021-07-29
Add capi/capz experimental types as if they were our own

fiunchinho:
In the more recent CAPI/CAPZ releases, the `MachinePool` and `AzureMachinePool` experimental types have changed their api group. If we would ensure the CRDs on the new releases, the CRD belonging to the old group may be deleted but our azure-operator still reconciles CRs on that api group. That's why we need to keep the CRD's on the old group, while the CAPZ controllers would reconcile CRs on the new one.

On this PR, the Go code is not really important, we only use it to generate the yaml manifests. The important bit is the `giantswarm.yaml` file, which is the one that will be installed/ensured. This one now contains the two types that are going away when upgrading to `v1alpha4`.

## Checklist


whites11 (approved):
I guess the go code was copied from capz 0.4.x.
It seems a reasonable approach to me.

fiunchinho:
> I guess the go code was copied from capz 0.4.x.
> It seems a reasonable approach to me.

Yes, those types don't exist on `v0.5.x`. I modified the go code a bit, but that's not really relevant. The important part is the generated yaml, because we are not using the clientsets or the go types, we just want to ensure the CRDs.

nprokopic (approved):

fiunchinho:
Me and Nikola talked over meet. Merging!

PR #773, opened 2021-08-02
Fix URL to node pools documentation

marians:


whites11 (approved):

marians:
When I hit `make` on this branch, it has weird side effects. Wait for the next commit...

marians:
I'm wrong. Now it hat no effect at all.

PR #776, opened 2021-08-09
Add support for new silence matcher types

QuentinBisson:
We can actually keep the existing silence CRD by adding a new value set to true by default

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add support for new matcher types in `silence.monitoring.giantswarm.io/v1alpha1` CRD
+
 ## [3.31.0] - 2021-08-06
 
 ### Added
--- a/config/crd/monitoring.giantswarm.io_silences.yaml
+++ b/config/crd/monitoring.giantswarm.io_silences.yaml
@@ -42,6 +42,8 @@ spec:
               matchers:
                 items:
                   properties:
+                    isEqual:
+                      type: boolean
                     isRegex:
                       type: boolean
                     name:
--- a/helm/crds-common/templates/giantswarm.yaml
+++ b/helm/crds-common/templates/giantswarm.yaml
@@ -2078,6 +2078,8 @@ spec:
               matchers:
                 items:
                   properties:
+                    isEqual:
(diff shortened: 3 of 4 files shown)


QuentinBisson:
I added the new field from https://github.com/prometheus/alertmanager/blob/c72c4d79f6d94429057944820b1c08767d540134/pkg/labels/matcher.go#L112

mig4:
A couple of questions, maybe daft but I feel like I'm missing some context:

- what's the goal exactly? the title says _support for new silence matcher types_ but in the diff I can only see a single field added; what are the new matcher types, etc.? Any chance you could link up to some references so I can catch up?
- in PR description you say _adding a new value false by default_ but in the code `IsEqual` is set to `true`, is this right?

QuentinBisson:
Sorry for the confusion, the value should be set to true by default.

Regarding the new field, the alertmanager folks added it to keep the api compatibility between v1 and v2 of the silence api.

They added the IsEqual field https://github.com/prometheus/alertmanager/blob/c72c4d79f6d94429057944820b1c08767d540134/pkg/labels/matcher.go#L95
This is set to true by default because the old api only supported equal matcher types so `=` or `~=`. 
As you can see here, the new Matcher type is the internal representation but the http object is still the old silence type:
https://github.com/prometheus/alertmanager/blob/c72c4d79f6d94429057944820b1c08767d540134/pkg/labels/matcher.go#L121

Basically, adding this field with a default value of true will not change the current behavior but we can now support negative regexes by setting isEqual to false in the Silence CRs

mig4:
> They added the IsEqual field [prometheus/alertmanager@`c72c4d7`/pkg/labels/matcher.go#L95](https://github.com/prometheus/alertmanager/blob/c72c4d79f6d94429057944820b1c08767d540134/pkg/labels/matcher.go#L95)
> This is set to true by default because the old api only supported equal matcher types so `=` or `~=`.
> As you can see here, the new Matcher type is the internal representation but the http object is still the old silence type:
> [prometheus/alertmanager@`c72c4d7`/pkg/labels/matcher.go#L121](https://github.com/prometheus/alertmanager/blob/c72c4d79f6d94429057944820b1c08767d540134/pkg/labels/matcher.go#L121)

Ah, got ya, so `IsEqual = false` means match not equal to, i.e. `!=` or `!~`. Nice, thanks for explaining 😁

mig4 (approved):

PR #684, opened 2021-02-15
Add sample CRs for Azure Cluster API types.

whites11:
## Checklist

- [x] Consider SIG UX feedback.
- [x] Update changelog in CHANGELOG.md.

PR #779, opened 2021-08-12
Add example CRs

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/15025

I'm adding two `MachinePools` and two `AzureMachinePools` but they belong to two different api groups. The ones with `exp` in the name will be removed eventually, as they got removed upstream.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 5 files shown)


tuommaki (approved):

marians:
I notice that these example CRs are added as YAML files. So far I have seen us create them in unit tests. [Example](https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/application/v1alpha1/catalog_types_test.go#L64-L92). I think the YAML works like a golden file. In a normal `go test` run the output is compared with the YAML in the repo. To generate the YAML, the `go test` is executed with `-update`.

Not sure if you are aware, and if you have good reasons to do it differently here.

fiunchinho:
I followed [the conversation you had with Chris on this PR](https://github.com/giantswarm/apiextensions/pull/684#issuecomment-779123440) and follow his lead.

marians:
Thanks for the reminder!

marians (approved):

PR #790, opened 2021-09-06
add upgrade annotations

anvddriesch:
Towards https://github.com/giantswarm/giantswarm/issues/18077

I am using these annotations for now while working on the upgrade-schedule-operator. This won't be merged until other things are in order so open to other suggestions.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `alpha.aws.giantswarm.io/update-schedule-target-release` and `alpha.aws.giantswarm.io/update-schedule-target-time` annotations.
+- Add `alpha.giantswarm.io/update-schedule-target-release` and `alpha.giantswarm.io/update-schedule-target-time` annotations.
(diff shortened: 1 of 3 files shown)


calvix:
i wonder if this should be AWS only, as it could be useful for other providers as well and the logic should does not need provider specific info

anvddriesch:
Yeah. If the upgrade mechanism is exactly the same on all providers there is nothing stopping us from making this a shared thing. We could still try it out on AWS first.

calvix:
> Yeah. If the upgrade mechanism is exactly the same on all providers there is nothing stopping us from making this a shared thing. We could still try it out on AWS first.

yeah, we can definitely alpha test it on AWS but then I would drop the was prefix from the annotation and maybe put it in a different file.

njuettner:
> > Yeah. If the upgrade mechanism is exactly the same on all providers there is nothing stopping us from making this a shared thing. We could still try it out on AWS first.
> 
> yeah, we can definitely alpha test it on AWS but then I would drop the AWS prefix from the annotation and maybe put it in a different file.

I agree. The annotation should be set on the Cluster CR, this should be independent of the infrastructure provider. I would move it to https://github.com/giantswarm/apiextensions/blob/master/pkg/annotation/cluster.go if possible. The operator should be anyway watch on Cluster CR, right?

anvddriesch:
Yeah, the operator is watching the cluster CR. I will move and rename the annotations! :)

njuettner (approved):

calvix (approved):

fiunchinho:
Remember that we decided on not adding more annotations and labels to `apiextensions` to decouple CRDs versioning from them. Can you add this to https://github.com/giantswarm/k8smetadata ?

anvddriesch:
Thanks for the heads up Jose! :) I will add it.

PR #801, opened 2021-10-06
Split CAPI CRD versions by provider

tfussell:
## Checklist


hack/assets.go
```diff
@@ -23,19 +24,20 @@ var upstreamReleaseAssets = []crd.ReleaseAssetFileDefinition{
...
 		Provider: "aws",
 	},
+	// azure
 	{
 		Owner:    "kubernetes-sigs",
-		Repo:     "cluster-api-provider-azure",
-		Version:  "v0.4.12",
-		Files:    []string{"infrastructure-components.yaml"},
+		Repo:     "cluster-api",
+		Version:  "v0.4.1",
```
tfussell, line 31: And CAPI v0.4.1 CRDs for Azure, KVM, and vSphere.

hack/patches.go
```diff
@@ -160,6 +160,7 @@ func patchReleaseValidation(crd *v1.CustomResourceDefinition) {
 
 var patches = map[string]crd.Patch{
 	"clusters.cluster.x-k8s.io":                                      patchCAPIWebhook,
+	"clusterclasses.cluster.x-k8s.io":                                patchCAPIWebhook,
```
tfussell, line 163: Unrelated change: I noticed that clusterclasses was missing the webhook patch. We don't use this CRD so it shouldn't affect anything.

helm/crds-vsphere/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
-name: crds-vmware
+name: crds-vsphere
```
tfussell, line 2: Unrelated change: renaming VMWare to vSphere for consistency.

njuettner (approved):
Ok, so we're putting CAPI CRD's to each provider because of different API versions, LGTMIIFY

marians:
> @giantswarm/sig-ux There might be some implications here around the CRD documentation.

@tfussell Could you check if the provider info in https://github.com/giantswarm/apiextensions/blob/split-capi-upstream/docs/config.yaml is then still up-to-date?

whites11 (approved):
LGTMIIFY

hack/patches.go
tfussell: Okay, I'll add it.

PR #831, opened 2021-11-04
Add Created At and Installed Version to App CRD additionalPrinterColumns

ubergesundheit:
This PR adds a `Created At` column to the output of `kubectl get app` and a `Installed Version` column to `kubectl get app -o wide`.

image

Edit: updated image to reflect changes in https://github.com/giantswarm/apiextensions/pull/831/commits/1fdc2d03a8522b9e616653dbdb02f92cbe134a33


rossf7 (approved):
LGTM

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -31,6 +31,8 @@ func NewAppCR() *App {
 }
 
 // +kubebuilder:printcolumn:name="Version",type=string,JSONPath=`.spec.version`,description="Version of the app"
+// +kubebuilder:printcolumn:name="Installed Version",type=string,priority=1,JSONPath=`.status.version`,description="Version of the app"
+// +kubebuilder:printcolumn:name="Created At",type=date,JSONPath=`.metadata.creationTimestamp`,description="Time of app creation"
```
rossf7, line 35: @ubergesundheit Thanks for adding both of these are useful. I think we just missed them when the printer cols were originally added.

marians:
I have a feeling that users will wonder about `VERSION` vs. `INSTALLED VERSION`. What does the former really mean then? Can we make it more clear? Should we change the former to something like `DESIRED VERSION`?

ubergesundheit:
> I have a feeling that users will wonder about VERSION vs. INSTALLED VERSION. What does the former really mean then? Can we make it more clear? Should we change the former to something like DESIRED VERSION? 

Yes, thanks for bringing that up! I would have already made this change but I hesitated to change existing behaviour of the current `Version` column.

Ideally I think the changes would be:

- Current `Version` column changes to `Installed version` (`.status.version`) whis is shown it for non `-o wide` output
- Show `.spec.version` field as `Desired Version` in `-o wide` output

tfussell (approved):

rossf7:
> Ideally I think the changes would be:

>  - Current Version column changes to Installed version (.status.version) whis is shown it for non -o wide output
> - Show .spec.version field as Desired Version in -o wide output

@ubergesundheit I agree this is better but I share your hesitation about changing the current value.

I wouldn't say the printer columns are part of the published interface. For automation the CR status should be used but there is still the chance this breaks some script. @marians WDYT?

marians:
I think we should be allowed to re-purpose this.

Yes, it could be that someone has created a script which used the default output format instead of a machine-friendly output like `-o json`. But that's definitely not recommended and I would rather now communicate this potentially breaking change and also use it to remind people that there is e. g. `-o json`.

marians:
To not forget this: @ubergesundheit Your suggestion makes total sense to me.

ubergesundheit:
Thanks for your input, applied the changes and will merge once CI is green :+1:

config/crd/application.giantswarm.io_apps.yaml
```diff
@@ -20,14 +20,14 @@ spec:
   scope: Namespaced
   versions:
   - additionalPrinterColumns:
-    - description: Version of the app
+    - description: Desired Version of the app
```
marians, line 23:
```suggestion
    - description: Desired version of the app
```

config/crd/application.giantswarm.io_apps.yaml
```diff
@@ -20,14 +20,14 @@ spec:
...
       jsonPath: .spec.version
-      name: Version
+      name: Desired Version
+      priority: 1
       type: string
-    - description: Version of the app
+    - description: Installed Version of the app
```
marians, line 28:
```suggestion
    - description: Installed version of the app
```

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -30,8 +30,8 @@ func NewAppCR() *App {
 	}
 }
 
-// +kubebuilder:printcolumn:name="Version",type=string,JSONPath=`.spec.version`,description="Version of the app"
-// +kubebuilder:printcolumn:name="Installed Version",type=string,priority=1,JSONPath=`.status.version`,description="Version of the app"
+// +kubebuilder:printcolumn:name="Desired Version",type=string,priority=1,JSONPath=`.spec.version`,description="Desired Version of the app"
```
marians, line 33:
```suggestion
// +kubebuilder:printcolumn:name="Desired version",type=string,priority=1,JSONPath=`.spec.version`,description="Desired Version of the app"
```

marians, line 33:
```suggestion
// +kubebuilder:printcolumn:name="Desired Version",type=string,priority=1,JSONPath=`.spec.version`,description="Desired version of the app"
```

pkg/apis/application/v1alpha1/app_types.go
```diff
@@ -30,8 +30,8 @@ func NewAppCR() *App {
...
+// +kubebuilder:printcolumn:name="Installed Version",type=string,JSONPath=`.status.version`,description="Installed Version of the app"
```
marians, line 34:
```suggestion
// +kubebuilder:printcolumn:name="Installed Version",type=string,JSONPath=`.status.version`,description="Installed version of the app"
```

marians:
Had a little uppercase/lowercase nitpick there

marians:
@ubergesundheit Are you going to make sure this update reaches the management clusters? Since it results in a UX change, I think it should happen more or less at the same time in all installations. I would also like to mention the effect in the next highlights / announcements after that.

ubergesundheit:
The way this will get rolled out is by creating a release of apiextensions, then `opsctl ensure crds --all` needs to be executed.

I'll probably do this today

PR #834, opened 2021-11-05
Add upstream cluster-name label to newClusterCR() function.

anvddriesch:
Towards https://github.com/giantswarm/giantswarm/issues/19683
## Checklist

### DIFF
--- a/pkg/apis/infrastructure/v1alpha3/cluster.go
+++ b/pkg/apis/infrastructure/v1alpha3/cluster.go
@@ -163,6 +163,7 @@ func newClusterCR(obj *AWSCluster, c ClusterCRsConfig) *apiv1alpha3.Cluster {
 		gsLabels := map[string]string{
 			label.ClusterOperatorVersion: c.ReleaseComponents["cluster-operator"],
 			label.Cluster:                c.ClusterID,
+			apiv1alpha3.ClusterLabelName: c.ClusterID,
 			label.Organization:           c.Owner,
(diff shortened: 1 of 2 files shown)


marians:
This seems to be missing the generated YAML. `make generate` I think.

pkg/apis/infrastructure/v1alpha3/cluster.go
```diff
@@ -163,6 +163,7 @@ func newClusterCR(obj *AWSCluster, c ClusterCRsConfig) *apiv1alpha3.Cluster {
 		gsLabels := map[string]string{
 			label.ClusterOperatorVersion: c.ReleaseComponents["cluster-operator"],
 			label.Cluster:                c.ClusterID,
+			apiv1alpha3.ClusterLabelName: c.ClusterID,
```
tfussell, line 166: Where are you using this function? I understand if you just want to make this small change here, but I would propose that you could alternatively move this function into the operator or component that calls it to reduce the dependence on apiextensions and CAPI libraries.

axbarsan (approved):

pkg/apis/infrastructure/v1alpha3/cluster.go
anvddriesch: This is used by `kubectl-gs`.
I agree with you that having this dependency might get more annoying in the future if we have more changes like this.

anvddriesch:
Closing in favor of https://github.com/giantswarm/kubectl-gs/pull/547

PR #840, opened 2021-11-15
Add AWSCNIPrefixDelegation annotation

njuettner:
This allows customers enabling the prefix delegation for AWS CNI which will be introduced within AWS release v16.1.
Enabling this feature will improve the speed of IP assignment on nitro based instances.

See https://github.com/aws/amazon-vpc-cni-k8s#enable_prefix_delegation-v190

Addresses: https://github.com/giantswarm/roadmap/issues/396

## Checklist


fiunchinho (changes requested):
Let's use https://github.com/giantswarm/k8smetadata instead

njuettner:
> Let's use https://github.com/giantswarm/k8smetadata instead

Shall this be used for the current GS project? I can't see k8smetadata is used in `aws-operator`?

njuettner:
This would move we need to move all AWSCNI annotations into k8smetadata otherwise we would have a split between the annotation settings on AWS CNI.

fiunchinho:
> > Let's use https://github.com/giantswarm/k8smetadata instead
> 
> Shall this be used for the current GS product? I can't see k8smetadata is used in `aws-operator`?

Yes, the idea is to migrate all labels and annotations to that repo to reduce dependency on the `apiextensions` repo. In `aws-operator`'s case it doesn't matter that much since it needs the dependency anyway, but I'd vote for use `k8smetadata` everywhere to avoid confusion.

fiunchinho:
> This would move we need to move all AWSCNI annotations into k8smetadata otherwise we would have a split between the annotation settings on AWS CNI.
> 
> Sorry, I just saw we already have the settings there, but I guess nobody updated the aws-operator

yeah I guess so.

njuettner:
@fiunchinho unfortunately we still need this PR merged, I think we need it to update the docs, see https://docs.giantswarm.io/ui-api/management-api/crd/awsclusters.infrastructure.giantswarm.io/#annotation-details-v1alpha2

njuettner:
We need to also add it here for the crd-docs-generator

fiunchinho (approved):
Alright then.

PR #845, opened 2021-11-29
Remove generated clientset

tfussell:
As part of the modernizing of `apiextensions`, I thought it would be helpful to remove the generated typed clients. So far it seems we can use `controller-runtime` client for everything (except evicting pods).

## Checklist


fiunchinho (approved):
This is the good stuff

image

marians:
Sorry, no idea what this really means and on what basis to review.

MarcelMue (approved):

anvddriesch (approved):

fiunchinho:
> Sorry, no idea what this really means and on what basis to review.

Using the code generation tools from upstream we were generating different clientsets for the different types (CRDs) that we registered in the k8s API. [A clientset is a http client](https://stackoverflow.com/questions/48388077/what-is-a-kubernetes-client-go-clientset) that knows how to talk to the k8s API to work with that specific type.

But the [controller-runtime folks](https://github.com/kubernetes-sigs/controller-runtime) offer a generic way of talking to the k8s API using the client on their controller-runtime library, [which doesn't need the typed client](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/example_test.go). This is mostly what we use today in our operators and libraries.

marians:
ACK, thanks for the info. Big breaking change, right? Will be released as a new Major version, I guess?

tfussell:
> ACK, thanks for the info. Big breaking change, right? Will be released as a new Major version, I guess?

Definitely. There will be some other breaking changes from my other PR as well.

PR #848, opened 2021-11-30
Use our CAPI and CAPA forks for CRDs in AWS MCs

nprokopic:
## Checklist


hack/assets.go
```diff
@@ -7,19 +7,17 @@ import (
...
 	{
-		Owner:   "kubernetes-sigs",
+		Owner:   "giantswarm",
 		Repo:    "cluster-api-provider-aws",
-		Version: "v0.6.5",
+		Version: "v1.2.0-gsalpha.1",
 		Files: []string{
-			"eks-bootstrap-components.yaml",
-			"eks-controlplane-components.yaml",
 			"infrastructure-components.yaml",
```
nprokopic, line 21: Release asset `infrastructure-components.yaml` in our CAPA fork is not automatically generated, I had to build it locally with `make` and then attach it to GitHub release.

nprokopic:
This PR changes file `helm/crds-aws/templates/upstream.yaml`, which is hard to review:
image

nprokopic:
> This PR changes file `helm/crds-aws/templates/upstream.yaml`, which is hard to review: image

Old file has 10619 lines, and the new one has 27323 lines. It's expected that the new one is larger, as it has more API versions, just not sure if this much larger.

hack/assets.go
AverageMarcus: What is this file? Do we need to update our release to build it?

AverageMarcus (approved):

nprokopic:
Changed it back to draft PR and renamed it, so it's more obvious that is should still not be merged, as we want to be sure that we will not break something.

hack/assets.go
nprokopic: It's a release asset that CAPA releases have, with all CRDs bundled inside. CAPI builds it with a GitHub workflow, and as those are defined in the repo code itself, we got it working in our fork out of the box.

OTOH for CAPA it's generated by running one of the `make` commands, but it's not in the GitHub workflow.

helm/crds-azure/templates/upstream.yaml
```diff
@@ -10,6 +10,19 @@ metadata:
     cluster.x-k8s.io/provider: cluster-api
   name: clusterresourcesetbindings.addons.cluster.x-k8s.io
 spec:
+  conversion:
```
nprokopic, line 13: This got added because I added `clusterresourcesetbindings.addons.cluster.x-k8s.io` to the list of patches above https://github.com/giantswarm/apiextensions/pull/848/files#diff-29beb93d62fe1968b6b3358f7ac76327fdc5e7b3108166be2abfc1f6282914beR254

nprokopic: This has to be generated for AWS, as there are 3 versions (v1alpha3, v1alpha4 and v1beta1).

The issue is that other providers also get this change applied, even without updating upstream CRDs, as the above `patches` list from `hack/patches.go` is used for all providers.

nprokopic: I guess that having a conversion webhook defined like this does not hurt when there is a single API version, it's just unnecessary IIUC.

tfussell:
I just merged https://github.com/giantswarm/apiextensions/pull/850. I don't want to interrupt your work here, but since it's a WIP I thought I would go ahead. There are some conflicts that will need to be resolved (cleaning CRDs and running `make generate` should mostly fix it). Let me know if you need a hand.

PR #850, opened 2021-12-06
Split CRDs into separate files

tfussell:
Inspired by problem encountered in https://github.com/giantswarm/apiextensions/pull/848. This is mainly to improve PR diffs when changing CRDs in the future.


tfussell:
The diff shows only 9 lines added. I don't think any CRDs have changed here, but I couldn't easily check all of them. It would be cool if teams could take a quick look at the CRDs they are responsible for.

MarcelMue (approved):
I scanned roughly and it looks fine to me.

PR #855, opened 2021-12-22
Upgrade CAPI CRDs to v1alpha4 on azure

fiunchinho:
We want to use v1alpha4 on Azure installations.

## Checklist


hack/patches.go
```diff
@@ -32,27 +32,8 @@ func patchCAPICoreWebhook(crd *v1.CustomResourceDefinition) {
...
-		patchCAPIKubeadmBootstrapWebhookV1Alpha3(crd)
-	}
-}
-
 // Keep in sync with https://github.com/giantswarm/cluster-api-bootstrap-provider-kubeadm-app/tree/main/helm/cluster-api-bootstrap-provider-kubeadm/templates
-func patchCAPIKubeadmBootstrapWebhookV1Alpha3(crd *v1.CustomResourceDefinition) {
+func patchCAPIKubeadmBootstrapWebhook(crd *v1.CustomResourceDefinition) {
 	port := int32(9443)
 	if _, ok := crd.Annotations["cert-manager.io/inject-ca-from"]; ok {
 		crd.Annotations["cert-manager.io/inject-ca-from"] = "giantswarm/cluster-api-bootstrap-cert"
```
nprokopic, line 39: This cert should not exist if `cluster-api-app` is deployed. All apps should use `cluster-api-core-cert`. This Jose guy made that change :grin: https://github.com/giantswarm/cluster-api-app/pull/3

If 3 separate apps are gonna get deployed, then I guess it can be fine.

fiunchinho: We want the single app on azure installations. Good catch, I'll change it.

nprokopic: The issue there is that I'm not sure how much the new app is ready to be deployed.

nprokopic: You probably want/need this in that case https://github.com/giantswarm/roadmap/issues/670 (see linked PR), but that's the  including also v1alpha2 types because of AWS MCs.

In order to get around that, I believe we should specify custom config for Azure MCs for overriding Docker images, so we use pure upstream instead of Docker images built from our fork.

nprokopic: This PR updates CAPI app to v1beta1 https://github.com/giantswarm/cluster-api-app/pull/17/files, it upgrades/fixes the manifests (not sure if everything is done).

The only issue there is that that PR is setting CAPI image version to `v1.1.0-gsalpha.2`, which is Docker image built from our fork and it also includes v1alpha2 types for AWS.

You could take over that CAPI app PR, test everything and merge it for Azure needs (and then we just use it for AWS later). Just set the CAPI image version you need for Azure, or better, set required version in config repo for desired Azure MCs (we need different CAPI images for Azure and AWS, and there would be less overriding per MC for Azure MCs, so I guess having default version set for AWS would be easier).

fiunchinho:
I have changed the patch mechanism a bit, so that it accepts a `provider` flag to cover the cases where we want to apply different patches on different providers.

whites11 (approved):
way too long to be reviewed, LGTMIIWFY

PR #860, opened 2022-01-21
Etcdbackup list clusters

whites11:
This PR adds a few new fields to etcdbackup CRD to add the possibility to backup specific clusters on demand rather than the whole installation.

## Checklist


njuettner (approved):
lgtm, just a question wouldn't adding new fields to a spec indicate we would need to add a new apiVersion? `v1alpha2` in this case, this would be the way how I would expected at least.

But I don't know how this is handled with GS CRD's in the past

calvix (approved):
lgtm

whites11:
> lgtm, just a question wouldn't adding new fields to a spec indicate we would need to add a new apiVersion? `v1alpha2` in this case, this would be the way how I would expected at least.
> 
> But I don't know how this is handled with GS CRD's in the past

since I only added stuff (and changed some fields to optional) this is not a breaking change.

PR #879, opened 2022-02-11
remove cluster CRD from KVM installations

glitchcrab:
towards https://github.com/giantswarm/giantswarm/issues/20702

whilst testing a new release, it was discovered that the existence of the `Cluster` CRD on KVM installations causes cert-operator to fail to provision certificates. `Cluster` CRs are not used on KVM, so the solution is to remove them.

## Checklist


yulianedyalkova (approved):
Tested with `opsctl ensure crds --apiextensions-reference remove-kvm-cluster -i gorgoth` and it works.

Might be nice to clean up all the other CRDs too but we can keep it minimal for now to speed things up.

PR #886, opened 2022-02-25
Use latest apiextensions-application

rossf7:
## Checklist


hack/assets.go
```diff
@@ -85,7 +85,7 @@ var remoteRepositories = []crd.RemoteRepositoryDefinition{
 		Owner:     "giantswarm",
 		Provider:  "common",
 		Name:      "apiextensions-application",
-		Reference: "v0.1.0",
+		Reference: "v0.3.1",
```
rossf7, line 88: I'm going to add this step to the apiextensions-application readme so we don't miss it in future.

rossf7:
Going with a ping.

PR #943, opened 2023-07-10
CRDs: Rename `nginx-ingress-controller-app` to `ingress-nginx`.

Gacko:
## Checklist


ubergesundheit:
This CRD seems to be not in use any more. I would suggest either asking the provider teams or just simply not to change this

Gacko:
Request review from @giantswarm/team-honeybadger, let's see. :)

